PR #2, opened 2022-04-27
Add implementation for RunContext methods

fmoletta:


src/compiler/instruction.rs
```diff
@@ -0,0 +1,75 @@
...
+    imm: Option<BigUint>,
+    dst_register: Register,
+    op0_register: Register,
+    op1_addr: Op1Addr,
+    res: Res,
+    pc_update: PcUpdate,
+    ap_update: ApUpdate
+    fp_update: FpUpdate,
+    opcode: Opcode
+
```
unbalancedparentheses, line 21: remove space @fmoletta

PR #4, opened 2022-04-29
Move instruction structure to vm

Juan-M-V:


src/vm/instruction.rs
```diff
@@ -1,14 +1,9 @@
...
-num Register {
-    AP,
-    FP
-}
-
 pub struct Instruction {
-    off0: BigUint,
-    off1: BigUint,
-    off2: BigUint,
+    off0: i16,
```
unbalancedparentheses, line 4: out of curiosity why are you changing the type @Juan-M-V ?

Juan-M-V: Because the specification divides the 64-bit words into off0, off1, off2 and flags, each taking 16-bits

unbalancedparentheses: Cool, thanks

unbalancedparentheses:
Is the PR ready to review? If it is still a draft please update it to be a draft. Thanks @Juan-M-V

Juan-M-V:
@unbalancedparentheses The PR is ready for review!

PR #9, opened 2022-05-03
Change get params

Juan-M-V:
Modify get function to work with BigInt's lack of cloning


Juan-M-V:
The memory method get was not behaving correctly with other parts of the code that expected a reference to a MaybeRelocatable. This PR aims to modify just the get method to avoid having to change all the other code

PR #12, opened 2022-05-04
Add build command

Juan-M-V:
Create makefile and added basic build command


Makefile
```diff
@@ -0,0 +1,16 @@
+# Make sure debug exists
+debug ?=
```
unbalancedparentheses, line 2: create new targets rather than using ifdef logic in this case

PR #16, opened 2022-05-05
Create Makefile

pefontana:
Create a Makefile with the following targets: 
* Build 
* run
* check
* test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+.PHONY: build run check test
+
+build:
+	cargo build
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,13 @@
...
+	cargo build
+
+run: 
+	cargo run
+
+check:
+	cargo check
+
+test:
+	cargo test
```
unbalancedparentheses, line 13: @pefontana add end of line

PR #18, opened 2022-05-06
Add Github Actions

pefontana:
Create .github/workflows/rust.yml

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,22 @@
+name: test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build
+      run: cargo build --verbose
+    - name: Run tests
+      run: cargo test --verbose
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push:
-    branches: [ main ]
+    branches: [ '*' ]
   pull_request:
     branches: [ main ]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -0,0 +1,22 @@
+name: test
+
+on:
```
igaray, line 3: This will trigger the action only on main, I find that developers often prefer having the CI run on their feature branches as well, so branches: '*' is better

.github/workflows/rust.yml
```diff
@@ -0,0 +1,22 @@
...
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build
```
igaray, line 19: This step won't succeed since the base ubuntu image does not have the rust toolchain installed. 
for that you need to either A) use a different image which includes rust or B) add a step to install rust: 

```
      - name: Install Rust
        uses: actions-rs/toolchain@v1
        with:
          profile: minimal
          toolchain: stable
          override: true
          components: clippy

      - name: Check examples
        uses: actions-rs/cargo@v1
        with:
          command: clippy
          args: --examples --all
```

etc

igaray (changes requested):

.github/workflows/rust.yml
pefontana: I think that the build is running OK, but maybe I am not getting something.
https://github.com/lambdaclass/cleopatra_cairo/actions

.github/workflows/rust.yml
pefontana: One question here Aki.

I made the changes here:
```
  push:
    branches: [ '*' ]
  pull_request:
    branches: [ main ]
```
And when I pushed the committed it to trigger two runs:

Screen Shot 2022-05-06 at 16 28 05

Maybe it is better to do:

```
  push:
    branches: [ main ]
  pull_request:
    branches: [ '*' ]
```

What do you think?

igaray (approved):

PR #23, opened 2022-05-09
Start deduce_memory_cell method

Juan-M-V:


unbalancedparentheses:
@Juan-M-V please update the files

Juan-M-V:
This function wont be implemented until deduction rules are in place

PR #52, opened 2022-05-17
Fix bug in run_instruction and add integrative test for step

fmoletta:


fmoletta:
This Pull Request should be merged after PR #50 
This will solve merge conflicts in decoder.rs

src/vm/vm_core.rs
```diff
@@ -2742,4 +2744,179 @@ mod tests {
 
         vm.opcode_assertions(&instruction, &operands);
     }
+
+    #[test]
+    ///Test for a simple program execution
```
unbalancedparentheses, line 2749: Use block comments for this /*   - Only a comment */

PR #53, opened 2022-05-19
Add integration test for step, using a simple program with a function call

fmoletta:


fmoletta:
Add test for a simple cairo program with a function call
Test is carried out by taking the initial data from a Python VM run, inserting it into a VirtualMachine structure in our Rust VM, running step in both VMs and checking that the affected data is equal in both runs

unbalancedparentheses:
> Add test for a simple cairo program with a function call Test is carried out by taking the initial data from a Python VM run, inserting it into a VirtualMachine structure in our Rust VM, running step in both VMs and checking that the affected data is equal in both runs

If you can please @fmoletta add this as some comment on the test

PR #196, opened 2022-06-28
Test makefile

Juan-M-V:
# Makefile test rule

## Description

Add setup to tests


Makefile
```diff
@@ -14,7 +14,9 @@ check:
 	cargo check
 
 test:
+	cd tests; ./setup_tests
 	cargo test
+	rm cairo_programs/*.json
```
pefontana, line 19: I think it would be more clear if  we add this removes in the setup_tests script

Makefile
```diff
@@ -14,7 +14,9 @@ check:
 	cargo check
 
 test:
+	cd tests; ./setup_tests
```
pefontana, line 17: lets rename this `setup_tests` to `setup_tests.sh`

Makefile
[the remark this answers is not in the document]
Oppen: `Make` simply invokes a shell to run the commands (after expression evaluation) so whatever changes of directories don't propagate outside of the target commands.

To illustrate, here's a `Makefile`:
```
test_call: test
	pwd

test:
	cd s5cmd; pwd

.PHONY: test test_call
```

If you run `make test_call`, which depends on `test` and thus triggers its build, you'll see output like this (mine is at the root of my code directory):
```
cd s5cmd; pwd
/Users/mario.rugiero/Code/s5cmd
pwd
/Users/mario.rugiero/Code
```

Note I never use `cd ..` and yet building the dependent target happens in the original dir.

pefontana:
@Juan-M-V one last thing :P
Please add a line to `.gitignore` that prevents pushing the generated cairo_programs/*.json

unbalancedparentheses:
> @Juan-M-V one last thing :P
> 
> Please add a regex match in the `.gitignore` that prevents pushing the generated `cairo_programs/*.json`

great comment. please create new pr for this

PR #199, opened 2022-06-29
 [coverage] Add memory errors display check

Juan-M-V:
# Increase codecov

## Description

Add testing for error displays

### DIFF
--- a/src/vm/vm_core.rs
+++ b/src/vm/vm_core.rs
@@ -669,10 +669,10 @@ mod tests {
     fn get_instruction_encoding_unsuccesful() {
         let mut vm = VirtualMachine::new(bigint!(39), Vec::new());
         vm.run_context.pc = MaybeRelocatable::from((0, 0));
-        assert_eq!(
-            Err(VirtualMachineError::InvalidInstructionEncoding),
-            vm.get_instruction_encoding()
-        );
+        if let Err(error) = vm.get_instruction_encoding() {
+            assert_eq!(error, VirtualMachineError::InvalidInstructionEncoding);
+            assert_eq!(error.to_string(), "Instruction should be an int. Found:");
+        }
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


src/vm/vm_core.rs
```diff
@@ -2243,13 +2258,13 @@ mod tests {
...
-                bigint!(5)
-            ))
-        );
+        if let Err(error) = vm.opcode_assertions(&instruction, &operands) {
+            assert_eq!(
+                error,
+                VirtualMachineError::CantWriteReturnPc(bigint!(9), bigint!(5))
+            );
+            assert_eq!(error.to_string(), "Call failed to write return-pc (inconsistent op0): 9 != 5. Did you forget to increment ap?");
+        }
```
Oppen, line 2267: Same here. We need to assert it's an error first.

Juan-M-V:
> Since there's the same suggestion (first assert it's an error) in some places and I don't want to spam you, here's my general advice: assert `.is_err` first, then `.unwrap_err` and make the assertion for display.

Thank you for the input, the error checking has been changed to use unwrap_err and remove the if let clause in commit 2f6e28d1cd396d9b3fce3a91d6baf609872b52f6

src/vm/vm_memory/memory.rs
```diff
@@ -208,6 +216,7 @@ mod memory_tests {
             error,
             Err(MemoryError::InconsistentMemory(key, val_a, val_b))
         );
+        assert_eq!(error.unwrap_err().to_string(), "Inconsistent memory assignment at address RelocatableValue(Relocatable { segment_index: 0, offset: 0 }). Int(5) != Int(6)");
```
Oppen, line 219: Nit: maybe split this.

Oppen (approved):
LGTM.

Oppen:
The performance assert is odd. We shouldn't observe any change in runtime from this changeset. Is the GH provided instance always the same type?

entropidelic (approved):
LGTM x2

Issue #75, opened 2022-05-27
Rustdoc documentation


igaray:
- The rustdoc comments should be polished and expanded in general.
- cleopatra is yet green to publish to crate.io and thus docs.rs but perhaps work could be done on the CI to add the rendered docs to the repo's github pages.

PartyLich:
> ... work could be done on the CI to add the rendered docs to the repo's github pages.

Are you accepting PRs for this?

If so, there's a working example in my fork. The [bench page](https://partylich.github.io/cleopatra_cairo/) remains the root, and has an additional link to the [docs](https://partylich.github.io/cleopatra_cairo/cleopatra_cairo/index.html).

pefontana:
Already done

PR #200, opened 2022-06-29
ci: add gh-pages docs workflow

PartyLich:
## Description

Partially addresses #75, providing rendered documentation on github pages. 

This pull adds a CI task and makefile target to build and deploy rustdocs. 

The current benchmark CI task should be unaffected, though its template is being updated in  [PR#201](https://github.com/lambdaclass/cleopatra_cairo/pull/201) to link to the generated html docs.

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added


.github/workflows/docs.yml
```diff
@@ -0,0 +1,29 @@
+name: docs
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ '*' ]
```
Oppen, line 6: We probably don't want to run this on PRs. Doing so could generate docs for bound to change features (say we iterate several times over a new API and no release includes it).

.github/workflows/docs.yml
```diff
@@ -0,0 +1,29 @@
...
+          toolchain: nightly
+          override: true
+          components: rustfmt, clippy
+    - uses: actions/checkout@v3
+    - name: Make Docs
+      run: make docs
+    - name: Jekyll fix
+      run: touch target/doc/.nojekyll
+    - name: Update gh-pages
+      uses: JamesIves/github-pages-deploy-action@v4.3.3
```
Oppen, line 24: This action seems to attempt to force push. That's actually why the pipelines is failing right now. Maybe there's a way to configure it so it creates a PR instead?

PartyLich: There is a `force` option enabled by default, with the stated alternative of 
> ..attempt to rebase new deployments onto any existing ones
when force is disabled. If that's not viable we may need another action entirely.

.github/workflows/docs.yml
PartyLich: Makes sense. Adjustment made.

PartyLich:
Thank you for taking the time to consider it! Updates have been made to address the review.

If the idea to have the task open a PR rather than commit directly really feels superior, I can do some investigation this weekend. The current action doesn't directly support it, but we might be able to do something along the lines of pushing to a new branch with the current action and using an additional action to open the PR. I could also fork the current action and adjust behavior, but the idea of maintaining that doesn't sound terribly appealing.

Oppen:
Hi @PartyLich, we just merged a PR removing the Rust pipelines from the `gh-pages` branch, if you want to rebase that should allow the workflows to pass.

Oppen:
Hi again @PartyLich. We've discussed internally with the team and came to the conclusion that the PR approach is much more desirable as it allows us to review the new docs before publishing. The idea of pushing to a new branch sounds appealing, maybe some `gh-pages_<short-ref>` would make sense.

PartyLich:
Okay, thanks for the update. I won't have time to write and test the alternative approach until the weekend, but I should have something going by or before Monday.

The proposed workflow will

1.  Build the docs when `main` is updated
2. Push a commit to `gh-pages_<short-ref>`
3. Open a PR from  `gh-pages_<short-ref>` --> `gh-pages` 

> That way we could also make it so pushes to that kind of branch create the PR or something similar (my Actions Fu is not really strong, so I'm not 100% sure that's feasible).

I was planning to include the above in a single docs workflow; are you also suggesting any push to a `gh-pages_<short-ref>` branch will likewise create a PR? Afaik that's feasible, we would just separate the PR creation step into its own workflow. So we'd have
- workflow 1:
  -  Steps 1 & 2 as above
- workflow 2:
  - Step 3 for *any* push to a branch with the `gh-pages_` prefix

Any specific desires for the PR title? It'll be easy to change later, but if there's a standard in mind already I'd rather get it as close to right as possible the first time.

PartyLich:
Updated. 

As currently written, runs of this workflow should produce a [PR like this one](https://github.com/PartyLich/cleopatra_cairo/pull/4):
image

Oppen:
I love this, thanks!

PR #208, opened 2022-06-29
Add error tests of functions in cairo_run

entropidelic:
# Add error tests of functions in cairo_run

## Description
Add tests for errors in `cairo_run`.
Some invalid programs were added to `tests/support` to make functions return errors.

## Checklist
- [ ] Unit tests added

### DIFF
--- a/src/cairo_run.rs
+++ b/src/cairo_run.rs
@@ -4,7 +4,7 @@ use crate::vm::runners::cairo_runner::CairoRunner;
 use crate::vm::trace::trace_entry::RelocatedTraceEntry;
 use num_bigint::BigInt;
 use std::fs::File;
-use std::io::{self, Write};
+use std::io::{self, Error, ErrorKind, Write};
 use std::path::Path;
 
 pub fn cairo_run(path: &Path) -> Result<CairoRunner, CairoRunError> {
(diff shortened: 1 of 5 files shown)


src/cairo_run.rs
```diff
@@ -49,16 +49,24 @@ pub fn write_output(cairo_runner: &mut CairoRunner) -> Result<(), CairoRunError>
...
-                "Failed to dump trace at position {}, serialize error: {}",
-                i, e
-            );
+        match bincode::serialize_into(&mut buffer, entry) {
+            Ok(()) => continue,
+            Err(e) => {
+                let error_string =
+                    format!("Failed to dump trace at position {i}, serialize error: {e}");
+                return Err(Error::new(ErrorKind::Other, error_string));
+            }
```
Oppen, line 65: This looks like `if let Err(e)` still did the job, just changing the body to the new one.

src/cairo_run.rs
```diff
@@ -155,34 +160,53 @@ mod tests {
...
-
-        assert!(expected_trace == serialized_program);
+        let expected_trace_path = Path::new("tests/support/struct.trace");
+        let cleopatra_trace_path = Path::new("tests/support/struct_cleopatra.trace");
+
+        // run test program until the end
+        let mut cairo_runner = match run_test_program(program_path) {
+            Ok(cairo_runner) => cairo_runner,
+            Err(_) => panic!("Could not run test program"),
+        };
```
Oppen, line 200: Why not `assert!(cairo_runner.is_ok())`?

src/cairo_run.rs
```diff
@@ -49,16 +49,21 @@ pub fn write_output(cairo_runner: &mut CairoRunner) -> Result<(), CairoRunError>
 
 /// Writes a trace as a binary file. Bincode encodes to little endian by default and each trace
 /// entry is composed of 3 usize values that are padded to always reach 64 bit size.
-pub fn write_binary_trace(relocated_trace: &[RelocatedTraceEntry], trace_file: &Path) {
-    let mut buffer = File::create(trace_file).expect("Error while creating trace file");
+pub fn write_binary_trace(
+    relocated_trace: &[RelocatedTraceEntry],
+    trace_file: &Path,
+) -> io::Result<()> {
+    let mut buffer = File::create(trace_file).expect("Could not create trace file");
```
Oppen, line 56: This should return an error as well.

PR #215, opened 2022-06-30
feat(CairoRunner): add a get_output function

tdelabro:
# Expose the raw output

## Description

Atm, you can only print the execution output. I need a way to get it directly from the lib, without printing it.
`write_output` now wrap the new `get_output`


Oppen:
Hi @tdelabro. Thanks for your contribution. May I suggest an inversion on the approach? Rather than generate everything in memory, which may be more costly in terms of memory for the real IO case, the `get_output` version could create a `Vec<8>` and passing that to `write_output` instead.

unbalancedparentheses:
> Hi @tdelabro. Thanks for your contribution. May I suggest an inversion on the approach? Rather than generate everything in memory, which may be more costly in terms of memory for the real IO case, the `get_output` version could create a `Vec<8>` and passing that to `write_output` instead.

+1

tdelabro:
I don't get what you mean.
In my PR `get_output` generate an in-memory `String` and pass it to `write_output`.
The only difference with what you describe is a having `Vec<u8>` instead of a `String`. When they are basically the same, at a memory level.
I think there are some missing pieces in your instructions :)

Oppen:
> I don't get what you mean.
> In my PR `get_output` generate an in-memory `String` and pass it to `write_output`.
> The only difference with what you describe is a having `Vec<u8>` instead of a `String`. When they are basically the same, at a memory level.
> I think there are some missing pieces in your instructions :)
> Do you mean a stream, or an iterator instead?

Sorry. The `Vec<8>` was my mistake. As long a `String` supports being written to, the type is OK.
The point of the comment wasn't that, it was that in the case I actually wanted to write to something other than memory (say a file), the way this PR implements it forces it to always create everything in a big buffer in memory first and only after flush it. By writing `get_output` as the wrapper and make it pass the `String` as an `io::Write` both cases get covered the same but the real IO case uses potentially less memory.

tdelabro:
The problem is that the `write_ouptut` function prepend the actual output with a "Program Output\n", which I don't want to be written when calling `get_output`.
So it seems to me that `get_output` should be the inner one. And `write_output` should then use it and add the customer formatting and writing destination it wants.

If the problem is about building in memory, I can make `get_output` return an iterator, wich are lazy.

Oppen:
> The problem is that the `write_ouptut` function prepend the actual output with a "Program Output\n", which I don't want to be written when calling `get_output`.

I should check the behavior of the original VM, but I wouldn't expect the function to include that line in the first place, I think that belongs to the caller.
@unbalancedparentheses @fmoletta, any opinions here?

tdelabro:
@Oppen @unbalancedparentheses @fmoletta ping

Oppen:
I'll make the call and say the caller should take care of printing the "Program output" line. This way we can just reuse `write_output` with a vec.

PR #224, opened 2022-07-01
Remove all .json files and improve setups

Juan-M-V:
# Remove .json files and improve setups

## Description

- Move all the .cairo programs to cairo_programs and delete .jsons.
- Add a benchmark folder to avoid compiling the benchmarks for regular tests
- Modify setups and move compiling functionality over to make.

Resolves: #214
Resolves: #227 

## Checklist
- [x] Linked to Github Issue


.github/workflows/bench.yml
```diff
@@ -40,3 +40,5 @@ jobs:
         alert-threshold: '130%'
         comment-on-alert: true
         alert-comment-cc-users: '@unbalancedparentheses'
+    - name: Clean benches
+      run: make clean
```
Oppen, line 44: Is this really necessary?

bench/criterion/setup_benchmarks.sh
```diff
@@ -1,9 +1,11 @@
...
 
-test_files=("bin-search")
+test_files=("integration" "fibonacci_1000" "linear-search")
 
 for file in ${test_files[@]}; do
-    cairo-compile "../../cairo_programs/$file.cairo" --output "../../cairo_programs/$file.json"
+    cairo-compile "../../cairo_programs/$file.cairo" --output "$file.json"
 done
+
+rm *.json
```
Oppen, line 11: Shouldn't we delete this file now?

.github/workflows/bench.yml
Juan-M-V: I'm not sure what happens if I leave the files after running. I'm trying to avoid any possible re-compiling of files (and ending up with `file.cairo.json.json`). I'm guessing that the whole point of using make is to avoid that re-compiling, right?

tests/compare_vm_output.sh
```diff
@@ -0,0 +1,24 @@
...
+        echo "Traces for $file match"
+    fi
+done
+
+for file in $(ls $tests_path | grep .cleopatra.memory | sed -E 's/\.cleopatra\.memory//'); do
+    if ! diff -q $tests_path/$file{,.cleopatra}.memory; then
+        echo "Memory differs for $file"
+        exit_code=1
+    else
+        echo "Memory mateches for $file"
```
Oppen, line 20:
```suggestion
        echo "Memory matches for $file"
```

Oppen (approved):
LGTM

PR #230, opened 2022-07-04
Add support to deserialize reference manager from json compiled programs

entropidelic:
# Deserialize Reference Manager

## Description
This PR adds support for deserializing the reference manager from compiled json cairo programs.
The approach is not the cleanest, and will be probably changed in the future, but it is good enough for a PoC.
The reference manager has information about the variables in the cairo program, so it is needed for hints execution implementation. The way the values of the variables of the program are referenced is by a string which has the information about the address in memory of the variable. This information is being parsed using regular expressions and then transformed to the struct `ValueAddress`.

## Checklist
- [x] Unit tests added

### DIFF
--- a/src/serde/deserialize_program.rs
+++ b/src/serde/deserialize_program.rs
@@ -63,8 +63,8 @@ pub struct Reference {
 
 #[derive(Deserialize, Debug, PartialEq)]
 pub struct ValueAddress {
-    pub register: Register,
-    pub offset: i32,
+    pub register: Option<Register>,
+    pub offset: Option<i32>,
 }
 
 struct BigIntVisitor;
@@ -177,24 +177,29 @@ impl<'de> de::Visitor<'de> for ValueAddressVisitor {
         E: de::Error,
     {
         let num_re = Regex::new(r"(-?\d+)").unwrap();
-        let reg_re = Regex::new(r"\w[.p]").unwrap();
+        let reg_re = Regex::new(r".p").unwrap();
 
-        let offset = num_re.captures(&value).unwrap();
-        let register_str = reg_re.captures(&value).unwrap();
+        let offset_capture = num_re.captures(value);
+        let register_capture = reg_re.captures(value);
 
-        println!("REGISTER: {}", &register_str[0]);
-        println!("OFFSET: {}", &offset[0]);
-
-        let reg = match &register_str[0] {
-            "fp" => Register::FP,
-            "ap" => Register::AP,
-            _ => return Err("hola").map_err(de::Error::custom),
+        let offset = match offset_capture {
+            Some(offset_str) => match i32::from_str(&offset_str[0]) {
+                Ok(offset) => Some(offset),
+                Err(e) => return Err(e).map_err(de::Error::custom),
+            },
+            _ => None,
         };
 
-        Ok(ValueAddress {
-            register: reg,
-            offset: i32::from_str(&offset[0]).unwrap(),
-        })
+        let register = match register_capture {
+            Some(register_str) => match &register_str[0] {
+                "fp" => Some(Register::FP),
+                "ap" => Some(Register::AP),
+                _ => None,
+            },
+            _ => None,
+        };
+
+        Ok(ValueAddress { offset, register })
     }
 }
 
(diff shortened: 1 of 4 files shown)


fmoletta (approved):
LGTM!

src/serde/deserialize_program.rs
```diff
@@ -137,6 +163,49 @@ impl<'de> de::Visitor<'de> for ReferenceIdsVisitor {
     }
 }
 
+struct ValueAddressVisitor;
+
+impl<'de> de::Visitor<'de> for ValueAddressVisitor {
+    type Value = ValueAddress;
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("Could not deserialize hexadecimal string")
```
pefontana, line 172: Change error message

pefontana (approved):
Nice one!
I think we are good to go!
Just a minor detail, change the ValueAddressVisitor error msg

src/serde/deserialize_program.rs
```diff
@@ -137,6 +163,49 @@ impl<'de> de::Visitor<'de> for ReferenceIdsVisitor {
...
+    }
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        // regex for capturing the register the address is referenced to
+        let register_re = Regex::new(r".p").unwrap();
+        // regex for capturing the offset of the reference
+        let offset_re = Regex::new(r"(-?\d+)").unwrap();
```
Oppen, line 182: We can precompile the regexes.

src/serde/deserialize_program.rs
```diff
@@ -31,16 +35,38 @@ pub struct FlowTrackingData {
...
+    #[serde(deserialize_with = "deserialize_value_address")]
+    #[serde(rename(deserialize = "value"))]
+    pub value_address: ValueAddress,
+}
+
+#[derive(Deserialize, Debug, PartialEq)]
+pub struct ValueAddress {
+    pub register: Option<Register>,
+    pub offset: Option<i32>,
+}
```
Oppen, line 68: In which scenario would we have a `ValueAddress` with no associated register?
Related, can't we just always treat the offset as `i32`? An offset of `0` is the same as no offset.

Oppen: Also, is the offset guaranteed to fit in 32 bits?

src/serde/deserialize_program.rs
```diff
@@ -137,6 +163,49 @@ impl<'de> de::Visitor<'de> for ReferenceIdsVisitor {
...
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("Could not deserialize hexadecimal string")
+    }
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: de::Error,
+    {
+        // regex for capturing the register the address is referenced to
+        let register_re = Regex::new(r".p").unwrap();
```
Oppen, line 180: It's better to restrict the matched character to the correct registers:
```suggestion
        let register_re = Regex::new(r"[af]p").unwrap();
```
The more specific the regex, the more efficient the resulting state machine and its compilation.

Oppen: (Typically)

src/serde/deserialize_program.rs
entropidelic: There are more references in the `reference manager` than just the ones used by hints. For the moment we don´t need them, so those references are set with None values in the `ValueAddress`.
If we wanted them, there should be another logic for parsing those references. They have to be present though,  because references are accessed by an id (`reference_ids` of `FlowTrackingData` struct), and that id corresponds to the index of the list of references in the reference manager.

entropidelic: If this sounds a little confusing let me know and I will explain in more detail :)

src/serde/deserialize_program.rs
entropidelic: How should I handle the precompilation?

src/serde/deserialize_program.rs
entropidelic: done!

src/serde/deserialize_program.rs
entropidelic: done :)

PR #247, opened 2022-07-06
Trace and memory compare

Juan-M-V:
# Trace and memory comparison

## Description

- Added a new script to compare the memory of the original (python) vm and the cleopatra vm (rust). This comparison is done only checking for values (and possible repetition) but not for order. 
- Trace is compared 1 to 1 with diff.
- Added a set of make rules to compare trace and/or memory
- Add the trace and memory check to github actions

We've found that there were negative values being inserted into the vm. This was resolved by using `mod_floor` instead of `%` and adding it to the subtraction.
There was also a bug in the serialization. The indexes started at 1 instead of 0 and we were filtering None values.

Resolves: #235 

## Checklist
- [x] Linked to Github Issue


Makefile
```diff
@@ -67,8 +67,14 @@ flamegraph:
...
-compare_vm_output: $(CLEO_TRACE) $(CAIRO_TRACE) $(CLEO_MEM) $(CAIRO_MEM)
-	cd tests; ./compare_vm_output.sh
+compare_trace_memory: $(CLEO_TRACE) $(CAIRO_TRACE) $(CLEO_MEM) $(CAIRO_MEM)
+	cd tests; ./compare_vm_state.sh trace memory
+
+compare-trace: $(CLEO_TRACE) $(CAIRO_TRACE)
+	cd tests; ./compare_vm_state.sh trace
+
+compare-memory: $(CLEO_MEM) $(CAIRO_MEM)
+	cd tests; ./compare_vm_state.sh memory
```
Oppen, line 77: Let's be consistent about target naming:
```suggestion
compare_trace: $(CLEO_TRACE) $(CAIRO_TRACE)
	cd tests; ./compare_vm_state.sh trace

compare_memory: $(CLEO_MEM) $(CAIRO_MEM)
	cd tests; ./compare_vm_state.sh memory
```

tests/memory_comparator.py
```diff
@@ -0,0 +1,24 @@
...
+    cairo_mem = Counter()
+    cleo_mem = Counter()
+
+    with open(filename1, 'rb') as f:
+        byte = f.read(40)
+        cairo_mem.update(byte)
+
+    with open(filename2, 'rb') as f:
+        byte = f.read(40)
+        cleo_mem.update(byte)
```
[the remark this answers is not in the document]
Oppen: But there's no loop. It'll read the first k/v only. Besides, the byte array is an iterable, `cleo_mem.update` will count by byte rather than the block. You would need to construct a list of byte arrays.
Probably something like:
```python3
cleo_mem.update(byte for byte in f.read(40))
```
We should see how EOF behaves tho, it may be wiser to explictly loop.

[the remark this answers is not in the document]
Oppen: The current behavior is similar to this:
```python3
>>> from collections import Counter
>>> b = bytes([1]*40)
>>> c = Counter()
>>> c.update(b)
>>> c
Counter({1: 40})
>>> c
```

src/types/relocatable.rs
```diff
@@ -129,12 +130,16 @@ impl MaybeRelocatable {
     ///Substracts two MaybeRelocatable values and returns the result as a MaybeRelocatable value.
     /// Only values of the same type may be substracted.
     /// Relocatable values can only be substracted if they belong to the same segment.
-    pub fn sub(&self, other: &MaybeRelocatable) -> Result<MaybeRelocatable, VirtualMachineError> {
+    pub fn sub(
+        &self,
+        other: &MaybeRelocatable,
+        prime: BigInt,
```
Oppen, line 136: Pass by reference.

tests/memory_comparator.py
```diff
@@ -0,0 +1,58 @@
...
+        for k in cairo_mem:
+            if k in cleo_mem:
+                continue
+            print(f'{k}:{v}')
+        print('keys in Cleopatra but not Cairo:')
+        for k in cleo_mem:
+            if k in cairo_mem:
+                continue
+            print(f'{k}:{v}')
+        print('mismatched values (Cairo <-> Cleopatra)):')
```
Oppen, line 46: This should be inverted to match the order of arguments passed by the script.

Juan-M-V: Resolved: https://github.com/lambdaclass/cleopatra_cairo/pull/247/commits/bfae7681847d6052fe8db5ad108c0eead2a88b1b

src/types/relocatable.rs
Juan-M-V: The rest of the functions in `relocatable.rs` are not being passed by reference. I opened an issue (#264)

tests/memory_comparator.py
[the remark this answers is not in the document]
Juan-M-V: https://github.com/lambdaclass/cleopatra_cairo/pull/247/commits/022a31ad7a43c233ed6e78c932d1e28b80dbd372

src/types/relocatable.rs
Oppen: We can start by doing it here. Several has been fixed already in a PR (which you approved and I merged just now).

Juan-M-V: Resolved: https://github.com/lambdaclass/cleopatra_cairo/pull/247/commits/88a3e7720eac9cae9e5a8ac8bdbb1a4db0ee0bc8

src/types/relocatable.rs
```diff
@@ -102,19 +102,10 @@ impl MaybeRelocatable {
...
-                    offset: new_offset,
-                }))
-            }
-            (&MaybeRelocatable::Int(ref num_ref), MaybeRelocatable::RelocatableValue(rel)) => {
-                let big_offset: BigInt = (num_ref + rel.offset).mod_floor(&prime);
+            (&MaybeRelocatable::RelocatableValue(ref rel), &MaybeRelocatable::Int(ref num_ref))
+            | (&MaybeRelocatable::Int(ref num_ref), &MaybeRelocatable::RelocatableValue(ref rel)) =>
+            {
+                let big_offset: BigInt = (num_ref + rel.offset) % prime;
                 let new_offset = match big_offset.to_usize() {
```
Oppen, line 109: Duplicate logic. You can remove the first line.

src/types/relocatable.rs
```diff
@@ -132,13 +123,13 @@ impl MaybeRelocatable {
...
-        prime: BigInt,
+        prime: &BigInt,
     ) -> Result<MaybeRelocatable, VirtualMachineError> {
         match (self, other) {
             (&MaybeRelocatable::Int(ref num_a_ref), &MaybeRelocatable::Int(ref num_b_ref)) => {
                 let num_a = Clone::clone(num_a_ref);
                 let num_b = Clone::clone(num_b_ref);
-                Ok(MaybeRelocatable::Int((num_a - num_b).mod_floor(&prime)))
+                Ok(MaybeRelocatable::Int((num_a - num_b).mod_floor(prime)))
             }
```
Oppen, line 133: Remove this line.

src/types/relocatable.rs
```diff
@@ -402,7 +393,7 @@ mod tests {
     fn sub_relocatable_from_relocatable_diff_offset() {
         let addr_a = &MaybeRelocatable::from((7, 17));
         let addr_b = &MaybeRelocatable::from((8, 7));
-        let error = addr_a.sub(addr_b, bigint!(23));
```
Oppen: Restore this line.

src/types/relocatable.rs
```diff
@@ -402,7 +393,7 @@ mod tests {
...
+        let error = addr_a.sub(addr_b, &bigint!(23));
         assert_eq!(error, Err(VirtualMachineError::DiffIndexSub));
```
Oppen, line 397: Remove this line.

src/types/relocatable.rs
```diff
@@ -383,15 +374,15 @@ mod tests {
...
-        let sub_addr = addr_a.sub(addr_b, bigint!(23));
+        let sub_addr = addr_a.sub(addr_b, &bigint!(23));
         assert_eq!(Ok(MaybeRelocatable::from(bigint!(2))), sub_addr);
     }
 
     #[test]
     fn sub_relocatable_from_relocatable_same_offset() {
         let addr_a = &MaybeRelocatable::from((7, 17));
         let addr_b = &MaybeRelocatable::from((7, 7));
-        let sub_addr = addr_a.sub(addr_b, bigint!(23));
```
Oppen: Restore.

src/types/relocatable.rs
```diff
@@ -383,15 +374,15 @@ mod tests {
...
+        let sub_addr = addr_a.sub(addr_b, &bigint!(23));
         assert_eq!(
```
Oppen, line 386: Remove.

src/types/relocatable.rs
```diff
@@ -383,15 +374,15 @@ mod tests {
     fn sub_int_from_int() {
         let addr_a = &MaybeRelocatable::from(bigint!(7));
         let addr_b = &MaybeRelocatable::from(bigint!(5));
-        let sub_addr = addr_a.sub(addr_b, bigint!(23));
+        let sub_addr = addr_a.sub(addr_b, &bigint!(23));
         assert_eq!(Ok(MaybeRelocatable::from(bigint!(2))), sub_addr);
```
Oppen, line 378: Remove.

src/types/relocatable.rs
```diff
@@ -132,13 +123,13 @@ impl MaybeRelocatable {
...
         &self,
         other: &MaybeRelocatable,
-        prime: BigInt,
+        prime: &BigInt,
     ) -> Result<MaybeRelocatable, VirtualMachineError> {
         match (self, other) {
             (&MaybeRelocatable::Int(ref num_a_ref), &MaybeRelocatable::Int(ref num_b_ref)) => {
                 let num_a = Clone::clone(num_a_ref);
                 let num_b = Clone::clone(num_b_ref);
-                Ok(MaybeRelocatable::Int((num_a - num_b).mod_floor(&prime)))
```
Oppen: Should use the references directly for the substraction, no need to clone.

Oppen (approved):

Issue #195, opened 2022-06-28
Separate each builtin into a different file


fmoletta:
The implementation of each builtin should be in a different file, instead of having every builtin in builtin_runners.rs

PR #248, opened 2022-07-06
Chore: Separate each builtin into a different file #195

PartyLich:
# Separate each builtin into a different file

## Description

Issue #195

- Moves the implementation of each builtin runner into its own file. 
- Export paths remain unchanged. 
- No additional tests have been added. Functionality should not be affected, it's just a reorganization chore

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added


Juan-M-V (approved):
Thank you again for your work!
The changes look good to me

PartyLich:
Glancing at the failing workflow my assumption is that the permissions issue is because the PR originates outside the org/repo. 

Eg for this PR it has
```
GITHUB_TOKEN Permissions
  Contents: read
  Deployments: read
  Metadata: read
```

For a PR from within the organization [(ex)](https://github.com/lambdaclass/cleopatra_cairo/runs/7237379619?check_suite_focus=true), the same workflow gets
```
GITHUB_TOKEN Permissions
  Contents: write
  Deployments: write
  Metadata: read
```

If it was allowed to pass, a random stranger on the internet (me in this case) would be creating automated commits on `gh-pages`. The benchmark workflow runs *and attempts to commit* on all PRs, including those outside the organization or repo. It could probably be improved, but preventing external commits that have no approval or merge seems like the right result. External parties like myself shouldn't have write access.

unbalancedparentheses:
Can we recheck this @Oppen so that we merge it at some point?

Oppen:
I'll try to see into this today.

Oppen:
Applied via a different PR.

PR #252, opened 2022-07-07
Remove regex from deserialization and generalize hint variable references

entropidelic:
# Remove regex from deserialization and generalize hint variable references

## Description
Linked to issue #236

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added


src/serde/deserialize_utils.rs
```diff
@@ -0,0 +1,359 @@
...
+                error.fmt(f)
+            }
+        }
+    }
+}
+
+// Checks if the hex string has an odd length.
+// If that is the case, prepends '0' to it.
+pub fn maybe_add_padding(mut hex: String) -> String {
+    if hex.len().rem(2) != 0 {
```
Oppen, line 37:
```suggestion
    if hex.len().is_odd() {
```

src/serde/deserialize_utils.rs
```diff
@@ -0,0 +1,359 @@
...
+    let mut offset1_str = splitted_value_str[1].split(',').collect::<Vec<_>>()[0].to_string();
+    offset1_str.retain(|c| !r#"()]"#.contains(c));
+
+    let offset1: i32 = match offset1_str.parse() {
+        Ok(offset1) => offset1,
+        // for the moment, references with values that overflow i32 are not important, they are just dummy references.
+        Err(e) => match e.kind() {
+            IntErrorKind::PosOverflow => 0,
+            _ => return Err(ReferenceParseError::IntError(e)),
+        },
```
Oppen, line 94: In fact, as long as they're offsets rather than values, they should remain in `i16` ranges, so this is correct code _under that assumption_.

src/serde/deserialize_utils.rs
```diff
@@ -0,0 +1,359 @@
...
+
+// parse string values of format `[cast(reg + offset1, *felt)]`
+fn parse_dereference_with_one_offset(
+    splitted_value_str: Vec<&str>,
+) -> Result<ValueAddress, ReferenceParseError> {
+    let register = match splitted_value_str[0].split('(').collect::<Vec<&str>>()[1] {
+        "ap" => Some(Register::AP),
+        "fp" => Some(Register::FP),
+        _ => None,
+    };
```
Oppen, line 83:
```suggestion
    let mut deref = parse_dereference_no_offsets(splitted_value_str);
```

Then you can overwrite `deref.offset1` right before returning.

src/serde/deserialize_utils.rs
```diff
@@ -0,0 +1,359 @@
...
+    offset1_str.retain(|c| !r#"()]"#.contains(c));
+
+    let offset1: i32 = match offset1_str.parse() {
+        Ok(offset1) => offset1,
+        // for the moment, references with values that overflow i32 are not important, they are just dummy references.
+        Err(e) => match e.kind() {
+            IntErrorKind::PosOverflow => 0,
+            _ => return Err(ReferenceParseError::IntError(e)),
+        },
+    };
```
Oppen, line 126: Similarly:
```suggestion
) -> Result<ValueAddress, ReferenceParseError> {
    let mut deref = parse_dereference_with_one_offset(splitted_value_str);
```
And modify `deref.offset2` at the end.

src/serde/deserialize_utils.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    Ok(ValueAddress {
+        register,
+        offset1,
+        offset2: 0,
+        immediate: Some(immediate),
+        dereference: false,
+    })
+}
```
Oppen, line 244: This looks oddly similar to the `dereference` cases. Specially the no offset and one offset cases.

src/vm/hints/hint_utils.rs
```diff
@@ -47,6 +46,42 @@ fn get_address_from_reference(
...
+        return None;
+    }
+    Some(MaybeRelocatable::from((
+        relocatable.segment_index,
+        (relocatable.offset as i32 + hint_reference.offset1) as usize,
+    )))
+}
+
+fn compute_addr_from_two_offsets(
+    relocatable: Relocatable,
```
Oppen, line 65: By ref?

src/vm/hints/hint_utils.rs
```diff
@@ -47,6 +46,42 @@ fn get_address_from_reference(
...
+    let addr = if hint_reference.offset1.is_negative()
+        && relocatable.offset < hint_reference.offset1.abs() as usize
+    {
+        return None;
+    } else {
+        Relocatable {
+            segment_index: relocatable.segment_index,
+            offset: (relocatable.offset as i32 + hint_reference.offset1) as usize,
+        }
+    };
```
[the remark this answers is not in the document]
Oppen: In fact, it seems like it could simply be a single function adding both offsets. If one of them is `0` then the value will be the same adding it or not.

src/vm/runners/cairo_runner.rs
```diff
@@ -225,11 +227,20 @@ impl CairoRunner {
...
-                    offset: reference.value_address.offset,
+                    offset1: reference.value_address.offset1,
+                    offset2: reference.value_address.offset2,
+                })
+            } else {
+                // just a dummy HintReference to maintian the same lenght of vectors Vec<References> and Vec<HintReference>
+                references.push(HintReference {
+                    register: Register::FP,
+                    offset1: 0,
+                    offset2: 0,
```
Oppen, line 243: Maybe we could store a tuple with the index into `Vec<References>` and the `HintReference` instead of artificially increasing memory use.

src/vm/hints/hint_utils.rs
```diff
@@ -47,6 +46,42 @@ fn get_address_from_reference(
...
+}
+
+fn compute_addr_from_two_offsets(
+    relocatable: Relocatable,
+    hint_reference: &HintReference,
+) -> Option<MaybeRelocatable> {
+    let addr = if hint_reference.offset1.is_negative()
+        && relocatable.offset < hint_reference.offset1.abs() as usize
+    {
+        return None;
```
Oppen, line 71: I'm not quite sure why we do this, could you explain this to me?

src/serde/deserialize_utils.rs
entropidelic: done!

src/vm/runners/cairo_runner.rs
```diff
@@ -225,11 +227,20 @@ impl CairoRunner {
...
         for reference in self.program.reference_manager.references.iter() {
             if let Some(register) = &reference.value_address.register {
                 references.push(HintReference {
                     register: register.clone(),
-                    offset: reference.value_address.offset,
+                    offset1: reference.value_address.offset1,
+                    offset2: reference.value_address.offset2,
+                })
+            } else {
+                // just a dummy HintReference to maintian the same lenght of vectors Vec<References> and Vec<HintReference>
```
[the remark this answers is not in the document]
entropidelic: done!

src/serde/deserialize_utils.rs
entropidelic: there are same cases where references (not used by hints, just in de deserializing process) have values of the form `cast(num, felt*)`, and `num` can be a very large integer, and the overflow happens. We are not interested in dealing with this values ATM so if the overflow occurs, returning 0 shoud be fine.

src/serde/deserialize_utils.rs
entropidelic: Yes you are right, I made a refactor to reduce redundant code

src/vm/hints/hint_utils.rs
entropidelic: This is just for checking that when we subtract the hint reference offset to the offset of the memory cell, we don't go outside the memory segment. I discussed this with @fmoletta and she told me this should be the expected behaviour

Oppen: Does `relocatable.offset` indicate the max offset for the segment?

src/serde/deserialize_program.rs
```diff
@@ -176,27 +178,20 @@ impl<'de> de::Visitor<'de> for ValueAddressVisitor {
...
+                Ok(res) => res,
                 Err(e) => return Err(e).map_err(de::Error::custom),
-            }
-        } else {
-            0
+            },
+            Some('c') => match deserialize_utils::parse_reference(value) {
+                Ok(res) => res,
+                Err(e) => return Err(e).map_err(de::Error::custom),
+            },
```
Oppen, line 189: Similarly here.

src/vm/hints/hint_utils.rs
```diff
@@ -39,12 +39,18 @@ fn compute_addr_from_reference(
...
     if let Some(index) = reference_id.to_usize() {
         if index < references.len() {
-            return compute_addr_from_reference(&references[index], run_context);
+            match references.get(&index) {
+                Some(hint_reference) => {
+                    return compute_addr_from_reference(hint_reference, run_context)
+                }
+                None => return None,
+            }
+            // return compute_addr_from_reference(&references.get(&index), run_context);
```
Oppen, line 53: Remove.

src/vm/hints/hint_utils.rs
```diff
@@ -39,12 +39,18 @@ fn compute_addr_from_reference(
...
 ) -> Option<MaybeRelocatable> {
     if let Some(index) = reference_id.to_usize() {
         if index < references.len() {
-            return compute_addr_from_reference(&references[index], run_context);
+            match references.get(&index) {
+                Some(hint_reference) => {
+                    return compute_addr_from_reference(hint_reference, run_context)
+                }
+                None => return None,
+            }
```
Oppen, line 52:
```suggestion
            if let Some(hint_reference) = references.get(&index) {
                compute_addr_from_reference(hint_reference, run_context)
            }
```

src/vm/runners/cairo_runner.rs
```diff
@@ -225,27 +223,32 @@ impl CairoRunner {
...
+                );
+            }
+            //     } else {
+            //         // Dummy HintReference to maintain the same length of vectors Vec<References> and Vec<HintReference>
+            //         references.push(HintReference {
+            //             register: Register::FP,
+            //             offset1: 0,
+            //             offset2: 0,
+            //         })
+            //     }
```
Oppen, line 247: Delete.

Oppen (approved):
LGTM

PR #260, opened 2022-07-11
Add implementation for hints on split_int (math.cairo)

fmoletta:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added


src/vm/errors/vm_errors.rs
```diff
@@ -120,6 +122,8 @@ impl fmt::Display for VirtualMachineError {
             },
             VirtualMachineError::UnknownHint(hint_code) => write!(f, "Unknown Hint: {:?}", hint_code),
             VirtualMachineError::MemoryError(memory_error) => memory_error.fmt(f),
+            VirtualMachineError::SplitIntNotZero => write!(f,"split_int(): value is out of range"),
```
Oppen, line 125: Why not include the value?

src/serde/deserialize_program.rs
```diff
@@ -188,10 +188,7 @@ impl<'de> de::Visitor<'de> for ValueAddressVisitor {
             None
         };
         let offset = if let Some((_, offset_str)) = regex_captures!(r"(-?\d+)", value) {
-            match i32::from_str(offset_str) {
-                Ok(offset) => offset,
-                Err(e) => return Err(e).map_err(de::Error::custom),
-            }
+            i32::from_str(offset_str).unwrap_or(-1)
```
Oppen, line 191: Why `-1` when it doesn't match? Wouldn't `0` make more sense (no offset -> no addition -> addition of `0`)?

src/vm/errors/vm_errors.rs
fmoletta: The value that is out of range is the one that the function split_int (and therefore this hint) receives when it its first called, this case is reached after n recursive calls, where the value received by the hint is no longer the one received by the function, so it doesn't make sense to show this number to the user.

src/serde/deserialize_program.rs
fmoletta: This is an intermediate solution, as this case is not currently handled and produces failed parse errors on references that don't belong to hints. For example: "cast(18446744073709551616, felt)" is not contemplated. Any offset value would be wrong as there is no register. I will change it to 0 for consistency so that it is None + 0 = None.

src/vm/hints/hint_utils.rs
```diff
@@ -407,3 +407,117 @@ pub fn assert_nn(
...
+    match vm.memory.get(&value_addr) {
+        Ok(Some(maybe_rel_value)) => {
+            //Check that the value at the ids address is an Int
+            let value = if let MaybeRelocatable::Int(ref value) = maybe_rel_value {
+                value
+            } else {
+                return Err(VirtualMachineError::ExpectedInteger(value_addr.clone()));
+            };
+            //Main logic (assert value == 0)
+            if *value != bigint!(0) {
```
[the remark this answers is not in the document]
fmoletta: Fixed!

Oppen (approved):

PR #267, opened 2022-07-12
Add implementation for hint on sqrt (math.cairo)

fmoletta:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/math_utils.rs
+++ b/src/math_utils.rs
@@ -1,7 +1,26 @@
-use crate::bigint;
+use crate::{bigint, vm::errors::vm_errors::VirtualMachineError};
 use num_bigint::BigInt;
 use num_integer::Integer;
-use num_traits::{abs, FromPrimitive};
+use num_traits::{abs, FromPrimitive, Signed};
+
+///Returns the integer square root of the nonnegative integer n.
+///This is the floor of the exact square root of n.
+///Unlike math.sqrt(), this function doesn't have rounding error issues.
+pub fn isqrt(n: &BigInt) -> Result<BigInt, VirtualMachineError> {
+    if !n.is_negative() {
+        return Err(VirtualMachineError::SqrtNegative(n.clone()));
+    }
+    let mut x = n.clone();
+    let mut y = (x.clone() + bigint!(1)).mod_floor(&bigint!(2));
+    while y < x.clone() {
+        x = y;
+        y = (x.clone() + n.mod_floor(&x)).mod_floor(&bigint!(2))
+    }
+    if x.pow(2) <= *n && *n < (x.clone() + bigint!(1)).pow(2) {
+        return Err(VirtualMachineError::FailedToGetExactSqrt(n.clone()));
+    };
+    Ok(x)
+}
 
 ///Returns x, y, g such that g = x*a + y*b = gcd(a, b).
 fn igcdex(num_a: BigInt, num_b: BigInt) -> (BigInt, BigInt, BigInt) {
--- a/src/vm/errors/vm_errors.rs
+++ b/src/vm/errors/vm_errors.rs
@@ -45,6 +45,8 @@ pub enum VirtualMachineError {
     AssertNotEqualFail(MaybeRelocatable, MaybeRelocatable),
     DiffIndexComp(Relocatable, Relocatable),
     ValueOutside250BitRange(BigInt),
+    SqrtNegative(BigInt),
+    FailedToGetExactSqrt(BigInt),
 }
 
 impl fmt::Display for VirtualMachineError {
@@ -131,6 +133,8 @@ impl fmt::Display for VirtualMachineError {
                 write!(f, "Failed to compare {:?} and  {:?}, cant compare two relocatable values of different segment indexes", a, b)
             },
             VirtualMachineError::ValueOutside250BitRange(value) => write!(f, "Value: {:?} is outside of the range [0, 2**250)", value),
+            VirtualMachineError::SqrtNegative(value) => write!(f, "Cant calculate the square root of negative number: {:?})", value),
+            VirtualMachineError::FailedToGetExactSqrt(value) => write!(f, "Failed to calculate the exact square root of: {:?})", value),
         }
     }
 }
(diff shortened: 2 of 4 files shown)


src/math_utils.rs
```diff
@@ -1,7 +1,29 @@
...
+    if !n.is_positive() {
+        return Err(VirtualMachineError::SqrtNegative(n.clone()));
+    }
+    let mut x = n.clone();
+    let mut y = (x.clone() + bigint!(1)).shr(1_i32);
+    while y < x.clone() {
+        x = y;
+        y = (x.clone() + n.div_floor(&x)).shr(1_i32);
+    }
+    if !(x.pow(2) <= *n && *n < (x.clone() + bigint!(1)).pow(2)) {
```
Oppen, line 22: I think we don't need any of the `x.clone()`s. Only `n` needs to be cloned due to it being a ref.

In any case, it may be better to call `n.sqrt()`. It's defined for `BigInt`.

src/math_utils.rs
```diff
@@ -1,7 +1,29 @@
...
+        return Err(VirtualMachineError::FailedToGetSqrt(n.clone()));
+    };
+    Ok(x)
```
Oppen, line 25:
```suggestion
    Ok(n.sqrt())
```

src/math_utils.rs
```diff
@@ -1,7 +1,29 @@
...
 use num_integer::Integer;
-use num_traits::{abs, FromPrimitive};
+use num_traits::{abs, FromPrimitive, Signed};
+
+///Returns the integer square root of the nonnegative integer n.
+///This is the floor of the exact square root of n.
+///Unlike math.sqrt(), this function doesn't have rounding error issues.
+pub fn isqrt(n: &BigInt) -> Result<BigInt, VirtualMachineError> {
+    //n.shr(1) = n.div_floor(2)
+    if !n.is_positive() {
```
Oppen, line 13: We probably want:
```suggestion
    if n.is_negative() {
```
`sqrt` is well defined for `0` (it's itself).

src/vm/errors/vm_errors.rs
```diff
@@ -129,6 +132,9 @@ impl fmt::Display for VirtualMachineError {
             VirtualMachineError::DiffIndexComp(a, b) => {
                 write!(f, "Failed to compare {:?} and  {:?}, cant compare two relocatable values of different segment indexes", a, b)
             },
+            VirtualMachineError::ValueOutside250BitRange(value) => write!(f, "Value: {:?} is outside of the range [0, 2**250)", value),
+            VirtualMachineError::SqrtNegative(value) => write!(f, "Cant calculate the square root of negative or zero number: {:?})", value),
```
Oppen, line 136:
```suggestion
            VirtualMachineError::SqrtNegative(value) => write!(f, "Can't calculate the square root of negative or zero number: {:?})", value),
```

Oppen, line 136: Otherwise, if we still consider `0` to not have a square root:
```suggestion
            VirtualMachineError::SqrtNegative(value) => write!(f, "Can't calculate the square root of non-positive number: {:?})", value),
```

src/vm/hints/hint_utils.rs
```diff
@@ -407,3 +408,55 @@ pub fn assert_nn(
...
+        (Ok(Some(maybe_rel_value)), Ok(_)) => {
+            let value = if let MaybeRelocatable::Int(value) = maybe_rel_value {
+                value
+            } else {
+                return Err(VirtualMachineError::ExpectedInteger(
+                    maybe_rel_value.clone(),
+                ));
+            };
+            let mod_value = value.mod_floor(&vm.prime);
+            if mod_value > bigint!(2).pow(250) {
```
Oppen, line 453: Much cheaper way to compute powers of `2`:
```suggestion
            if mod_value > bigint!(1).shl(250) {
```

Oppen: Or avoiding the creation of an object:
```suggestion
            if !mod_value.shr(250).is_zero() {
```
This exploits the fact that, if `a > b` then `a / b >= 1 > 0`. A right shift by `n` is equivalent to perform `.div_floor` by `2**n`. So, if `mod_value <= 2**250` then `mod_value.shr(250).is_zero()` holds. Negating that we negate that it's greater.

Oppen: (Always in the real of positive numbers of course)

src/math_utils.rs
fmoletta: I could remove the clone on the comparison, but removing the rest of the clones() in x is not possible (at least not without further changes)

Oppen: Why not?

fmoletta: The hint specifically uses this version of square root (there is this version and a normal square root in math_utils)

Oppen: I think that answers the other comment.

src/math_utils.rs
fmoletta: Fixed!

src/vm/errors/vm_errors.rs
fmoletta: Fixed!

fmoletta: Chose previous solution

src/vm/hints/hint_utils.rs
```diff
@@ -396,7 +397,7 @@ pub fn assert_nn(
             };
 
             // assert 0 <= ids.a % PRIME < range_check_builtin.bound
-            if a.mod_floor(&vm.prime).is_positive()
+            if !a.mod_floor(&vm.prime).is_negative()
```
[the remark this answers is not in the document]
fmoletta: The method .is_positive() doesn't take into account 0 as a positive number, that's why it was changed. Removed comparison

src/vm/hints/hint_utils.rs
fmoletta: In the second version, the shr takes ownership of mod_value, which is used later (which means a clone is needed somewhere). Applied first option

Oppen (approved):

PR #289, opened 2022-07-18
[HINTS] Set add hint

Juan-M-V:
# Set add hint implementation

## Description
Implements hint set_add: https://github.com/starkware-libs/cairo-lang/blob/2abd303e1808612b724bc1412b2b5babd04bb4e7/src/starkware/cairo/common/set.cairo#L34-L45

## Checklist
- [x] Unit tests added


src/vm/vm_memory/memory.rs
```diff
@@ -113,6 +113,20 @@ impl Memory {
...
+        size: usize,
+    ) -> Result<Vec<Option<&MaybeRelocatable>>, MemoryError> {
+        let mut values = Vec::new();
+
+        for i in 0..size {
+            values.push(self.get(&addr.add_usize_mod(i, None))?);
+        }
+
+        Ok(values)
+    }
```
Oppen, line 129: I'd much rather have an iterator return elements one by one. But even if we keep it this way, you should probably do it like this:
```suggestion
    pub fn get_range(
        &self,
        addr: &MaybeRelocatable,
        size: usize,
    ) -> Result<Vec<Option<&MaybeRelocatable>>, MemoryError> {
        let addr = if let RelocatableValue(addr) = addr {
            addr
        } else {
            return // some error; we'll write a helper eventually
        }
        let segment = self.data[addr.segment_index];
        let (i, n) = (addr.offset, addr.offset + size);
        if n > segment.len() {
            return // some error
        }
        Ok(Vec::from(segment[i..n]));
    }
```
It's fine to access `data` from within the `memory` module. The problem is only if we do it from the outside.

src/vm/hints/set.rs
```diff
@@ -139,80 +39,72 @@ pub fn set_add(
...
+                .as_any()
+                .downcast_ref::<RangeCheckBuiltinRunner>()
+                .ok_or(VirtualMachineError::NoRangeCheckBuiltin)?;
+            let elm_size = if let MaybeRelocatable::Int(ref elm_size) = maybe_rel_elm_size {
+                if elm_size.is_zero() {
+                    return Err(VirtualMachineError::ValueNotPositive(elm_size.clone()));
+                }
+                elm_size
+                    .to_usize()
+                    .ok_or_else(|| VirtualMachineError::ValueNotPositive(elm_size.clone()))?
```
Oppen, line 57: This could also mean you passed a negative number.

src/vm/hints/set.rs
```diff
@@ -139,80 +39,72 @@ pub fn set_add(
...
+                    set_end_ptr.clone(),
+                ));
             }
-            Err(VirtualMachineError::NoRangeCheckBuiltin)
+
+            let set_span = set_end_ptr.sub(set_ptr, &vm.prime)?;
+            // sub method always returns a MaybeRelocatable::Int
+            let range_limit = if let MaybeRelocatable::Int(ref range_limit) = set_span {
+                range_limit
+                    .to_i32()
```
Oppen, line 80: Shouldn't this be `to_usize`? Either that or you're returning the wrong error.

src/vm/hints/set.rs
Juan-M-V: I can't find the code but to_usize [returns None](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3ca52cc031a19f887154c56a6f2c57ab) if the given BigInt is negative

src/vm/hints/set.rs
Juan-M-V: resolved: https://github.com/lambdaclass/cleopatra_cairo/pull/289/commits/6463e1761b015ebf6c0c867d02673a7aaf105dda

src/vm/vm_memory/memory.rs
Juan-M-V: I can't resolve this trivially. Do you think I can change this in a different PR and make a proper iterator?

Oppen: OK.

src/vm/hints/set.rs
Oppen: Exactly. But it also returns `None` if it's greater than `usize::MAX`, [doesn't it](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3ca52cc031a19f887154c56a6f2c57ab)?
The error message is not considering that case.

Juan-M-V: resolved: https://github.com/lambdaclass/cleopatra_cairo/pull/289/commits/4f77a8115240fada313453bcdccd79776a81675b

Oppen: I meant "too big" :facepalm:

src/vm/hints/set.rs
```diff
@@ -0,0 +1,460 @@
...
+            )
+            .expect("Unexpected memory insert fail");
+        vm.memory
+            .insert(
+                &MaybeRelocatable::from((2, 0)),
+                if let Some(rel) = elm_a {
+                    rel
+                } else {
+                    &elm_a_default
+                },
```
Oppen, line 215: You can use https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or

Oppen:
LGTM. Just some minor refactor by using [unwrap_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or) for tests and it's good to go.

src/vm/hints/set.rs
Juan-M-V: resolve: https://github.com/lambdaclass/cleopatra_cairo/pull/289/commits/0f008d1c33335380eb9dc3b7f52a9aededd12905

Oppen (approved):

PR #295, opened 2022-07-18
Add support for ap tracking

entropidelic:
# Add support for ap tracking

## Description

Some references are referenced respect to `ap`, which can change while the program is executing. AP tracking data from the compiled program json is used for adjusting the value of the reference. This PR adds support for these cases.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/hints/hint_utils.rs
+++ b/src/vm/hints/hint_utils.rs
@@ -1,6 +1,7 @@
 use crate::bigint;
 use crate::math_utils::as_int;
 use crate::math_utils::isqrt;
+use crate::serde::deserialize_program::ApTracking;
 use crate::types::{instruction::Register, relocatable::MaybeRelocatable};
 use crate::vm::{
     context::run_context::RunContext, errors::vm_errors::VirtualMachineError,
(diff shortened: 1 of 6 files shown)


Juan-M-V (approved):

src/vm/hints/hint_utils.rs
```diff
@@ -13,29 +14,63 @@ use num_traits::{FromPrimitive, Signed, ToPrimitive, Zero};
 use std::collections::HashMap;
 use std::ops::{Neg, Shl, Shr};
 
+fn apply_ap_tracking_correction(
+    ap: MaybeRelocatable,
+    ref_ap_tracking: Option<&ApTracking>,
+    hint_ap_tracking: Option<&ApTracking>,
```
Oppen, line 20: We may check this outside and not call the function instead of using `Option`.

src/vm/hints/hint_utils.rs
```diff
@@ -13,29 +14,63 @@ use num_traits::{FromPrimitive, Signed, ToPrimitive, Zero};
...
+        }
+        if let MaybeRelocatable::RelocatableValue(relocatable) = ap {
+            let ap_diff = hint_tracking_data.offset - ref_tracking_data.offset;
+
+            Ok(MaybeRelocatable::from((
+                relocatable.segment_index,
+                relocatable.offset - ap_diff,
+            )))
+        } else {
+            Err(VirtualMachineError::InvalidApValue(ap))
```
Oppen, line 39: Is this an error or does revoking a reference (change in `group`) simply mean `ap` is a new address? We need to check this I think.

src/vm/hints/hint_utils.rs
```diff
@@ -13,29 +14,63 @@ use num_traits::{FromPrimitive, Signed, ToPrimitive, Zero};
...
+        if ref_tracking_data.group != hint_tracking_data.group {
+            return Err(VirtualMachineError::InvalidTrackingGroup(
+                ref_tracking_data.group,
+                hint_tracking_data.group,
+            ));
+        }
+        if let MaybeRelocatable::RelocatableValue(relocatable) = ap {
+            let ap_diff = hint_tracking_data.offset - ref_tracking_data.offset;
+
+            Ok(MaybeRelocatable::from((
```
Oppen, line 34:
```suggestion
            Ok(Relocatable::from((
```

src/vm/hints/hint_utils.rs
entropidelic: this error raises if the `ap` register is not holding a memory address, which would be an unexpected error

Oppen: Aren't addresses the only thing `ap` (as register) can contain?

src/vm/hints/hint_utils.rs
entropidelic: I think this change would break the hints code, maybe we could leave it as a `MaybeRelocatable`?

PR #301, opened 2022-07-19
[HINTS] Add implementation for hints on squash_dict.cairo

fmoletta:
### squash_dict_inner

**squash_dict_inner_first_iteration**
Hint Code:
```
        current_access_indices = sorted(access_indices[key])[::-1]
        current_access_index = current_access_indices.pop()
        memory[ids.range_check_ptr] = current_access_index`
``` 
- [x] Implementation
- [x] Unit tests

**squash_dict_inner_skip_loop**
Hint Code: `ids.should_skip_loop = 0 if current_access_indices else 1 `
- [x] Implementation
- [x] Unit tests

**squash_dict_inner_check_access_index**
Hint Code: 
```
        new_access_index = current_access_indices.pop()
        ids.loop_temps.index_delta_minus1 = new_access_index - current_access_index - 1
        current_access_index = new_access_index
```
- [x] Implementation
- [x] Unit tests

**squash_dict_inner_continue_loop**
Hint Code: `ids.loop_temps.should_continue = 1 if current_access_indices else 0 `
- [x] Implementation
- [x] Unit tests

**squash_dict_inner_len_assert**
Hint Code: `assert len(current_access_indices) == 0` 
- [x] Implementation
- [x] Unit tests

**squash_dict_inner_used_accesses_assert**
Hint Code: `assert ids.n_used_accesses == len(access_indices[key])`
- [x] Implementation
- [x] Unit tests

**squash_dict_inner_assert_len_keys**
`%{ assert len(keys) == 0 %}`
- [x] Implementation
- [x] Unit tests

**squash_dict_inner_next_key**
Hint Code: 
```
        assert len(keys) > 0, 'No keys left but remaining_accesses > 0.'
        ids.next_key = key = keys.pop()
```
- [x] Implementation
- [x] Unit tests

### squash_dict

 **enter_scope**
Hint Code: `vm_enter_scope()`
Already implemented

**exit_scope**
Hint Code: `vm_exit_scope()`
- [x] Implementation
- [x] Unit tests

**squash_dict_outer**
Hint Code:
```
        dict_access_size = ids.DictAccess.SIZE
        address = ids.dict_accesses.address_
        assert ids.ptr_diff % dict_access_size == 0, \
            'Accesses array size must be divisible by DictAccess.SIZE'
        n_accesses = ids.n_accesses
        if '__squash_dict_max_size' in globals():
            assert n_accesses <= __squash_dict_max_size, \
                f'squash_dict() can only be used with n_accesses<={__squash_dict_max_size}. ' \
                f'Got: n_accesses={n_accesses}.'
        # A map from key to the list of indices accessing it.
        access_indices = {}
        for i in range(n_accesses):
            key = memory[address + dict_access_size * i]
            access_indices.setdefault(key, []).append(i)
        # Descending list of keys.
        keys = sorted(access_indices.keys(), reverse=True)
        # Are the keys used bigger than range_check bound.
        ids.big_keys = 1 if keys[0] >= range_check_builtin.bound else 0
        ids.first_key = key = keys.pop()
```
- [x] Implementation
- [x] Unit tests

- [x] Integration test

Extra stuff:

- Added helper functions to hint_utils: get_range_check_builtin, get_int_from_scope and get_list_from_scope
- Replaced ids with reference in helper function get_variable_address_from_name and edited affected functions (This was done to avoid unneeded cloning of ids when calling this function more than once)


src/vm/hints/hint_utils.rs
```diff
@@ -16,6 +16,43 @@ use num_traits::{FromPrimitive, Signed, ToPrimitive, Zero};
...
+    }
+    val
+}
+
+//Returns the value in the current execution scope that matches the name and is of type List
+pub fn get_list_from_scope(vm: &mut VirtualMachine, name: &str) -> Option<Vec<BigInt>> {
+    let mut val: Option<Vec<BigInt>> = None;
+    if let Some(variables) = vm.exec_scopes.get_local_variables() {
+        if let Some(PyValueType::List(py_val)) = variables.get(name) {
+            val = Some(py_val.clone());
```
Oppen, line 35: I'm a bit worried about the perf of cloning a list of `BigInt`s. Would it be possible to return a slice?

src/vm/hints/squash_dict_utils.rs
```diff
@@ -0,0 +1,2485 @@
...
+    let mut current_access_indices = access_indices
+        .get(&key)
+        .ok_or_else(|| VirtualMachineError::NoKeyInAccessIndices(key.clone()))?
+        .clone();
+    current_access_indices.sort();
+    current_access_indices.reverse();
+    //Get current_access_index
+    let first_val = current_access_indices
+        .pop()
+        .ok_or(VirtualMachineError::EmptyCurrentAccessIndices)?;
```
Oppen, line 65: We could just keep a heap in the scope, since we're extracting one item per iteration. Right now this is `O(n^2logn)`, while the heap makes it `O(nlogn)`. The semantics of the hint are still preserved.

Oppen: Or even create a heap locally, at least that's locally linear and ultimately quadratic.

src/vm/hints/squash_dict_utils.rs
```diff
@@ -0,0 +1,2485 @@
...
+};
+
+use super::{
+    dict_hint_utils::DICT_ACCESS_SIZE,
+    hint_utils::{
+        get_address_from_var_name, get_int_from_scope, get_list_from_scope, get_range_check_builtin,
+    },
+};
+
+fn get_access_indices(vm: &mut VirtualMachine) -> Option<HashMap<BigInt, Vec<BigInt>>> {
```
Oppen, line 20: I think we should be able to use `u64` for indices. We can't have a dict bigger than addressable memory.

src/vm/hints/squash_dict_utils.rs
```diff
@@ -0,0 +1,2485 @@
...
+        .memory
+        .get(&dict_accesses_addr)
+        .map_err(VirtualMachineError::MemoryError)?
+        .ok_or_else(|| VirtualMachineError::MemoryGet(n_accesses_addr.clone()))?
+        .clone();
+    //Get range_check_builtin
+    let range_check_builtin = get_range_check_builtin(vm)?;
+    let range_check_bound = range_check_builtin._bound.clone();
+    //Main Logic
+    if ptr_diff % DICT_ACCESS_SIZE != bigint!(0) {
```
Oppen, line 347:
```suggestion
    if !(ptr_diff % DICT_ACCESS_SIZE).is_zero() {
```

src/vm/hints/squash_dict_utils.rs
```diff
@@ -0,0 +1,2485 @@
...
+            ));
+        };
+    };
+    let n_accesses_usize = n_accesses
+        .to_usize()
+        .ok_or(VirtualMachineError::NAccessesTooBig(n_accesses))?;
+    //A map from key to the list of indices accessing it.
+    let mut access_indices = HashMap::<BigInt, Vec<BigInt>>::new();
+    for i in 0..n_accesses_usize {
+        let key_addr = address.add_int_mod(&(DICT_ACCESS_SIZE * bigintusize!(i)), &vm.prime)?;
```
Oppen, line 364: This will be much more efficient than multiplication:
```suggestion
    let mut pos = 0;
    for i in 0..n_accesses_usize {
        let key_addr = address.add_int_mod(&bigint!(pos), &vm.prime)?;
        pos += DICT_ACCESS_SIZE;
```

Oppen (approved):
Very valuable addition. There are some places where performance could be less than ideal, but I think it's worth merging as it is. I'll take a look at performance soon after merging.

PR #303, opened 2022-07-19
[HINTS] Add implementation for memset on pow (memset.cairo)

pefontana:
# Add implementation for memset on pow (memset.cairo)

## Description

https://github.com/starkware-libs/cairo-lang/blob/167b28bcd940fd25ea3816204fa882a0b0a49603/src/starkware/cairo/common/memset.cairo

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/cairo_programs/memset.cairo
+++ b/cairo_programs/memset.cairo
@@ -1,0 +1,15 @@
+from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.memset import memset
+
+func main():
+    alloc_locals
+    let (local string1 : felt*) = alloc()
+    memset(string1,'I dont know why',20)
+    let (local string2 : felt*) = alloc()
+    memset(string2,'Do you know why',20)
+
+    assert string1[19] = 'I dont know why'
+    assert string2[1] = 'Do you know why'
+
+return ()
+end
--- a/src/vm/hints/execute_hint.rs
+++ b/src/vm/hints/execute_hint.rs
@@ -14,6 +14,7 @@ use crate::vm::hints::hint_utils::{
     memcpy_continue_copying, memcpy_enter_scope, signed_div_rem, split_felt, split_int,
     split_int_assert_range, sqrt, unsigned_div_rem,
 };
+use crate::vm::hints::memset_utils::{memset_continue_loop, memset_enter_scope};
 use crate::vm::vm_core::VirtualMachine;
 
 #[derive(Debug, PartialEq, Clone)]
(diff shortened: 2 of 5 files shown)


cairo_programs/memset.cairo
```diff
@@ -0,0 +1,19 @@
+from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.memset import memset
+
+func main():
+    alloc_locals
+    let (local strings : felt*) = alloc()
+    memset(strings,'Lambda',20)
+    assert strings[0] = 'Lambda'
+    assert strings[19] = 'Lambda'
```
Oppen, line 9: We can use compare arrays to make sure it's true for all elements.

cairo_programs/memset.cairo
```diff
@@ -0,0 +1,19 @@
...
+    assert strings[20] = 'can insert new value'
+
+    let numbers : felt* = alloc()
+    memset(numbers, 10, 100)
+    assert numbers[0] = 10
+    assert numbers[99] = 10
```
Oppen, line 15: Here as well.

cairo_programs/memset.cairo
Oppen: (An analog really, but same essence)

src/vm/hints/memset_utils.rs
```diff
@@ -0,0 +1,396 @@
...
+    let n = match vm.exec_scopes.get_local_variables() {
+        Some(variables) => match variables.get("n") {
+            Some(PyValueType::BigInt(n)) => n,
+            _ => {
+                return Err(VirtualMachineError::VariableNotInScopeError(String::from(
+                    "n",
+                )))
+            }
+        },
+        None => return Err(VirtualMachineError::ScopeError),
```
Oppen, line 63: I'd collapse these two to return only `VariableNotInScope` instead.

src/vm/hints/memset_utils.rs
```diff
@@ -0,0 +1,396 @@
...
+    // else, insert 0
+    if new_n.is_positive() {
+        vm.memory
+            .insert(&continue_loop_addr, &MaybeRelocatable::Int(bigint!(1)))
+            .map_err(VirtualMachineError::MemoryError)?;
+    } else {
+        vm.memory
+            .insert(&continue_loop_addr, &MaybeRelocatable::Int(bigint!(0)))
+            .map_err(VirtualMachineError::MemoryError)?;
+    }
```
Oppen, line 79:
```suggestion
    // if `new_n` is positive, insert 1 in the address of `continue_loop`
    // else, insert 0
    let continue = new_n.is_positive() as i32;
    vm.memory
        .insert(&continue_loop_addr, &MaybeRelocatable::Int(bigint!(continue)))
        .map_err(VirtualMachineError::MemoryError)?;
```

src/vm/hints/memset_utils.rs
pefontana: I couldn't collapse these two errors.
Since `get_local_variables` returns an option I always need to use two arms.
My best try was implementing this:
```
    // get `n` variable from vm scope
    let n = match vm.exec_scopes.get_local_variables()?.get("n") {
        Some(PyValueType::BigInt(n)) => n,
        _ => {
            return Err(VirtualMachineError::VariableNotInScopeError(String::from(
                "n",
            )))
        }
    };
```

But I got the following error:
`the `?` operator can only be used on `Result`s, not `Option`s, in a function that returns `Result``

cairo_programs/memset.cairo
pefontana: [Done!](https://github.com/lambdaclass/cleopatra_cairo/pull/303/commits/6e106e036ff92c6cd09ef6e561fc7950f9ae8e4c)

cairo_programs/memset.cairo
pefontana: [Done!](https://github.com/lambdaclass/cleopatra_cairo/pull/303/commits/6e106e036ff92c6cd09ef6e561fc7950f9ae8e4c)

src/vm/hints/memset_utils.rs
Oppen: `Option`s support the `ok_or` method to convert to a `Result`, where the argument is the error it should report if `None`. Then you can use `?` on the result.

pefontana: :party_wizard
[Done!](https://github.com/lambdaclass/cleopatra_cairo/pull/303/commits/c18877ad20e37e07f60be40bf0d57a0a7167d11b)

src/vm/hints/memset_utils.rs
```diff
@@ -51,32 +51,31 @@ pub fn memset_continue_loop(
...
-            }
-        },
-        None => return Err(VirtualMachineError::ScopeError),
+
+    // get `n` variable from vm scope
+    let n = match vm
+        .exec_scopes
+        .get_local_variables()
+        .ok_or(VirtualMachineError::ScopeError)?
+        .get("n")
```
Oppen, line 60: `.ok_or_else(|_| VirtualMachineError::VariableNotInScopeError(String::from("n")))`

Oppen (approved):

PR #306, opened 2022-07-21
Add nom parsers for references deserialization

entropidelic:
# Nom references parsing

## Description
This PR replaces the manual parsing implemented for values of references in the compiled json cairo programs. For the review, this link could be helpful:
https://github.com/Geal/nom/blob/main/doc/choosing_a_combinator.md

Essentially, the same functionality as the manual parser has been achieved, but adjusting the parser is now much easier thanks to the composability of nom parsers.
A default value for ValueAddress has been set when the parsing throws an error. This is done to avoid unnecessary errors in the deserialization process. The string formats of values used by hints is well supported, and if a new format appears, an error (`VirtualMachine::IdNotFound`) will make it clear that a new string format should be handled.

## Checklist
- [x] Unit tests added

### DIFF
--- a/src/serde/deserialize_utils.rs
+++ b/src/serde/deserialize_utils.rs
@@ -268,12 +268,12 @@ mod tests {
     }
 
     #[test]
-    fn parse_value_test() {
-        let value_1 = "[cast([fp + (-1)] + 2, felt*)]";
-        let parsed_1 = parse_value(value_1);
+    fn parse_value_with_inner_dereference_test() {
+        let value = "[cast([fp + (-1)] + 2, felt*)]";
+        let parsed = parse_value(value);
 
         assert_eq!(
-            parsed_1,
+            parsed,
             Ok((
                 "",
                 ValueAddress {
@@ -286,12 +286,15 @@ mod tests {
                 }
             ))
         );
+    }
 
-        let value_2 = "cast(ap + 2, felt*)";
-        let parsed_2 = parse_value(value_2);
+    #[test]
+    fn parse_value_with_no_inner_dereference_test() {
+        let value = "cast(ap + 2, felt*)";
+        let parsed = parse_value(value);
 
         assert_eq!(
-            parsed_2,
+            parsed,
             Ok((
                 "",
                 ValueAddress {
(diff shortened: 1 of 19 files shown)


Oppen (approved):
LGTM

src/vm/hints/hint_utils.rs
```diff
@@ -168,7 +168,8 @@ pub fn get_address_from_var_name(
 ) -> Result<MaybeRelocatable, VirtualMachineError> {
     let var_ref = ids
         .get(&String::from(var_name))
-        .ok_or(VirtualMachineError::FailedToGetIds)?;
+        .ok_or_else(|| VirtualMachineError::IdNotFound(var_name.to_string()))?;
+    // .ok_or(VirtualMachineError::IdNotFound(var_name.to_string()))?;
```
igaray, line 172: Rememer to remove the commented line before merging

src/serde/deserialize_utils.rs
```diff
@@ -229,135 +196,130 @@ mod tests {
...
-            offset2: 0,
-            immediate: None,
-            dereference: false,
-            inner_dereference: false,
-        };
+        assert_eq!(parsed, Ok(("", (false, Register::AP, 3_i32))));
+    }
 
-        assert_eq!(value_address, parsed_value);
+    #[test]
```
igaray, line 272: These tests are good but since they are completely independent I would have them in three separate test functions.

src/serde/deserialize_utils.rs
```diff
@@ -41,185 +54,139 @@ pub fn maybe_add_padding(mut hex: String) -> String {
     hex
 }
 
-fn parse_register(splitted_value_str: &[&str]) -> Option<Register> {
-    let str_tmp: Vec<&str> = splitted_value_str[0].split(',').collect();
-
-    let mut raw_reg_str = str_tmp[0].to_string();
-
-    raw_reg_str.retain(|c| !r#"["#.contains(c));
+/* NOM PARSERS */
```
igaray, line 57: Could you document the grammar being parsed (just an overview) in the dev docs and link to the documentation in the comments at the beginning of the parser?

igaray:
The parser code is tight and the tests look good, I would only request:
1. The grammer parsed be documented in the dev docs
2. The docs be linked to in the relevent code.

igaray (changes requested):

src/serde/deserialize_utils.rs
Oppen: Will be done afterwards #340

Oppen (approved):

PR #317, opened 2022-07-26
[HINTS]Add implementation for hints & utils for function blake2s (cairo_blake2s)

fmoletta:
- [x] blake2s in rust
- [x] unit tests for blake2s
- [x] hint implementation
- [x] unit tests for hint
- [x] integration test


cairo_programs/blake2s_hello_world_hash.cairo
```diff
@@ -0,0 +1,20 @@
...
+from starkware.cairo.common.cairo_blake2s.blake2s import blake2s
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+
+#Computes the hash of "Hello World"
+func main{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}():
+    alloc_locals
+    let inputs: felt* = alloc()
+    assert inputs[0] = 1819043144
+    assert inputs[1] = 1870078063
+    assert inputs[2] = 653255322
```
Oppen, line 13: Cairo supports expressing this as short string literals up to 31 characters.

src/vm/hints/blake2s_hash.rs
```diff
@@ -0,0 +1,335 @@
...
+    [9, 0, 5, 7, 2, 4, 10, 15, 14, 1, 11, 12, 6, 8, 3, 13],
+    [2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9],
+    [12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11],
+    [13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10],
+    [6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5],
+    [10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0],
+];
+
+fn right_rot(value: u64, n: u64) -> u64 {
+    (value >> n) | ((value & (2_u64.pow(n as u32) - 1)) << (32 - n))
```
Oppen, line 19:
```suggestion
    (value >> n) | ((value & u64::MAX) << (32 - n))
```

src/vm/hints/blake2s_hash.rs
```diff
@@ -0,0 +1,335 @@
...
+
+fn mix(a: u64, b: u64, c: u64, d: u64, m0: u64, m1: u64) -> (u64, u64, u64, u64) {
+    let a = (a + b + m0) % 2_u64.pow(32);
+    let d = right_rot(d ^ a, 16);
+    let c = (c + d) % 2_u64.pow(32);
+    let b = right_rot(b ^ c, 12);
+    let a = (a + b + m1) % 2_u64.pow(32);
+    let d = right_rot(d ^ a, 8);
+    let c = (c + d) % 2_u64.pow(32);
+    let b = right_rot(b ^ c, 7);
```
Oppen, line 30:
```suggestion
    let a = (a + b + m0) & u32::MAX as u64;
    let d = right_rot(d ^ a, 16);
    let c = (c + d) & u32::MAX as u64;
    let b = right_rot(b ^ c, 12);
    let a = (a + b + m1) & u32::MAX as u64;
    let d = right_rot(d ^ a, 8);
    let c = (c + d) & u32::MAX as u64;
    let b = right_rot(b ^ c, 7);
```

Oppen: As an explanation for the suggested patch: binary and with `2^n - 1` is the same as modulus with `2^n` for unsigned arithmetic. In general, dividing by a power of your base is shifting places, and the remainder is necessarily what you would lose after that shift, that is whatever is in a position to the left of the power itself. That set to all `1` is also `2^n - 1`, which is the maximum number you can represent with `n` bits.
So, `x % 2.pow(32) == x & (2.pow(32) - 1) == x & u32::MAX`.

Oppen: `&` with a constant is much more efficient for the computer to perform than division.

src/vm/hints/blake2s_hash.rs
```diff
@@ -0,0 +1,335 @@
...
+        state[14],
+        message[sigma[14]],
+        message[sigma[15]],
+    );
+    state
+}
+
+pub fn blake2s_compress(
+    h: [u64; 8],
+    message: [u64; 16],
```
Oppen, line 104:
```suggestion
    h: &[u64; 8],
    message: &[u64; 16],
```

cairo_programs/blake2s_hello_world_hash.cairo
fmoletta: Applied change!

src/vm/hints/blake2s_utils.rs
```diff
@@ -0,0 +1,338 @@
...
+    vm::{
+        errors::vm_errors::VirtualMachineError,
+        vm_memory::{memory::Memory, memory_segments::MemorySegmentManager},
+    },
+};
+use num_bigint::BigInt;
+use num_traits::FromPrimitive;
+
+fn get_fixed_size_u64_array<const T: usize>(
+    h_range: Vec<Option<&MaybeRelocatable>>,
```
Oppen, line 22:
```suggestion
    h_range: &Vec<Option<&MaybeRelocatable>>,
```

src/vm/hints/blake2s_utils.rs
```diff
@@ -0,0 +1,338 @@
...
+) -> Result<[u64; T], VirtualMachineError> {
+    let mut u64_vec = Vec::<u64>::new();
```
Oppen, line 24:
```suggestion
    let mut u64_vec = Vec::<u64>::with_capacity(h_range.len());
```

src/vm/hints/blake2s_utils.rs
```diff
@@ -0,0 +1,338 @@
...
+            return Err(VirtualMachineError::ExpectedInteger(mayberel.clone()));
+        };
+        u64_vec.push(num.to_u64().ok_or(VirtualMachineError::BigintToU64Fail)?);
+    }
+    u64_vec
+        .try_into()
+        .map_err(|_| VirtualMachineError::FixedSizeArrayFail(T))
+}
+
+fn get_maybe_relocatable_array_from_u64(array: Vec<u64>) -> Vec<MaybeRelocatable> {
```
Oppen, line 39:
```suggestion
fn get_maybe_relocatable_array_from_u64(array: &Vec<u64>) -> Vec<MaybeRelocatable> {
```

src/vm/hints/blake2s_hash.rs
fmoletta: This change affects the output of the hash

src/vm/hints/blake2s_utils.rs
```diff
@@ -0,0 +1,338 @@
...
+    let mut new_array = Vec::<MaybeRelocatable>::new();
```
Oppen, line 40:
```suggestion
    let mut new_array = Vec::<MaybeRelocatable>::with_capacity(array.len());
```

src/vm/hints/blake2s_hash.rs
fmoletta: Thanks for the explanation! applied change!

src/vm/vm_memory/memory.rs
```diff
@@ -100,6 +100,20 @@ impl Memory {
...
+        &self,
+        key: &MaybeRelocatable,
+    ) -> Result<&BigInt, VirtualMachineError> {
+        match self.get(key) {
+            Ok(Some(MaybeRelocatable::Int(int))) => Ok(int),
+            Ok(_) => Err(VirtualMachineError::ExpectedInteger(key.clone())),
+            Err(memory_error) => Err(VirtualMachineError::MemoryError(memory_error)),
+        }
+    }
+
```
Oppen, line 116: It's probably better if we just call it `as_int`.

src/vm/hints/blake2s_hash.rs
fmoletta: Applied change

src/vm/vm_memory/memory.rs
Oppen, line 116: It's probably better to just implement an `as_int` method for for `MaybeRelocatable`.
Also, let's just start using `Relocatable` for addresses. We can make `Memory::get` generic to allow for both temporarily I think.

src/vm/hints/blake2s_utils.rs
fmoletta: Applied change!

src/vm/hints/blake2s_utils.rs
fmoletta: Applied change!

src/vm/hints/blake2s_hash.rs
Oppen: Ohhh, I missed the `n` inside the `pow`. It should be:
```suggestion
    (value >> n) | ((value & (1.shl(n as u32) - 1)) << (32 - n))
```

src/vm/hints/blake2s_utils.rs
fmoletta: Applied change!

src/types/relocatable.rs
```diff
@@ -153,6 +153,33 @@ impl MaybeRelocatable {
...
+                }
+                let new_offset = rel.offset - other;
+                Ok(MaybeRelocatable::RelocatableValue(Relocatable {
+                    segment_index: rel.segment_index,
+                    offset: new_offset,
+                }))
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
fmoletta: I can make a smaller implementation for Relocatable that doesn't take prime, but using Sub is not possible as we don't want the vm to crash if we subtract with overflow, we would want to return an Error instead

src/vm/hints/blake2s_hash.rs
```diff
@@ -0,0 +1,335 @@
...
+    let a = (a + b + m0) & u32::MAX as u64;
+    let d = right_rot(d ^ a, 16);
+    let c = (c + d) & u32::MAX as u64;
+    let b = right_rot(b ^ c, 12);
+    let a = (a + b + m1) & u32::MAX as u64;
+    let d = right_rot(d ^ a, 8);
+    let c = (c + d) & u32::MAX as u64;
+    let b = right_rot(b ^ c, 7);
+    (a, b, c, d)
+}
```
Oppen, line 32: Now that I pay closed attention and compare with [the algorithm in Wikipedia](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#Algorithm), this is exactly like working on `u32` with `wrapping_add`, so it would look like:
```suggestion
fn mix(a: u32, b: u32, c: u32, d: u32, m0: u32, m1: u32) -> (u32, u32, u32, u32) {
    let a = a.wrapping_add(b).wrapping_add(m0);
    let d = right_rot(d ^ a, 16);
    let c = c.wrapping_add(d);
    let b = right_rot(b ^ c, 12);
    let a = a.wrapping_add(b).wrapping_add(m1);
    let d = right_rot(d ^ a, 8);
    let c = c.wrapping_add(d);
    let b = right_rot(b ^ c, 7);
    (a, b, c, d)
}
```

src/types/relocatable.rs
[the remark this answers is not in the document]
Oppen: We can use `checked_sub` then. That one can return an error.

src/vm/hints/blake2s_utils.rs
```diff
@@ -0,0 +1,343 @@
...
+    h_range: &Vec<Option<&MaybeRelocatable>>,
+) -> Result<[u64; T], VirtualMachineError> {
+    let mut u64_vec = Vec::<u64>::with_capacity(h_range.len());
+    for element in h_range {
+        let mayberel = element.ok_or(VirtualMachineError::UnexpectMemoryGap)?;
+        let num = if let MaybeRelocatable::Int(num) = mayberel {
+            num
+        } else {
+            return Err(VirtualMachineError::ExpectedInteger(mayberel.clone()));
+        };
```
Oppen, line 31: If we write the `.as_int` helper for `MaybeRelocatable` we could do:
```suggestion
        let num = element
            .ok_or(VirtualMachineError::UnexpectMemoryGap)?
            .as_int()?;
```

Oppen:
Quick question, is this up-to-date with main? I see the perf regression has timings corresponding to the previous, failing cases.

src/vm/hints/blake2s_hash.rs
fmoletta: It works!

src/types/relocatable.rs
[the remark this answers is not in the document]
fmoletta: checked_sub returns an option, so mapping to an error would be necessary each time it is used. I think its simpler to add `pub fn sub(&self, other: usize) -> Result<Self, VirtualMachineError>` for Relocatable, so we can propagate the error with ?

src/vm/vm_memory/memory.rs
fmoletta: After the previous refactors this function is no longer needed, and can be replaced with get_integer. I think making Memory::get generic is a good idea, but it exceeds the scope of this PR

src/vm/hints/blake2s_hash.rs
fmoletta: This change affects the output of the hash

src/vm/hints/blake2s_utils.rs
fmoletta: Added helper method `get_int_ref` to MaybeRelocatable

fmoletta:
> Quick question, is this up-to-date with main? I see the perf regression has timings corresponding to the previous, failing cases.

It should be no longer than  a couple days behind, I will update it now once I solve the conflicts

fmoletta:
> > Quick question, is this up-to-date with main? I see the perf regression has timings corresponding to the previous, failing cases.
> 
> It should be no longer than a couple days behind, I will update it now once I solve the conflicts

Branch updated!

src/vm/hints/blake2s_hash.rs
fmoletta: Applied change!

Oppen (approved):

PR #328, opened 2022-07-29
Add macros memory and generic version of bigint

fmoletta:
- Add macro BigInt as a generic converter to BigInt
- Replace all bigInt macros for a generic version (except bigint_str)
- Add macro memory, that creates a Memory from a List (to be used in tests)
- Add an example of use for memory macro in test  for blake2_utils::tests::compute_blake2s_output_input_bigger_than_u32

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -9,55 +9,6 @@ macro_rules! bigint {
 }
 
 #[macro_export]
-macro_rules! bigint32 {
-    ($val : expr) => {
-        BigInt::from_i32($val).unwrap()
-    };
-}
-
-#[macro_export]
-macro_rules! bigint64 {
-    ($val : expr) => {
-        BigInt::from_i64($val).unwrap()
-    };
-}
-
-#[macro_export]
-macro_rules! bigint_u32 {
-    ($val : expr) => {
-        BigInt::from_u32($val).unwrap()
-    };
-}
-
-#[macro_export]
-macro_rules! bigint_u64 {
-    ($val : expr) => {
-        BigInt::from_u64($val).unwrap()
-    };
-}
-
-#[macro_export]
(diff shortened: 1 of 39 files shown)


src/utils.rs
```diff
@@ -92,9 +48,73 @@ pub fn to_field_element(num: BigInt, prime: BigInt) -> BigInt {
...
+            }
+        };
+        ($mem:expr, ($si:expr, $off:expr), $val:expr) => {
+            let mut res = $mem.insert(&mayberelocatable!($si, $off), &mayberelocatable!($val));
+            while matches!(res, Err(MemoryError::UnallocatedSegment(_, _))) {
+                $mem.data.push(Vec::new());
+                res = $mem.insert(&mayberelocatable!($si, $off), &mayberelocatable!($val));
+            }
+        };
+    }
```
Oppen, line 99: We can make this code more brief by creating vars:
```suggestion
            let (k, v) = (&mayberelocatable!($si, $off), &mayberelocatable!($sival, $offval));
            let mut res = $mem.insert(k, v);
            while matches!(res, Err(MemoryError::UnallocatedSegment(_, _))) {
                $mem.data.push(Vec::new());
                res = $mem.insert(k, v);
            }
        };
        ($mem:expr, ($si:expr, $off:expr), $val:expr) => {
            let (k, v) = (&mayberelocatable!($si, $off), &mayberelocatable!($val));
            let mut res = $mem.insert(k, v);
            while matches!(res, Err(MemoryError::UnallocatedSegment(_, _))) {
                $mem.data.push(Vec::new());
                res = $mem.insert(k, v);
            }
        };
    }
```

Oppen (changes requested):

Oppen (approved):

PR #337, opened 2022-08-03
Separate VM from Hint Execution: Phase 1 (VMProxy)

fmoletta:
Major Changes:
Add structure VMProxy to hold references to data used by hints
Swap VM reference for VMProxy reference in execute_hint
Remove HintExecutor from VM (It is now held by CairoRunner and passes to VM in step())
Swap `insert_int_into_ap` for generic version` insert_value_into_ap`
Comments:
Some clones will be removed on later stages of this refactor, but are needed for now.


src/vm/hints/blake2s_utils.rs
```diff
@@ -109,17 +109,17 @@ pub fn compute_blake2s(
     segments.write_arg(ids.blake2s_ptr_end, padding)
 */
 pub fn finalize_blake2s(
-    vm: &mut VirtualMachine,
+    variables: HintVisibleVariables,
```
Oppen, line 112: I think a better name would be something like `VMProxy`. Also, a (mut?) reference may be more appropriate.

src/vm/hints/blake2s_utils.rs
```diff
@@ -71,19 +71,19 @@ fn compute_blake2s_func(
...
     let output = get_ptr_from_var_name(
         "output",
         ids,
-        &vm.memory,
-        &vm.references,
-        &vm.run_context,
+        variables.memory,
+        variables.references,
+        variables.run_context,
         hint_ap_tracking,
```
Oppen, line 84: This could receive the `variables` object directly, right?

src/vm/hints/blake2s_utils.rs
```diff
@@ -71,19 +71,19 @@ fn compute_blake2s_func(
    compute_blake2s_func(segments=segments, output_ptr=ids.output)
 */
 pub fn compute_blake2s(
-    vm: &mut VirtualMachine,
+    variables: HintVisibleVariables,
     ids: &HashMap<String, BigInt>,
     hint_ap_tracking: Option<&ApTracking>,
```
Oppen, line 76: We can probably include `ids` and `hint_ap_tracking` in the new structure.

Oppen (changes requested):

src/vm/hints/blake2s_utils.rs
fmoletta: Done!

src/vm/hints/blake2s_utils.rs
fmoletta: Done!

src/vm/hints/blake2s_utils.rs
fmoletta: Will get to this on a later phase of the refactor

igaray (approved):

PR #344, opened 2022-08-04
Add macros for vm, ids, references + example

fmoletta:
Add the following macros:

- `references`: Initializes a HintReference hasp given the number of references
- `ids`: Creates an ids dictionary from a list of names
- `vm_with_range_check`: Creates a VM with RangeCheckBuiltin
- `vm`: Creates a VM

Adds use example in dict_hint_utils.rs (Test `run_dict_read_valid`)


Oppen (approved):
Beautiful!

PR #345, opened 2022-08-04
Optimize build workflow

Oppen:
## Description

Build GitHub Action is taking real long due to a series of factors, much longer than even the benchmarks. This is an attempt to fix it.

This first commit just triggers the job for time measurement.


Oppen:
I won't be working on this until I'm back, so I'll close the PR and rework it after. It'll probably be superseded by #348 anyway.

PR #351, opened 2022-08-04
[HINTS] Add implementation of cairo_keccak/keccak hints

entropidelic:
# Hints from cairo_keccak hints

## Description
This PR adds support for all the hints in the `cairo_keccak` folder of the common library. The crate [keccak](https://crates.io/crates/keccak).
Cairo programs have been added for integration testing of the hints implemented.

NOTE: The `cairo_keccak_finalize` hint has no unit test because it has to be called only after calling keccak functions, so the integration test should be enough. 

## Checklist
- [x] Unit tests added


src/vm/hints/cairo_keccak/keccak_hints.rs
```diff
@@ -0,0 +1,414 @@
...
+) -> Result<(), VirtualMachineError> {
+    if KECCAK_STATE_SIZE_FELTS >= 100 {
+        return Err(VirtualMachineError::InvalidKeccakStateSizeFelts(
+            KECCAK_STATE_SIZE_FELTS,
+        ));
+    }
+
+    if BLOCK_SIZE >= 10 {
+        return Err(VirtualMachineError::InvalidBlockSize(BLOCK_SIZE));
+    }
```
Oppen, line 226: If these are compile time constants there's no point in checking they're in range.

src/vm/hints/cairo_keccak/keccak_hints.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    let mut inp = [0u64; KECCAK_STATE_SIZE_FELTS];
+    keccak::f1600(&mut inp);
+
+    let mut padding = vec![bigint!(0_u64); KECCAK_STATE_SIZE_FELTS];
+    padding.extend(u64_array_to_bigint_vec(&inp));
```
Oppen, line 232:
```suggestion
    let mut padding = [bigint!(0); KECCAK_STATE_SIZE_FELTS + KECCAK_STATE_SIZE_FELTS];
    keccak::f1600(&mut padding[KECCAK_STATE_SIZE_FELT .. ]);
```
NOTE: `keccak:f1600` will need to receive a mutable slice of `BigInt`.

src/vm/hints/cairo_keccak/keccak_hints.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    let base_padding = padding.clone();
+
+    for _ in 0..(BLOCK_SIZE - 1) {
+        padding.extend(base_padding.clone());
+    }
```
Oppen, line 238:
```suggestion
    for _ in 0..(BLOCK_SIZE - 1) {
        padding.extend_from_slice(padding.as_slice());
    }
```

Oppen: The first clone may still be needed, but the second part applies and avoids cloning in a loop.

src/vm/hints/cairo_keccak/keccak_hints.rs
```diff
@@ -0,0 +1,414 @@
...
+fn maybe_reloc_vec_to_u64_array(
+    vec: &[Option<&MaybeRelocatable>],
+) -> Result<[u64; KECCAK_STATE_SIZE_FELTS], VirtualMachineError> {
+    let array: [u64; KECCAK_STATE_SIZE_FELTS] = vec
+        .iter()
+        .map(|n| {
+            if let Some(MaybeRelocatable::Int(num)) = n {
+                num.to_u64().ok_or(VirtualMachineError::BigintToU64Fail)
+            } else {
+                Err(VirtualMachineError::NoneInMemoryRange)
```
Oppen, line 267: `ExpectedInt` would be a better error. Otherwise, a `Relocatable` would give us a misleading error.

src/vm/hints/cairo_keccak/keccak_hints.rs
```diff
@@ -0,0 +1,414 @@
...
+    let mut u64_values = maybe_reloc_vec_to_u64_array(&values)?;
+
+    // this function of the keccak crate is the one used instead of keccak_func from
+    // keccak_utils.py
+    keccak::f1600(&mut u64_values);
+
+    let bigint_values = u64_array_to_bigint_vec(&u64_values);
+
+    vm.segments
+        .write_arg(&mut vm.memory, &keccak_ptr, &bigint_values, Some(&vm.prime))
```
Oppen, line 194: Is there any chance to pass `bigint_values` as an iterator instead? This would save us from actually creating the vector.

Oppen (changes requested):

pefontana (approved):

PR #358, opened 2022-08-09
[TESTS] Memory hints integration test

entropidelic:
# Memory module integration tests

## Description
Tests for the integration of the functions `memcpy` and `memset` were added. The test focuses first on using the functions for a number of iterations, and then make the two interact with each other.
Another cairo program was added for benchmarking purposes, `memory_integration_benchmark.cairo` so that the number of iterations can be easily changed while doing benchmarks. 

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


Juan-M-V (approved):

cairo_programs/memory_integration_tests.cairo
```diff
@@ -0,0 +1,108 @@
...
+        return ()
+    end
+
+    let (dst : felt*) = alloc()
+    memcpy(dst=dst, src=src, len=len)
+
+    let result : felt = compare_arrays(src, dst, len, 0)
+    assert result = TRUE
+
+    return test_memcpy(src, len, iter + 1)
```
pefontana, line 50: I don't see the point in iterating over the same function with the same inputs.
I think it is better to run the function once, and if you want to modify the function execution time you pass a greater `len` as an input.

cairo_programs/memory_integration_tests.cairo
```diff
@@ -0,0 +1,108 @@
...
+        return ()
+    end
+
+    let (dst : felt*) = alloc()
+    memset(dst=dst, value=1234, n=n)
+
+    let result : felt =  check_array(dst, 1234, n, 0)
+    assert result = TRUE
+
+    return test_memset(n, iter + 1)
```
pefontana, line 65: Same here and in `func test_integration`

pefontana (changes requested):
Nice one @entropidelic !
Just let's see what we do with this iter parameter and we will be good to go

fmoletta (approved):

pefontana (approved):

PR #362, opened 2022-08-10
[TESTS] Operations with data structures

Juan-M-V:
# Operations with data 

## Description

Added tests for `find_element`, `search_sorted_lower` and `usort`

## Checklist

### DIFF
--- a/cairo_programs/operations_with_data.cairo
+++ b/cairo_programs/operations_with_data.cairo
@@ -49,17 +49,7 @@ func fill_bike_array{range_check_ptr:felt, bitwise_ptr: BitwiseBuiltin*}(arr: Bi
     return fill_bike_array(arr, len, index + 1)
 end
 
-func main{range_check_ptr:felt, bitwise_ptr: BitwiseBuiltin*}() -> ():
-    alloc_locals
-    let bike_arr:Bicycle* = alloc()
-    fill_bike_array(bike_arr, 30, 0)
-    let (element_ptr : Bicycle*) = find_element(
-        array_ptr=bike_arr,
-        elm_size=Bicycle.SIZE,
-        n_elms=30,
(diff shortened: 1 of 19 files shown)


cairo_programs/operations_with_data.cairo
```diff
@@ -0,0 +1,95 @@
...
+    let bike_arr:Bicycle* = alloc()
+    fill_bike_array(bike_arr, 30, 0)
+    let (element_ptr : Bicycle*) = find_element(
+        array_ptr=bike_arr,
+        elm_size=Bicycle.SIZE,
+        n_elms=30,
+        key=29,
+    )
+
+    assert element_ptr.gear_ratios.crankset = 0
```
entropidelic, line 63: Could these assertions be made with a function that checks the assertion element by element?

entropidelic: If this adds a lot of additional work don't worry

entropidelic:
These test will be used to benchmark hints and compare the rust hints implementation vs the python interpreted ones, so it would be nice to have a way to iterate over the tests, just like we are doing with the other integration tests. For this you can check `math_integration_tests.cairo`, `memory_integration_tests.cairo` and `math_cmp_and_pow_integration_tests.cairo`

cairo_programs/operations_with_data.cairo
Juan-M-V: I refactored the asserts where I could to be away from the main function. But I'm unsure if this is what you pointed out. https://github.com/lambdaclass/cleopatra_cairo/pull/362/commits/4452ab506a61988ddab8685f53d606719f2c5491

entropidelic (approved):

pefontana (approved):

Issue #226, opened 2022-07-02
reference_ids should be a HashMap<String, usize> instead of HashMap<String, BigInt>


entropidelic:
No need for the value to be BigInt.

https://github.com/lambdaclass/cleopatra_cairo/blob/6091bb8c95b1c10b6c38d9b2562d7ddbecbbf865/src/serde/deserialize_program.rs#L32

PR #375, opened 2022-08-17
Refactor reference_ids from HashMap<String, BigInt> to HashMap<String, usize>

entropidelic:
## Description
This PR solves issue #226 
Reference id's from the JSON compiled programs were first modeled as a `HashMap<String, BigInt>`. With time, we discovered there was no need for the values to be of type `BigInt` because they are referring to indexes. 

## Checklist
- [x] Linked to Github Issue

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -198,12 +198,12 @@ pub mod test_utils {
     macro_rules! ids {
         ( $( $name: expr ),* ) => {
             {
-                let mut ids = HashMap::<String, BigInt>::new();
+                let mut ids = HashMap::<String, usize>::new();
+                //let mut num = -1;
                 let mut num = -1;
                 $(
                     num += 1;
-                    ids_inner!($name, num, ids);
-
+                    ids_inner!($name, num.try_into().unwrap(), ids);
                 )*
                 ids
             }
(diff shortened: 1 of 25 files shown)


src/utils.rs
```diff
@@ -198,12 +198,11 @@ pub mod test_utils {
...
         ( $( $name: expr ),* ) => {
             {
-                let mut ids = HashMap::<String, BigInt>::new();
+                let mut ids = HashMap::<String, usize>::new();
                 let mut num = -1;
                 $(
                     num += 1;
-                    ids_inner!($name, num, ids);
-
+                    ids_inner!($name, num.try_into().unwrap(), ids);
```
Juan-M-V, line 205: Is there any possibility to avoid the unwrap? Like we ended up doing with the bignint macro:
`macro_rules! bigint {
    ($val : expr) => {
        Into::<BigInt>::into($val)
    };
}`

entropidelic: I have tested that and I couldn't find a quick way to avoid the unwrap. This macro is only used in tests, so I wouldn't worry too much anyway, but I could open an issue to handle that later on

Juan-M-V: Okay, maybe an issue would be good for the time being. Perhaps with all the refactoring, we won't even need to worry about this

Juan-M-V (approved):

PR #378, opened 2022-08-18
Refactor vm_core tests

Juan-M-V:
# Refactor tests

## Description

Use macros whenever possible for the initialization of: memory, VM, range check, etc
Resolves: #377


src/vm/vm_core.rs
```diff
@@ -441,7 +439,11 @@ impl VirtualMachine {
         self.opcode_assertions(&instruction, &operands)?;
 
         if let Some(ref mut trace) = &mut self.trace {
-            if let (RelocatableValue(ref pc), RelocatableValue(ref ap), RelocatableValue(ref fp)) = (
```
fmoletta, line 444: Why did this change?

src/vm/vm_core.rs
```diff
@@ -2276,22 +2215,15 @@ mod tests {
...
-
-        let mut vm = VirtualMachine::new(bigint!(127), Vec::new(), false);
+        let mut vm = vm!();
+        vm.memory = memory![((0, 0), 0x206800180018001_i64)];
 
-        vm.memory =
-            memory_from(mem_arr.clone(), 1).expect("Unexpected memory initialization failure");
         vm.run_context.pc = MaybeRelocatable::from((0, 0));
         vm.run_context.ap = MaybeRelocatable::from((0, 0));
         vm.run_context.fp = MaybeRelocatable::from((0, 0));
```
fmoletta, line 2223: This initialization can be removed, as VirtualMachine::new() already sets this values to (0,0)

src/vm/vm_core.rs
```diff
@@ -2489,58 +2408,28 @@ mod tests {
...
-        );
+        let mut vm = vm!(true);
         vm.accessed_addresses = Some(Vec::new());
-        for _ in 0..4 {
-            vm.memory.data.push(Vec::new());
-        }
+
         vm.run_context.pc = MaybeRelocatable::from((0, 0));
         vm.run_context.ap = MaybeRelocatable::from((1, 2));
         vm.run_context.fp = MaybeRelocatable::from((1, 2));
```
fmoletta, line 2416: We could add a macro for RunContext initialization, something like run_context!(0,2,2) that would generate a RunContext from the register's offsets (as their indexes are constant), it will be really helpful for tests on cairo_runner too

src/vm/vm_core.rs
```diff
@@ -3601,191 +3091,53 @@ mod tests {
...
         //Initialzie registers
         vm.run_context.pc = MaybeRelocatable::from((0, 3));
         vm.run_context.ap = MaybeRelocatable::from((1, 2));
         vm.run_context.fp = MaybeRelocatable::from((1, 2));
+
+        //Create program and execution segments
+        for _ in 0..2 {
+            vm.segments.add(&mut vm.memory, None);
+        }
+
```
fmoletta, line 3103: This block of code is not needed, its already covered by the memory macro

src/vm/vm_core.rs
fmoletta: It would also let us abstract the structure of the run context in tests so we can easily swap it

src/vm/vm_core.rs
Juan-M-V: Resolved: https://github.com/lambdaclass/cleopatra_cairo/pull/378/commits/4cc1177126c96cf299bb20c3bc0b53b01b0fc3aa

src/vm/vm_core.rs
Juan-M-V: I removed all instances of the segment adding where I could. However, when running this particular test (test_step_for_preset_memory_with_alloc_hint), an extra segment is needed.
```rust
assert_eq!(
            vm.memory.data[2],
            vec![Some(MaybeRelocatable::from(bigint!(1)))]
        );
```

src/vm/vm_core.rs
Juan-M-V: This change was made to avoid having the import
```rust
...
use crate::relocatable::MaybeRelocatable;
use crate::relocatable::MaybeRelocatable::RelocatableValue;
...
```
Honestly it was just a personal aesthetic choice

src/vm/vm_core.rs
Juan-M-V: Resolved: https://github.com/lambdaclass/cleopatra_cairo/pull/378/commits/1f23cd481892418f40eb3dd4f12a4ae6b2939cb9

fmoletta (approved):
Nice Work!

entropidelic:
Great work :). That's a lot of lines of code less. I think it would be good to add some unit tests to the new macros to quickly catch if there are any errors

Juan-M-V:
> Great work :). That's a lot of lines of code less. I think it would be good to add some unit tests to the new macros to quickly catch if there are any errors

Resolved: https://github.com/lambdaclass/cairo-rs/pull/378/commits/a0b970edc60a0c32e5d0c8767563db7e53a6e970

entropidelic (approved):

PR #397, opened 2022-08-26
Simplify BuiltinRunners + verify_auto_deductions

### LINKED ISSUE #169: Refactor if clause in verify_auto_deductions function
```
   pub fn verify_auto_deductions(&mut self) -> Result<(), VirtualMachineError> {
        for (i, segment) in self.memory.data.iter().enumerate() {
            for (j, value) in segment.iter().enumerate() {
                for (name, builtin) in self.builtin_runners.iter_mut() {
                    match builtin.base() {
                        Some(builtin_base) => {
                            if builtin_base.segment_index == i {
                                match builtin.deduce_memory_cell(
                                    &MaybeRelocatable::from((i, j)),
                                    &self.memory,
                                ) {
                                    Ok(None) => None,
                                    Ok(Some(deduced_memory_cell)) => {
                                        if Some(&deduced_memory_cell) != value.as_ref()
                                            && value != &None
                                        {
                                            return Err(
                                                VirtualMachineError::InconsistentAutoDeduction(
                                                    name.to_owned(),
                                                    deduced_memory_cell,
                                                    value.to_owned(),
                                                ),
                                            );
                                        }
                                        Some(deduced_memory_cell)
                                    }
                                    _ => {
                                        return Err(VirtualMachineError::InvalidInstructionEncoding)
                                    }
                                };
                            }
                        }
                        _ => return Err(VirtualMachineError::InvalidInstructionEncoding),
                    }
                }
            }
        }
        Ok(())
    }

### PR DESCRIPTION
**Pull request by fmoletta**
Remove unused included parameter from BuiltinRunner structs
Swap Option<Relocatable> for usize for BuiltinRunner base
 Explanation: 
   BuiltinRunners always have a base, so wrapping it inside an option doesn't make sense
   base always have 0 as an offset (as they hold the first address of a segment, so we can keep just the segment index)

Reduce the complexity of verify_auto_deductions due to the above changes
Invert the logic of verify_auto_decutions so that it iterates over each builtin's segment instead of iterating over the whole memory
 
Resolves #169

### CODE REVIEW

**Review by entropidelic — APPROVED** [review]

**Comment by entropidelic** [comment]
Excellent work 🎉

File: src/vm/runners/builtin_runner.rs [hunk 1]
```diff
@@ -92,24 +86,17 @@ impl RangeCheckBuiltinRunner {
...
-            } else {
-                Err(RunnerError::UninitializedBase)
-            }
-        } else {
-            Ok(Vec::new())
-        }
+
+    fn initial_stack(&self) -> Vec<MaybeRelocatable> {
+        vec![MaybeRelocatable::RelocatableValue(Relocatable::from((
+            self.base, 0,
```
Juan-M-V, line 94: Can't `MaybeRelocatable::from(..., ...)` be used here?

fmoletta: Sure, good call!

Juan-M-V (approved):

PR #382, opened 2022-08-23
Refactor RunContext registers

pefontana:
# Change RunContext registers data structure. 

pub struct RunContext {
    pub pc: MaybeRelocatable -> Relocatable,
    pub ap: MaybeRelocatable -> usize,
    pub fp: MaybeRelocatable -> usize,
..}

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #400, opened 2022-08-26
Runner tests

Juan-M-V:
# TITLE

## Description

Resolves: #382 

## Checklist
- [x] Linked to Github Issue

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -423,10 +423,14 @@ mod tests {
     use num_traits::FromPrimitive;
 
     use super::*;
-    use crate::hint_processor::builtin_hint_processor::builtin_hint_processor_definition::BuiltinHintProcessor;
-    use crate::serde::deserialize_program::ReferenceManager;
-    use crate::vm::trace::trace_entry::TraceEntry;
-    use crate::{bigint_str, relocatable};
+    use crate::{
+        bigint_str,
+        hint_processor::builtin_hint_processor::builtin_hint_processor_definition::BuiltinHintProcessor,
+        relocatable,
+        serde::deserialize_program::ReferenceManager,
+        utils::test_utils::*,
+        vm::{trace::trace_entry::TraceEntry, vm_memory::memory::Memory},
+    };
     use std::collections::HashMap;
 
     static HINT_EXECUTOR: BuiltinHintProcessor = BuiltinHintProcessor {};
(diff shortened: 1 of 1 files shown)


src/vm/runners/cairo_runner.rs
```diff
@@ -2844,6 +1678,7 @@ mod tests {
                 .segments
                 .add(&mut cairo_runner.vm.memory, None);
         }
+        // Memory initialization without macro
```
fmoletta, line 1681: Why did this have to be done without a macro?

src/vm/runners/cairo_runner.rs
```diff
@@ -3305,24 +2126,24 @@ mod tests {
...
                     b"3618502788666131213697322783095070105623107215331596699973092056135872020474"
                 )),
-                MaybeRelocatable::from(BigInt::from_i64(5189976364521848832).unwrap()),
+                MaybeRelocatable::from(bigint!(5189976364521848832_i64)),
                 MaybeRelocatable::from(bigint!(17)),
-                MaybeRelocatable::from(BigInt::from_i64(1226245742482522112).unwrap()),
+                MaybeRelocatable::from(bigint!(1226245742482522112_i64)),
                 MaybeRelocatable::from(bigint_str!(
                     b"3618502788666131213697322783095070105623107215331596699973092056135872020470"
                 )),
```
fmoletta, line 2145: We could add a macro (ie program_data! or data!) to avoid this verbose intialization

src/vm/runners/cairo_runner.rs
Juan-M-V: The test fails when the memory isn't initialized this way, I believe it has something to do with the memory gaps. I tried using the macro for all values but the last one and it also didn't work

src/vm/runners/cairo_runner.rs
Juan-M-V: Resolved: https://github.com/lambdaclass/cairo-rs/pull/400/commits/fe319b5c656d5b62e3b3f4cbe224d72758ba5b09

entropidelic:
The PR looks good, nice work! Just don't forget to add tests for the new macros. Also, consider opening an issue for the memory initializations where the macro is not working so that we can tackle that later.

Juan-M-V:
> The PR looks good, nice work! Just don't forget to add tests for the new macros. Also, consider opening an issue for the memory initializations where the macro is not working so that we can tackle that later.

Done! See commit: https://github.com/lambdaclass/cairo-rs/pull/400/commits/7539c03c1ec42277b2202e9dc26001d1d4a4a0c5 and PR: #406

src/utils.rs
```diff
@@ -499,4 +516,14 @@ mod test {
             ]
         );
     }
+
+    #[test]
+    fn data_vec_test() {
+        from_bigint_str![3];
```
pefontana, line 522: I don't get why is this here

src/vm/runners/cairo_runner.rs
```diff
@@ -2891,6 +1724,7 @@ mod tests {
             .segments
             .relocate_segments()
             .expect("Couldn't relocate after compute effective sizes");
+        println!("{:?}", rel_table);
```
pefontana, line 1727: remove

pefontana (changes requested):

src/vm/runners/cairo_runner.rs
Juan-M-V: Resolved: https://github.com/lambdaclass/cairo-rs/pull/400/commits/8f0223cb2acd374f19161171b55fa1785c8e73b9

src/utils.rs
Juan-M-V: That macro is needed so that the trait `MaybeRelocatable::from(([u8; u32], integer)` is implemented

Juan-M-V: Removed line: https://github.com/lambdaclass/cairo-rs/pull/400/commits/2cdcae7b2ef38863c744a86f090c7b4edd296ed0

pefontana (approved):

PR #401, opened 2022-08-26
Fix modulo operations

### LINKED ISSUE #261: Check that all modulo operations are being done correctly

entropidelic:
Resolves #261

Replace all modulo operations in the code where the `%` was being used to a `mod_floor()` method call.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,6 +1,7 @@
 use crate::types::relocatable::Relocatable;
 use num_bigint::BigInt;
 use num_integer::Integer;
+use std::ops::Shr;
 
 #[macro_export]
 macro_rules! bigint {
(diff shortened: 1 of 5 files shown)


fmoletta (approved):
Nice Work!

src/utils.rs
```diff
@@ -54,7 +55,7 @@ pub fn from_relocatable_to_indexes(relocatable: Relocatable) -> (usize, usize) {
 ///equivalent to val modulo prime.
 pub fn to_field_element(num: BigInt, prime: BigInt) -> BigInt {
     let half_prime = prime.clone() / bigint!(2);
```
Juan-M-V, line 57: Idk if this is out of scope for the PR, but the half prime could be calculated with a shift right and the addition could be calculated with a ref instead of cloning

entropidelic: I think it was a bit out of scope but since it is very little change I made it

Juan-M-V: Nice! Thanks!

Juan-M-V (approved):

PR #422, opened 2022-09-13
Deserialize constants

### LINKED ISSUE #355: Deserialize constants from JSON compiled programs
Constant values from cairo are being hardcoded when used in hints (you can check keccak_hints for reference). In the future if this constants change in cairo, they could make the hints crash. This can be handled correctly if the constants are correctly deserialized.

pefontana:
# Deserialize constants

Solves #355 

* Deserialize Identifier.{value, type_}
* Add Program.identifiers field

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


src/serde/deserialize_program.rs
```diff
@@ -55,9 +56,22 @@ impl Default for ApTracking {
...
+    #[serde(deserialize_with = "bigint_from_number")]
+    pub value: Option<BigInt>,
+}
+
+fn bigint_from_number<'de, D>(deserializer: D) -> Result<Option<BigInt>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let n = Number::deserialize(deserializer)?;
+    return Ok(Some(bigint_str!(n.to_string().as_bytes())));
```
fmoletta, line 74: The macro `bigint_str` uses unwraps and should only be used in tests, I would recommend using `BigInt::parse_bytes()` and handling the error case

Juan-M-V (approved):
Nice!

src/serde/deserialize_program.rs
pefontana: Thanks @fmoletta !
[Done](https://github.com/lambdaclass/cairo-rs/pull/422/commits/6d84bd49d9d635fe36b346ca52b59a610dacdea0)
I didn't add any error handling because in the case the value from the JSON is not a number the` let n = Number::deserialize(deserializer)?;` line handles the error

fmoletta (approved):
Amazing work!

PR #428, opened 2022-09-16
Refactor vm + vm proxy to use Rc<RefCell> pattern

fmoletta:
Refactors VM and VMProxy structs to wrap Memory, MemorySegmentManager & MemoryProxy inside Rc<RefCell>

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -28,7 +28,7 @@ jobs:
       with:
         python-version: '3.9'
     - name: Install test dependencies
-      run: pip install ecdsa fastecdsa sympy cairo-lang
+      run: pip install ecdsa fastecdsa sympy cairo-lang==0.9.1
     - name: Run benchmark
       run: make benchmark-action
     - name: Store benchmark result
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,7 +25,7 @@ jobs:
       with:
         python-version: '3.9'
     - name: Install test dependencies
-      run: pip install ecdsa fastecdsa sympy cairo-lang
+      run: pip install ecdsa fastecdsa sympy cairo-lang==0.9.1
     - name: format
       run: cargo fmt --all -- --check
     - name: Build
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ num-traits = "0.2"
 num-integer = "0.1.45"
 serde = { version = "1.0", features = ["derive"] }
 serde_bytes = "0.11.1"
-serde_json = "1.0"
+serde_json = { version = "1.0", features = ["arbitrary_precision"] }
 hex = "0.4.3" 
 bincode = "1.2.1"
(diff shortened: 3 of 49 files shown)


Oppen:
Is there any chance to wrap `VM` and `VMProxy` instead? That would reduce both verbosity, allocations and locks.

fmoletta:
> Is there any chance to wrap `VM` and `VMProxy` instead? That would reduce both verbosity, allocations and locks.

The aim of this refactor is to make these structures compatible with PyO3 pyclasses. To achieve this, the python context needs to be able to own these structures (hence why we can't have simple references). So if we were to wrap the vm proxy and use it we would still have references and lifetimes inside it, which are not compatible with pyclasses and we would also have to use the whole vm proxy for each object in python which is not desired. Wrapping the memory and the memory segment manager allows us to send these structures to a python context.

fmoletta:
### Performance regressions for this PR:
(Clones were used to solve conflicts with this refactor so we could start working on pyo3 integration. We will soon evaluate each situation and discuss how we may solve this without using clones. This has led to the following performance regressions)
Hints: Benchmarks using the files under cairo_programs/benchmarks

- Keccak: 
**Regression**(+3.5%) 
2.33s(main) vs 2.4s(refactor-vm-proxy) 

- Secp: 
**Regression**(+2.6%) 
2.89s(main) vs 2.97s(refactor-vm-proxy)

- Blake2s: 
**Regression**(+4%)
2.27s(main) vs 2.37s(refactor-vm-proxy)

- Dict: 
**Regression**(+5.47%)
1.81s(main) vs 1.91s(refactor-vm-proxy)

- Math: 
**Regression**(+5.8%)
998ms(main) vs 1s(refactor-vm-proxy)

- Memcpy & Memset: 
**Regression**(+4.21%)
1.22s(main) vs 1.27s(refactor-vm-proxy)

- Math_cmp & Pow: 
**Regression**(+6.2%)
32.91ms(main) vs 34.95ms(refactor-vm-proxy)

- Operations with Data Structures:
**Regression**(+5.7%)
3.89s(main) vs 4.11s(refactor-vm-proxy)

- Uint256: 
**Regression**(+6%)
2.38s(main) vs 2.53s(refactor-vm-proxy)

- Set: 
**Regression**(+107%)
233ms(main) vs 484.32 ms(refactor-vm-proxy)

Juan-M-V (approved):
Nice work!

Oppen:
I thought of a different way, explained in [this comment](https://github.com/lambdaclass/cairo-rs/issues/334#issuecomment-1253963087).
While I don't recommend trying it as explained in there as it's just too different to apply easily now, a similar approach could work by wrapping `Rc<RefCell<struct VM>>` inside of `struct VMProxy`, and then converting the accesses to fields to calls to accessors. The first part should be little code and the second part should be search and replace. That would save us from storing references inside the structures while allowing `struct VM` to stay unaltered and reducing the number of `.borrow()` calls.
Does that sound reasonable to you?

PR #432, opened 2022-09-20
Add the feature to add an external hint executor to the builtin hint processor

fmoletta:
Adds the method `add_external_executor` and `with_external_executor `for BuiltinHintExecutor
Adds the trait `ExternalHintExecutor `with the method `execute_hint`
When executing a hint, before returning an UnknownHintError, the BuiltinHintProcessor will now try to execute the hint with an external hint processor if it has one.
Proposes a way to add a python interpreter without overriding rust hints or breaking the abstraction provided by the HintProcessorTrait.
Only supports one external hint executor at a time.

### DIFF
--- a/src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
+++ b/src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
@@ -89,22 +89,53 @@ pub struct HintFunc(
         ) -> Result<(), VirtualMachineError>,
     >,
 );
+
+pub trait ExternalHintExecutor {
+    //Executes the hint which's data is provided by a dynamic structure previously created by compile_hint
+    fn execute_hint(
+        &self,
+        //Proxy to VM, contains refrences to necessary data
+        //+ MemoryProxy, which provides the necessary methods to manipulate memory
+        vm_proxy: &mut VMProxy,
+        //Proxy to ExecutionScopes, provides the necessary methods to manipulate the scopes and
+        //access current scope variables
+        exec_scopes_proxy: &mut ExecutionScopesProxy,
+        //Data structure containing ids data + hint code + ap tracking
+        hint_data: &HintProcessorData,
+    ) -> Result<(), VirtualMachineError>;
+}
 pub struct BuiltinHintProcessor {
     pub extra_hints: HashMap<String, HintFunc>,
+    pub external_executor: Option<Box<dyn ExternalHintExecutor>>,
 }
 impl BuiltinHintProcessor {
     pub fn new_empty() -> Self {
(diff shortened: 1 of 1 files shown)


src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
```diff
@@ -89,23 +89,54 @@ pub struct HintFunc(
...
+        //Proxy to ExecutionScopes, provides the necessary methods to manipulate the scopes and
+        //access current scope variables
+        exec_scopes_proxy: &mut ExecutionScopesProxy,
+        //Data structure containing ids data + hint code + ap tracking
+        hint_data: &HintProcessorData,
+    ) -> Result<(), VirtualMachineError>;
+}
 pub struct BuiltinHintProcessor {
     pub extra_hints: HashMap<String, HintFunc>,
+    pub external_executor: Option<Box<dyn ExternalHintExecutor>>,
```
Oppen, line 109: Maybe just use:
```suggestion
    pub external_executor: Option<Box<dyn HintExecutor>>,
```

src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
```diff
@@ -463,7 +494,12 @@ impl HintProcessor for BuiltinHintProcessor {
             hint_code::EC_MUL_INNER => {
                 ec_mul_inner(vm_proxy, &hint_data.ids_data, &hint_data.ap_tracking)
             }
-            code => Err(VirtualMachineError::UnknownHint(code.to_string())),
+            code => {
+                if let Some(ref executor) = self.external_executor {
+                    return executor.execute_hint(vm_proxy, exec_scopes_proxy, hint_data);
+                }
```
Oppen, line 500: What about the compilation step?

fmoletta: As the external hint executor comes as an add-on to the builtin hint processor it uses the same compilation step as the builtin hint processor

Oppen: But if it's going to be handling a superset of hints then this step also needs to call a compile step for the external executor. Otherwise it will only handle built-ins.

Oppen: But that's the thing, that's true for the builtin processor only. An arbitrary external processor may have other needs. That's why we provide the ability to compile them in the first place.
For example, a Python executor that we may add would probably want to cache the compiled code instead of compiling at every call.

PR #443, opened 2022-09-27
Decouple VM from CairoRunner

fmoletta:
- Remove VM from CairoRunner
- Now all CairoRunner methods receive a mutable reference to VM
- Add method `initialize_builtins` in order to remove the creation of builtin runners from CairoRunner::new()
- VirtualMachine::new() now creates an empty builtin runner vac
- Move write_output call to cairo_run (and therefore now cairo_run also receives the print_output boolean argument)
- Fix tests

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -178,16 +178,17 @@ pub mod test_utils {
     pub(crate) use references;
 
     macro_rules! vm_with_range_check {
-        () => {
-            VirtualMachine::new(
+        () => {{
+            let mut vm = VirtualMachine::new(
                 BigInt::new(Sign::Plus, vec![1, 0, 0, 0, 0, 0, 17, 134217728]),
-                vec![(
-                    "range_check".to_string(),
-                    Box::new(RangeCheckBuiltinRunner::new(bigint!(8), 8)),
-                )],
                 false,
-            )
-        };
+            );
+            vm.builtin_runners = vec![(
+                "range_check".to_string(),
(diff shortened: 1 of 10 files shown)


Oppen:
Most of `CairoRunner`'s modified methods only operate on `vm`. This suggests these should just become `VirtualMachine` methods.

fmoletta:
> Most of `CairoRunner`'s modified methods only operate on `vm`. This suggests these should just become `VirtualMachine` methods.

CairoRunner is in charge of initializing the vm and carrying out the relocation process. Moving these methods to VirtualMachine would delegate all responsibility to VirtualMachine which's responsibility begins once we enter the main execution loop (step). Also, these methods operate on both the runner and the vm.

entropidelic (approved):

src/utils.rs
```diff
@@ -178,32 +178,31 @@ pub mod test_utils {
...
                 false,
-            )
-        };
+            );
+            vm.builtin_runners = vec![(
+                "range_check".to_string(),
+                Box::new(RangeCheckBuiltinRunner::new(bigint!(8), 8)),
+            )];
+            vm
+        }};
```
Oppen, line 191: This looks strange, why do we have extra braces for the block?

Oppen (approved):

src/utils.rs
fmoletta: The extra braces are needed because of the let statement

PR #457, opened 2022-09-29
Replace add segments

Juan-M-V:
# TITLE

## Description

Replace memory.add_segment in hints to use VMProxy methods

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/hint_processor/builtin_hint_processor/dict_manager.rs
+++ b/src/hint_processor/builtin_hint_processor/dict_manager.rs
@@ -252,7 +252,7 @@ mod tests {
             dict_manager.trackers.get(&0),
             Some(&DictTracker::new_empty(&relocatable!(0, 0)))
         );
-        assert_eq!(vm_proxy.get_num_segments(), 1);
+        assert_eq!(vm.segments.num_segments, 1);
     }
 
     #[test]
@@ -271,7 +271,7 @@ mod tests {
                 None
             ))
         );
-        assert_eq!(vm_proxy.get_num_segments(), 1);
+        assert_eq!(vm.segments.num_segments, 1);
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


src/hint_processor/builtin_hint_processor/dict_manager.rs
```diff
@@ -323,7 +313,7 @@ mod tests {
                 Some(initial_dict)
             ))
         );
-        assert_eq!(segments.num_segments, 1);
+        assert_eq!(vm_proxy.segments.num_segments, 1);
```
pefontana, line 316: Will this fail when we turn vm_proxy.segments into a private field?

pefontana (changes requested):
Nice one @Juan-M-V 
Let's check the testing of private fields and we will be good to go

src/hint_processor/builtin_hint_processor/dict_manager.rs
Juan-M-V: resolved: https://github.com/lambdaclass/cairo-rs/pull/457/commits/441652db416c59a732477ba43525048061c6a791

pefontana: It is a bad idea to add a VMProxy method that is only used in testing.
Wont be better to assert `assert_eq!(vm.segments.num_segments, 1);`?

Juan-M-V: Actually resolved(?): https://github.com/lambdaclass/cairo-rs/pull/457/commits/85b1591b4965b500a41972722a4a7f9af85a34c9

pefontana (approved):

Oppen (approved):

PR #475, opened 2022-10-05
Change visibility of fields and methods

entropidelic:
In order for `cairo-rs-py` to execute programs via `cairo_run_py`, some methods were created and the visibility of other methods and fields were changed. In the future this could be done in a better way, but for the moment this should be ok for having something operational.


fmoletta (approved):
LGTM

pefontana (approved):

Issue #480, opened 2022-10-13
Add method Memory.relocate_value() and integrate it into Memory.get()


fmoletta:
Add method [relocate_value](https://github.com/starkware-libs/cairo-lang/blob/b614d1867c64f3fb2cf4a4879348cfcf87c3a5a7/src/starkware/cairo/lang/vm/memory_dict.py#L145) to Memory.
This method should also be called on the return value of Memory.get().
Adding this method will break the current Memory.get() signature (as relocate_value() may return a newly created value). Both Memory.relocate_value() and Memory.get() should now return a Cow<(MaybeRelocatable)>, and functions that use Memory.get() should have different logic implemented for when the value is owned or referenced.

PR #484, opened 2022-10-18
Implement `Memory::relocate value`

azteca1998:
# Implement `Memory::relocate_value`

## Description

This PR implements temporary segments. Following issue #480.

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added


src/vm/runners/builtin_runner/bitwise.rs
```diff
@@ -58,9 +58,13 @@ impl BuiltinRunner for BitwiseBuiltinRunner {
...
         let y_addr = x_addr.add_usize_mod(1, None);
-        if let (Ok(Some(MaybeRelocatable::Int(num_x))), Ok(Some(MaybeRelocatable::Int(num_y)))) =
-            (memory.get(&x_addr), memory.get(&y_addr))
-        {
+
+        let num_x = memory.get(&x_addr);
+        let num_y = memory.get(&y_addr);
+        if let (Ok(Some(MaybeRelocatable::Int(num_x))), Ok(Some(MaybeRelocatable::Int(num_y)))) = (
+            num_x.as_ref().map(|x| x.as_ref().map(|x| x.as_ref())),
+            num_y.as_ref().map(|x| x.as_ref().map(|x| x.as_ref())),
```
Oppen, line 66: These `map`s look weird. Why doesn't `as_ref` suffice?

src/vm/runners/builtin_runner/hash.rs
```diff
@@ -58,15 +58,18 @@ impl BuiltinRunner for HashBuiltinRunner {
...
-            memory.get(&MaybeRelocatable::RelocatableValue(Relocatable {
-                segment_index: address.segment_index,
-                offset: address.offset - 1,
-            })),
-            memory.get(&MaybeRelocatable::RelocatableValue(Relocatable {
-                segment_index: address.segment_index,
-                offset: address.offset - 2,
-            })),
+            num_a.as_ref().map(|x| x.as_ref().map(|x| x.as_ref())),
+            num_b.as_ref().map(|x| x.as_ref().map(|x| x.as_ref())),
```
Oppen, line 72: Same question.

src/vm/vm_core.rs
```diff
@@ -103,9 +103,10 @@ impl VirtualMachine {
     ///Returns the encoded instruction (the value at pc) and the immediate value (the value at pc + 1, if it exists in the memory).
     fn get_instruction_encoding(
         &self,
-    ) -> Result<(&BigInt, Option<&MaybeRelocatable>), VirtualMachineError> {
-        let encoding_ref: &BigInt = match self.memory.get(&self.run_context.pc) {
-            Ok(Some(MaybeRelocatable::Int(ref encoding))) => encoding,
+    ) -> Result<(Cow<BigInt>, Option<Cow<MaybeRelocatable>>), VirtualMachineError> {
+        let encoding_ref = match self.memory.get(&self.run_context.pc) {
+            Ok(Some(Cow::Owned(MaybeRelocatable::Int(encoding)))) => Cow::Owned(encoding),
+            Ok(Some(Cow::Borrowed(MaybeRelocatable::Int(encoding)))) => Cow::Borrowed(encoding),
```
Oppen, line 109: Just using the references here should be enough, I think.

src/vm/vm_core.rs
```diff
@@ -439,7 +440,7 @@ impl VirtualMachine {
         let (instruction_ref, imm) = self.get_instruction_encoding()?;
         match instruction_ref.to_i64() {
             Some(instruction) => {
-                if let Some(MaybeRelocatable::Int(imm_ref)) = imm {
+                if let Some(MaybeRelocatable::Int(imm_ref)) = imm.as_ref().map(|x| x.as_ref()) {
```
Oppen, line 443: As before.

Oppen (approved):
It looks mostly OK. Check the comments for some minor stuff.

src/vm/runners/builtin_runner/bitwise.rs
azteca1998: `num_x` and `num_y` are `Result<Option<Cow<MaybeRelocatable>>, _>`, while we're interested in getting a `&MaybeRelocatable`. `Result` and `Option` can be matched no problem, but to match the `Cow` we'd need two branches (`Owned` and `Borrowed`).

In order to avoid that, the first two `.as_ref().map(|_| todo!())` map the internal values of the `Result` (when `Ok`) and `Option` wrappers by reference (otherwise the value is moved out) while the last `x.as_ref()` converts `Cow<MaybeRelocatable>` to a `&MaybeRelocatable` regardless of its actual variant so that it can be matched without code duplication.

Using a single `.as_ref()` would provide a `Result<&Option<Cow<MaybeRelocatable>>, &_>` which does not remove the `Cow`.

A `match` could also be used, but it'd require `2^n` combined patterns (4 in this case), one for each combination of `Owned` and `Borrowed`. For example, something like this:
```rs
match (num_x, num_y) {
    (Ok(Some(Cow::Borrowed(num_x))), Ok(Some(Cow::Borrowed(num_y)))) |
    (Ok(Some(Cow::Borrowed(num_x))), Ok(Some(Cow::Owned(ref num_y)))) |
    (Ok(Some(Cow::Owned(ref num_x))), Ok(Some(Cow::Borrowed(num_y)))) |
    (Ok(Some(Cow::Owned(ref num_x))), Ok(Some(Cow::Owned(ref num_y)))) => {
        // `if` statement body here
        todo!()
    },
    _ => {},
}
```

Please let me know if a `match` would be preferred in this case over the `map` and `as_ref` chain.

src/vm/vm_core.rs
azteca1998: Is it impossible to return an `Owned` value on the program segment?

src/vm/runners/builtin_runner/bitwise.rs
azteca1998: Now that I think about it, there's a potentially cleaner solution which requires a trait utility. An example can be found [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=213ed496e3fe6851e4b4fa4a8b09f949).

Basically, a trait is implemented for each `Result<Option<Cow<'a, T>>, E>` that provides a method to transform it to `Result<Option<&'a T>, E>`. It moves the `as_ref`-`map` chain into the trait while leaving a cleaner source code where used.

Oppen: > Please let me know if a match would be preferred in this case over the map and as_ref chain.

Most certainly not D:

> Basically, a trait is implemented for each Result<Option<Cow<'a, T>>, E> that provides a method to transform it to Result<Option<&'a T>, E>. It moves the as_ref-map chain into the trait while leaving a cleaner source code where used.

That does seem like a better solution in that it's less error prone and explodes the code much less. It's OK as it is given the explanation tho.

Juan-M-V (approved):

src/vm/vm_memory/memory.rs
```diff
@@ -66,44 +78,117 @@ impl Memory {
         self.validate_memory_cell(&MaybeRelocatable::from(key))
     }
 
-    pub fn get<'a, K: 'a>(&self, key: &'a K) -> Result<Option<&MaybeRelocatable>, MemoryError>
+    /// Retrieve a value from memory (either normal or temporary) and apply relocation rules
+    pub fn get<'a, 'b: 'a, K: 'a>(
```
Oppen, line 82: Effectively we concluded this one should be `pub(crate)` rather than `pub`. Only `VirtualMachine::get` is part of the pubic interface, but not `Memory::get`.

Oppen (approved):

HermanObst:
[add implementation of relocate_values for support Temporary Segments](https://trello.com/c/tNp6aeHr/10-add-implementation-of-relocatevalues-for-support-temporary-segments)

PR #487, opened 2022-10-19
First implementation of signature builtin

Jrigada:
# Signature Builtin

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/vm/runners/builtin_runner/signature.rs
```diff
@@ -61,33 +62,48 @@ impl BuiltinRunner for SignatureBuiltinRunner {
...
                   -> Result<Vec<MaybeRelocatable>, MemoryError> {
-                let pubkey: Relocatable;
-                let msg;
+                let address = if let MaybeRelocatable::Int(_address) = address {
+                    return Err(MemoryError::AddressNotRelocatable);
+                } else if let MaybeRelocatable::RelocatableValue(address) = address {
+                    address
+                } else {
+                    unreachable!()
+                };
```
Oppen, line 71: Nit:
```suggestion
                let address = match address {
                    MaybeRelocatable::RelocatableValue(address) => address,
                    _ => return Err(MemoryError::AddressNotRelocatable),
                };
```
There's only two types of `MaybeRelocatable` so the catch-all is correct.

Oppen: Alternatively, make the second case the `else`.

Oppen (changes requested):

juanbono (changes requested):
Please add missing assert in `validate_existing_memory_for_signature` and add meaningful names to the rest of the tests.

Oppen (commented):
Looks mostly OK. We may need to double check the errors this returns, some don't seem to correlate to what's actually failing, but it doesn't sound like the greatest priority.

src/vm/runners/builtin_runner/range_check.rs
```diff
@@ -73,28 +73,24 @@ impl RangeCheckBuiltinRunner {
...
-                match memory.get(address)? {
-                    Some(Cow::Owned(MaybeRelocatable::Int(ref num)))
-                    | Some(Cow::Borrowed(MaybeRelocatable::Int(ref num))) => {
-                        if &BigInt::zero() <= num && num < &BigInt::one().shl(128u8) {
-                            Ok(address.to_owned())
-                        } else {
-                            Err(MemoryError::NumOutOfBounds)
-                        }
+             -> Result<Vec<MaybeRelocatable>, MemoryError> {
+                if let Some(Cow::Borrowed(MaybeRelocatable::Int(ref num))) = memory.get(address)? {
```
fmoletta, line 77: ` Cow::Owned(MaybeRelocatable::Int)` should also be a valid case (Treated as equal as `Cow::Borrowed(MaybeRelocatable::Int)`).
 As we are already calling `to_owned()` on the return value, we can instead call `into_owned() `before the match ( `match memory.get(address)?.into_owned()` and get rid of the Cow before the comparison (to avoid these two cases).

src/vm/runners/builtin_runner/signature.rs
```diff
@@ -0,0 +1,469 @@
...
+};
+
+use k256::ecdsa::{signature::Verifier, Signature, VerifyingKey};
+use num_integer::{div_ceil, Integer};
+use num_traits::ToPrimitive;
+use std::{any::Any, collections::HashMap};
+
+#[derive(Debug)]
+pub struct SignatureBuiltinRunner {
+    _name: String,
```
fmoletta, line 24: Is this field needed?

juanbono, line 24: what is the purpose of this field?

src/vm/runners/builtin_runner/range_check.rs
```diff
@@ -55,10 +55,10 @@ impl BuiltinRunner for RangeCheckBuiltinRunner {
         let rule: ValidationRule = ValidationRule(Box::new(
             |memory: &Memory,
              address: &MaybeRelocatable|
-             -> Result<MaybeRelocatable, MemoryError> {
+             -> Result<Vec<MaybeRelocatable>, MemoryError> {
```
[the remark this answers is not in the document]
juanbono: Please test if this doesn't break cairo-rs-py

juanbono (approved):
LGTM. I left some comments please address them :)

src/vm/runners/builtin_runner/signature.rs
```diff
@@ -0,0 +1,469 @@
...
+                            .map_err(|_| MemoryError::AddressNotRelocatable)?;
+                        let msg_addr = address.clone();
+                        (pubkey_addr, msg_addr)
+                    }
+                    _ => return Ok(Vec::new()),
+                };
+
+                let (_sign, msg) = memory
+                    .get_integer(&msg_addr)
+                    .map_err(|_| MemoryError::AddressNotRelocatable)?
```
fmoletta, line 118: This error management here doesn't make sense, get_integer will never return a relocatable value

src/vm/runners/builtin_runner/signature.rs
```diff
@@ -0,0 +1,469 @@
...
+                    .to_bytes_be();
+                let (_sign, pubkey) = memory
+                    .get_integer(&pubkey_addr)
+                    .map_err(|_| MemoryError::AddressNotRelocatable)?
+                    .to_bytes_be();
```
fmoletta, line 123: Same here

juanbono: I think he left that as a placeholder for a more descriptive error

src/vm/runners/builtin_runner/signature.rs
```diff
@@ -0,0 +1,469 @@
...
+
+                let signature = signatures
+                    .get(&pubkey_addr)
+                    .ok_or(MemoryError::AddressNotRelocatable)?;
+
+                verify_key
+                    .verify(&msg, signature)
+                    .map_err(|_| MemoryError::AddressNotRelocatable)?;
+                Ok(Vec::new())
+            },
```
fmoletta, line 136: These errors also don't make sense, we should create specific error types for these cases

src/vm/vm_memory/memory.rs
```diff
@@ -527,6 +541,82 @@ mod memory_tests {
...
+            )
+            .unwrap();
+        memory
+            .insert(
+                &MaybeRelocatable::from((1, 1)),
+                &MaybeRelocatable::from(bigint_str!(b"-1472574760335685482768423018116732869320670550222259018541069375211356613248")),
+            ).unwrap();
+        builtin.add_validation_rule(&mut memory).unwrap();
+        let error = memory.validate_existing_memory();
+        assert!(error.is_err())
```
fmoletta, line 572: Can't we use an assert_eq! and compare the error to a concrete error here?,  keeping the error info available is more useful when debugging

src/vm/vm_memory/memory.rs
```diff
@@ -527,6 +541,82 @@ mod memory_tests {
...
+                &MaybeRelocatable::from((1, 1)),
+                &MaybeRelocatable::from(bigint!(1_i32)),
+            )
+            .unwrap();
+
+        builtin.add_validation_rule(&mut memory).unwrap();
+
+        let result = memory.validate_existing_memory();
+
+        assert!(result.is_ok())
```
fmoletta, line 617:
```suggestion
        assert_eq!(result, Ok(())
```
This is so that we can see what the error is if this test fails

fmoletta (changes requested):
Please fix the logic on RangeCheckBuiltinRunner::ad_validation_rule()
Aside from that, the PR looks quite good aside from minor error management-related comments

entropidelic (approved):

fmoletta (approved):

Oppen (changes requested):

Oppen (approved):

PR #495, opened 2022-10-25
Add method to expose program constants

fmoletta:
Add method `get_constants` to CairoRunner in order to expose the program's constants without exposing the whole compiled program

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -2249,4 +2249,29 @@ mod tests {
         assert_eq!(vm.builtin_runners[3].0, String::from("bitwise"));
         assert_eq!(vm.builtin_runners[4].0, String::from("ec_op"));
     }
+
+    #[test]
+    fn get_constants() {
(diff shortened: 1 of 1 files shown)


src/vm/runners/cairo_runner.rs
```diff
@@ -277,6 +277,10 @@ impl<'a> CairoRunner<'a> {
         Ok(hint_data_dictionary)
     }
 
+    pub fn get_constants(&self) -> &HashMap<String, BigInt> {
+        &self.program.constants
+    }
```
HermanObst, line 282: @fmoletta please add a test for this method

HermanObst (changes requested):

src/vm/runners/cairo_runner.rs
fmoletta: Test added!

Oppen (approved):

Oppen:
It seems like you need to rebase after the `CairoRunner` changes.

PR #496, opened 2022-10-25
Add disable_trace method to VirtualMachine

fmoletta:
Adds `disable_trace` method to VirtualMachine in order to allow disabling the VM's trace after its creation


Oppen:
In which context is this disabled at runtime rather than creation?

Oppen (approved):

HermanObst (approved):

PR #501, opened 2022-10-26
Implement `CairoRunner::mark_as_accessed()`.

azteca1998:
# Implement `CairoRunner::mark_as_accessed()`.

## Description

Implementation `CairoRunner::mark_as_accessed()`.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -297,6 +297,22 @@ impl CairoRunner {
         Ok(())
     }
 
+    /// Mark a memory address as accesed.
+    pub fn mark_as_accessed(
+        &mut self,
+        address: Relocatable,
+        size: usize,
+        vm: &mut VirtualMachine,
+    ) -> Result<(), VirtualMachineError> {
+        let accessed_addressess = match &mut vm.accessed_addresses {
+            Some(x) => x,
+            None => return Err(VirtualMachineError::RunIsNotEnded),
+        };
+
+        accessed_addressess.extend((0..size).map(|i| &address + i));
+        Ok(())
+    }
+
     ///Relocates the VM's memory, turning bidimensional indexes into contiguous numbers, and values into BigInts
     /// Uses the relocation_table to asign each index a number according to the value on its segment number
     fn relocate_memory(
@@ -426,7 +442,7 @@ mod tests {
         vm::{trace::trace_entry::TraceEntry, vm_memory::memory::Memory},
     };
     use num_bigint::Sign;
-    use std::collections::HashMap;
+    use std::collections::{HashMap, HashSet};
 
     #[test]
     fn initialize_builtins_with_disordered_builtins() {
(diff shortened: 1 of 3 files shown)


azteca1998:
[add mark_as_accessed to PyCairoRunner](https://trello.com/c/tXfCSZB1/7-add-markasaccessed-to-pycairorunner)

src/vm/runners/cairo_runner.rs
```diff
@@ -297,6 +297,22 @@ impl CairoRunner {
...
+    /// Mark a memory address as accesed.
+    pub fn mark_as_accessed(
+        &mut self,
+        address: Relocatable,
+        size: usize,
+        vm: &mut VirtualMachine,
+    ) -> Result<(), VirtualMachineError> {
+        let accessed_addressess = match &mut vm.accessed_addresses {
+            Some(x) => x,
+            None => return Err(VirtualMachineError::RunIsNotEnded),
```
fmoletta, line 309: This error doesn't make sense, it would be better to return something similar to "mark_as_accessed called with no accessed_addresses"

fmoletta: Looking at the source code there is misunderstanding, `cairo_runner.accessed_addresses` contains the relocated addresses, and not the vm's `accessed_addresses`. So the error returned is correct, but we should add the field accessed_addresses (or even better, relocated_accessed_addresses to avoid future confusion) to CairoRunner (this field should have the type `Option<Vec<BigInt>>`, as it contains relocated addresses) and use `self.accessed_addresses `inside the method instead of `self.vm.accessed_adresses`

Oppen: Nit: `RunNotFinished`.

src/vm/vm_core.rs
```diff
@@ -47,7 +51,7 @@ pub struct VirtualMachine {
     pub(crate) segments: MemorySegmentManager,
     pub(crate) _program_base: Option<MaybeRelocatable>,
     pub(crate) memory: Memory,
-    accessed_addresses: Option<Vec<Relocatable>>,
```
Oppen, line 50: This was made a `Vec` for a reason, `HashSet` lookups were introducing significant overhead. Just appending is enough to mark a position as accessed as relocation will take care of uniqueness later on.

src/vm/runners/cairo_runner.rs
azteca1998: Alright, it should be fixed now.

src/vm/runners/cairo_runner.rs
```diff
@@ -2268,4 +2289,47 @@ mod tests {
...
+            },
+            identifiers: HashMap::new(),
+        };
+
+        let mut cairo_runner = CairoRunner::new(&program).unwrap();
+
+        assert_eq!(
+            cairo_runner.mark_as_accessed((0, 0).into(), 3),
+            Err(VirtualMachineError::RunNotFinished),
+        );
```
fmoletta, line 2315: This should be two separate tests, one for each case (accessed_addresses being None and being Some)

src/vm/vm_core.rs
Oppen: At least for now. If it proves to add too much overhead later on we can change it.

Oppen (approved):

fmoletta (approved):

PR #502, opened 2022-10-26
Implement `CairoRunner::get_memory_holes()` and its dependencies.

azteca1998:
# Implement `CairoRunner::get_memory_holes()` and its dependencies.

## Description

This PR will unlock the following tasks:
  - [`get_execution_resources()`](https://hackmd.io/vRQ3yuA9QhuvGA6igOLwUA?view#Task-list-for-CairoRunnerget_execution_resources)
  - [`proof_mode` section of `CairoRunner::end_run`](https://hackmd.io/vRQ3yuA9QhuvGA6igOLwUA?view#Task-list-for-proof_mode-section-of-CairoRunnerend_run)

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/vm/runners/builtin_runner/bitwise.rs
```diff
@@ -96,6 +97,10 @@ impl BuiltinRunner for BitwiseBuiltinRunner {
     fn as_any(&self) -> &dyn Any {
         self
     }
+
+    fn get_memory_accesses(&self, _runner: &CairoRunner) -> std::collections::HashSet<Relocatable> {
```
fmoletta, line 101: Please implement this method, (In a separate PR, and add that PR as a dependency to this one), as we can't have todo!() in the main branch

azteca1998: [Done](https://github.com/lambdaclass/cairo-rs/pull/508).

src/vm/vm_memory/memory_segments.rs
```diff
@@ -128,6 +136,56 @@ impl MemorySegmentManager {
...
+            offset_set.insert(offset);
+        }
+
+        Ok((0..self.segment_sizes.len())
+            .chain(0..segment_used_sizes.len())
+            .collect::<HashSet<_>>()
+            .into_iter()
+            .map(|index| accessed_offsets_sets.get(&index).unwrap())
+            .map(|(segment_size, offsets_set)| segment_size - offsets_set.len())
+            .sum())
```
Oppen, line 180: This is equivalent to:
```suggestion
        let max = cmp::max(self.segment_sizes.len(), segment_used_sizes.len());
        Ok((0..max)
            .map(|index| accessed_offsets_sets.get(&index).unwrap())
            .map(|(segment_size, offsets_len)| segment_size - offsets_set.len())
            .sum())
```
but saving a hash creation and several accesses and reallocations.
I'm also a bit wary of the `unwrap` call. Wouldn't it break if any segment was skipped or created but not used?

Oppen (changes requested):

src/vm/vm_core.rs
```diff
@@ -47,7 +47,7 @@ pub struct VirtualMachine {
     pub(crate) segments: MemorySegmentManager,
     pub(crate) _program_base: Option<MaybeRelocatable>,
     pub(crate) memory: Memory,
-    accessed_addresses: Option<Vec<Relocatable>>,
+    pub(crate) accessed_addresses: Option<Vec<Relocatable>>,
```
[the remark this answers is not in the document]
azteca1998: This isn't needed any more. It has been reverted in the other PR. If it ever happens again (and for real), should I make a getter?

I'll change its usage (in this PR) to `self.accessed_addresses` which is the correct `accessed_addresses`. A merge (from main to this one) should fix it.

src/vm/vm_memory/memory_segments.rs
azteca1998: No excuse for this one :C
I completely missed this by focusing on replicating the functionality from the starknet reference.

azteca1998: About the `.unwrap()`, Python doesn't have that problem because it uses the keys from the `accessed_offsets` dictionary where we use the index. Knowing this, I assumed it'd be ok to unwrap here, but now that you mention it...

In our case, if a segment isn't accessed then no value will be created. Since Python would not crash in that case I think modifying it to filter out invalid segments is the right choice.

src/vm/vm_memory/memory_segments.rs
```diff
@@ -73,7 +78,7 @@ impl MemorySegmentManager {
     pub fn get_segment_used_size(&self, index: usize) -> Option<usize> {
         self.segment_used_sizes
             .as_ref()
-            .map(|used_sizes| used_sizes[index])
+            .and_then(|used_sizes| used_sizes.get(index).copied())
     }
```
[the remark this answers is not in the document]
azteca1998: Because it slipped my mind. Fixed.

src/vm/vm_memory/memory_segments.rs
```diff
@@ -128,6 +133,54 @@ impl MemorySegmentManager {
...
+        let segment_used_sizes = self
+            .segment_used_sizes
+            .as_ref()
+            .ok_or(MemoryError::MissingSegmentUsedSizes)?;
+
+        let mut accessed_offsets_sets = HashMap::new();
+        for addr in accessed_addresses {
+            if addr.segment_index < 0 {
+                return Err(MemoryError::AddressInTemporarySegment(addr.segment_index));
+            }
```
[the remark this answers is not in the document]
azteca1998: [It is](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/vm/memory_segments.py#L197), but in assert form.
I just moved it so that no extra computations are executed if this condition is not met.

fmoletta (approved):
Awesome work!

src/vm/runners/builtin_runner/hash.rs
```diff
@@ -100,8 +101,17 @@ impl BuiltinRunner for HashBuiltinRunner {
...
+        let segment_size = vm
+            .segments
+            .get_segment_size(
+                self.base
+                    .try_into()
+                    .map_err(|_| MemoryError::AddressInTemporarySegment(self.base))?,
+            )
+            .ok_or(MemoryError::MissingSegmentUsedSizes)?;
+
+        Ok((0..segment_size).map(|i| (self.base, i).into()).collect())
```
Oppen, line 114: Is this the same implementation for every builtin runner? Can't this be implemented only once?

src/vm/runners/cairo_runner.rs
```diff
@@ -303,17 +303,32 @@ impl CairoRunner {
...
+    /// Mark a memory address as accesed.
+    pub fn mark_as_accessed(
+        &mut self,
+        address: Relocatable,
+        size: usize,
+    ) -> Result<(), VirtualMachineError> {
+        let accessed_addressess = match &mut self.accessed_addresses {
+            Some(x) => x,
+            None => return Err(VirtualMachineError::RunNotFinished),
+        };
```
Oppen, line 315: This looks like a candidate for:
```suggestion
        let accessed_addressess = self.accessed_addresses.ok_or(VirtualMachineError::RunNotFinished)?;
```

src/vm/runners/builtin_runner/hash.rs
azteca1998: Not in every builtin, the output has some differences.

May I change `BuiltinRunner::base()` to return the segment only? Since the offset is always zero it should pose no problem and would make it easier to implement the method on the trait (avoid unnecessary comparisons, for example).

src/vm/runners/builtin_runner/bitwise.rs
```diff
@@ -143,4 +144,41 @@ mod tests {
...
+        assert_eq!(
+            builtin.get_memory_accesses(&vm),
+            Ok(vec![
+                (builtin.base, 0).into(),
+                (builtin.base, 1).into(),
+                (builtin.base, 2).into(),
+                (builtin.base, 3).into(),
+            ]),
+        );
+    }
```
Oppen, line 183: We can also deduplicate the tests now that they're all in `mod.rs`.

src/vm/runners/builtin_runner/hash.rs
Oppen: I think that's a good solution.

Oppen (approved):
Scratch my last comment, I think it's better as is.

PR #506, opened 2022-10-27
Add layouts to CairoRunner

fmoletta:
Add field layout to CairoRunner
Adds`CairoLayout` struct
Extracts builtin arguments from layout
Add tests for new methods

Merging this PR will require small changes in cairo-rs-py to maintain compatibility


src/vm/runners/cairo_runner.rs
```diff
@@ -94,21 +105,70 @@ impl CairoRunner {
...
-                builtin_runners.push((builtin_name.clone(), EcOpBuiltinRunner::new(256).into()));
+                builtin_runners.push((
+                    builtin_name.clone(),
+                    EcOpBuiltinRunner::new(self.layout.builtins.ec_op.as_ref().ok_or_else(
+                        || {
+                            RunnerError::NoBuiltinForInstance(
+                                self.layout.name.clone(),
+                                builtin_name.to_string(),
+                            )
+                        },
```
Oppen, line 168: This lambda is repeated many times, it may be a good idea to just extract it as a variable.

src/vm/runners/cairo_runner.rs
```diff
@@ -48,10 +48,21 @@ pub struct CairoRunner {
 }
 
 impl CairoRunner {
-    pub fn new(program: &Program) -> Result<CairoRunner, RunnerError> {
+    pub fn new(program: &Program, layout: Option<String>) -> Result<CairoRunner, RunnerError> {
+        let layout = layout.unwrap_or_else(|| String::from("plain"));
```
Oppen, line 52: I think given it makes sense to just handle this at the option level with a default value and drop the `Option<_>`.

Oppen (commented):
I skimmed through the PR and need to properly read the docs about layouts before doing a proper review, but I have some little comments anyway.

src/vm/runners/cairo_runner.rs
fmoletta: I don't understand this comment, We have to do this check for every builtin which uses the corresponding builtin instance data in its creation, we can't just do it once

src/vm/runners/cairo_runner.rs
fmoletta: Changed (7735799)

src/vm/runners/cairo_runner.rs
Oppen: The check, yes. The lambda, no. AFAIK lambdas can be assigned, so you can do:
```
let nobuiltin = || {
    RunnerError::NoBuiltinForInstance(
        self.layout.name.clone(),
        builtin_name.to_string(),
    )
};
```
Then you just pass `nobuiltin` in the call places.

fmoletta: I understand now, thanks for explaining! good call!

src/main.rs
```diff
@@ -28,6 +28,8 @@ struct Args {
     trace: Option<PathBuf>,
     #[structopt(long = "--memory_file")]
     memory_file: Option<PathBuf>,
+    #[structopt(long = "--layout", default_value = "plain")]
+    layout: String,
```
juanbono, line 32: Here we can check that the layout we receive is one of the 6 defined layouts.

src/vm/runners/cairo_runner.rs
fmoletta: Changed(6990cf9)

juanbono (approved):
LGTM. Maybe we can check that we don't receive an invalid layout from the CLI (using an enum for the layout types) instead of failing when creating the CairoRunner.

src/main.rs
fmoletta: I added a validation function to the argument (6990cf9)

Oppen (approved):

PR #507, opened 2022-10-27
Implement `CairoRunner::end_run` and its dependencies

azteca1998:
# Implement `CairoRunner::end_run` and its dependencies

## Description

Implementation `CairoRunner::end_run` and its dependencies.

Depends on #502.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -26,6 +26,7 @@ use crate::{
 use num_bigint::BigInt;
 use std::{
     any::Any,
+    borrow::Cow,
     collections::{HashMap, HashSet},
     io,
 };
@@ -39,6 +40,7 @@ pub struct CairoRunner {
     initial_ap: Option<Relocatable>,
     initial_fp: Option<Relocatable>,
     initial_pc: Option<Relocatable>,
+    run_ended: bool,
     pub relocated_memory: Vec<Option<BigInt>>,
     pub relocated_trace: Option<Vec<RelocatedTraceEntry>>,
     pub exec_scopes: ExecutionScopes,
@@ -55,6 +57,7 @@ impl CairoRunner {
             initial_ap: None,
             initial_fp: None,
             initial_pc: None,
+            run_ended: false,
             relocated_memory: Vec::new(),
             relocated_trace: None,
             exec_scopes: ExecutionScopes::new(),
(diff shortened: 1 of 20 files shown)


azteca1998:
[Add and export run_from_entrypoint to CairoRunner](https://trello.com/c/IvCBIHHw/34-add-and-export-runfromentrypoint-to-cairorunner)

src/vm/runners/cairo_runner.rs
```diff
@@ -297,6 +304,80 @@ impl CairoRunner {
...
+                .ok_or_else::<VirtualMachineError, _>(|| {
+                    MemoryError::MissingAccessedAddresses.into()
+                })?;
+            let mut new_accessed_addresses = Vec::with_capacity(accessed_addresses.len());
+
+            for addr in accessed_addresses {
+                let relocated_addr = vm
+                    .memory
+                    .relocate_value(Some(Cow::Owned(addr.into())))?
+                    .unwrap()
```
fmoletta, line 350: Please remove this unwrap

src/vm/runners/cairo_runner.rs
```diff
@@ -297,6 +304,80 @@ impl CairoRunner {
...
+
+            vm.accessed_addresses = Some(new_accessed_addresses);
+        }
+
+        self.relocate(vm)
+            .map_err(VirtualMachineError::TracerError)?;
+        vm.end_run(&self.exec_scopes)?;
+
+        if !disable_finalize_all {
+            // vm.memory.freeze();
```
fmoletta, line 364: Please remove this comment

src/vm/runners/cairo_runner.rs
```diff
@@ -297,6 +304,80 @@ impl CairoRunner {
...
+            vm.segments.compute_effective_sizes(&vm.memory);
+
+            // if self.proof_mode && !disable_trace_padding {
+            //     self.run_until_next_power_of_2();
+            //     while !self.check_used_cells() {
+            //         self.run_for_steps(1);
+            //         self.run_until_next_power_of_2();
+            //     }
+            // }
```
fmoletta, line 373: Same here

fmoletta (changes requested):

src/vm/runners/cairo_runner.rs
azteca1998: This unwrap will never be triggered because as long as the parameter to `relocate_value` is `Some(_)` it will return `Some(_)`.

What should I do in this case?

fmoletta: Makes sense, then we should remove the Option from `Memory::relocate_value'`s input and output, as it serves no purpose

src/vm/runners/cairo_runner.rs
```diff
@@ -334,6 +337,53 @@ impl CairoRunner {
         vm.segments.get_memory_holes(&builtin_accessed_addresses)
     }
 
+    pub fn end_run(
+        &mut self,
+        _disable_trace_padding: Option<bool>,
+        disable_finalize_all: Option<bool>,
```
Oppen, line 343: These `Option<_>`s look weird. Why not bare `bool`?

fmoletta (approved):

src/vm/runners/cairo_runner.rs
```diff
@@ -334,6 +337,53 @@ impl CairoRunner {
...
+                .as_ref()
+                .ok_or_else::<VirtualMachineError, _>(|| {
+                    MemoryError::MissingAccessedAddresses.into()
+                })?;
+            let mut new_accessed_addresses = HashSet::with_capacity(accessed_addresses.len());
+
+            for addr in accessed_addresses {
+                let relocated_addr = vm
+                    .memory
+                    .relocate_value(Cow::Owned(addr.into()))?
```
Oppen, line 366: I think `.relocate_value` should receive a reference directly.

Oppen (changes requested):

src/vm/runners/cairo_runner.rs
azteca1998: In Python there's a default value, which can't be done in Rust.

Oppen: Then it makes no sense to attempt it. The point of default parameters is that you don't need to pass one. Here, you need to always pass one, it's now just more inconvenient since my "default" case is, rather than passing what I wanted, passing `None`, but when I want to be explicit now I need to do `Some<truth_value>`.

Oppen: Note we do accept this idiom in `cairo-rs-py` because PyO3 knows to consider `Option<_>` arguments in pymethods as optional arguments inside of Python.

src/vm/runners/cairo_runner.rs
```diff
@@ -376,10 +364,8 @@ impl CairoRunner {
             .map_err(VirtualMachineError::TracerError)?;
         vm.end_run(&self.exec_scopes)?;
 
-        if !disable_finalize_all {
-            vm.segments.compute_effective_sizes(&vm.memory);
-            self.run_ended = true;
-        }
+        vm.segments.compute_effective_sizes(&vm.memory);
+        self.run_ended = true;
```
Oppen, line 368: This should still be supported. What I meant we don't need to try is to use `Option<_>` to give a default, as the point is not having to pass any argument. Just pass the `bool`.

Oppen (changes requested):

Oppen (approved):

PR #508, opened 2022-10-27
Implement `BuiltinRunner::get_memory_accesses()`

azteca1998:
# Implement `BuiltinRunner::get_memory_accesses()`

## Description

Implementation of the method `get_memory_accesses()` for the implementors of the trait `BuiltinRunner`.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/runners/builtin_runner/ec_op.rs
+++ b/src/vm/runners/builtin_runner/ec_op.rs
@@ -192,7 +192,11 @@ impl BuiltinRunner for EcOpBuiltinRunner {
     fn get_memory_accesses(&self, vm: &VirtualMachine) -> Result<Vec<Relocatable>, MemoryError> {
         let segment_size = vm
             .segments
-            .get_segment_size(self.base as _)
+            .get_segment_size(
+                self.base
+                    .try_into()
+                    .map_err(|_| MemoryError::AddressInTemporarySegment(self.base))?,
+            )
             .ok_or(MemoryError::MissingSegmentUsedSizes)?;
 
         Ok((0..segment_size).map(|i| (self.base, i).into()).collect())
@@ -205,6 +209,7 @@ mod tests {
     use crate::utils::test_utils::*;
     use crate::vm::errors::memory_errors::MemoryError;
     use crate::vm::errors::runner_errors::RunnerError;
+    use num_bigint::Sign;
 
     #[test]
     fn point_is_on_curve_a() {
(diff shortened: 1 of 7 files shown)


src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -188,8 +189,13 @@ impl BuiltinRunner for EcOpBuiltinRunner {
         self
     }
 
-    fn get_memory_accesses(&self, _runner: &CairoRunner) -> std::collections::HashSet<Relocatable> {
-        todo!()
+    fn get_memory_accesses(&self, vm: &VirtualMachine) -> Result<Vec<Relocatable>, MemoryError> {
+        let segment_size = vm
+            .segments
+            .get_segment_size(self.base as _)
```
fmoletta, line 195: Use try_into() here instead of casting

fmoletta (changes requested):
Please add tests

src/vm/runners/builtin_runner/ec_op.rs
azteca1998: Should I avoid isize to usize casts even when it's impossible for a value to be negative?

src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -609,4 +620,29 @@ mod tests {
...
+        assert_eq!(
+            builtin.get_memory_accesses(&vm),
+            Ok(vec![
+                (builtin.base, 0).into(),
+                (builtin.base, 1).into(),
+                (builtin.base, 2).into(),
+                (builtin.base, 3).into(),
+            ]),
+        );
+    }
```
fmoletta, line 647: Please separate these into different tests, as they test different cases

fmoletta (changes requested):

src/vm/runners/builtin_runner/ec_op.rs
azteca1998: What's the criteria for deciding when to split a test or not?

fmoletta: Tests are meant to try specific cases, so if we make a change and a test fails we can know specifically which case is the one failing. For example here, there are 3 different cases, the first one doesn't have used_sizes, the second one has an empty vec for used_sizes and the third one has a filled used_sizes. If we were to have failures on cases 1 and 3, with this test we would only be able to see the failure in test 1, and we want as much information as possible when a functionality fails.

azteca1998: I think I more or less get it now. Thanks for the explanation.

src/vm/runners/builtin_runner/ec_op.rs
fmoletta: The type of base is isize, so we can't asume it won't be negative, we should instead create an issue to change it into usize

fmoletta (approved):

PR #512, opened 2022-10-27
feat: make BuiltinHintProcessor Sync by requiring HintFunc to be so

tdelabro:
I want to run multiple `cairo_run` in parallel, with the same hint executor.
To do so it has to be Sync, so the functions boxed into HintFunc must be Sync too.

### DIFF
--- a/src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
+++ b/src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
@@ -83,12 +83,13 @@ impl HintProcessorData {
 pub struct HintFunc(
     pub  Box<
         dyn Fn(
-            &mut VirtualMachine,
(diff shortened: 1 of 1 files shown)


Oppen (approved):

Oppen:
Quick question, these parallel runs would be over the same runner instance or different instances?

tdelabro:
> Quick question, these parallel runs would be over the same runner instance or different instances?

It will be different instances.
My usecase is a testing suite.
Rather than executing the main entrypoint, I execute multiple `Programs`, one for each testing function defined in the compiled file.
You can see the code here: https://github.com/onlydustxyz/cairo-foundry

azteca1998 (approved):

Oppen:
Hi @tdelabro. Could you merge main? It should fix the benchmark action (it's been changed for PRs to not attempt upload).

PR #515, opened 2022-10-28
Implement `get_memory_segment_addresses` for builtins

azteca1998:
# Implement `get_memory_segment_addresses` for builtins

## Description

Implementation of `get_memory_segment_addresses` for builtins.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/vm/runners/builtin_runner/hash.rs
+++ b/src/vm/runners/builtin_runner/hash.rs
@@ -13,7 +13,7 @@ use crate::vm::vm_memory::memory_segments::MemorySegmentManager;
 
 pub struct HashBuiltinRunner {
     pub base: isize,
-    stop_ptr: Option<Relocatable>,
+    stop_ptr: Option<usize>,
     name: String,
     _ratio: usize,
     cells_per_instance: usize,
@@ -102,13 +102,10 @@ impl BuiltinRunner for HashBuiltinRunner {
         self
     }
 
-    fn get_memory_segment_addresses(&self) -> HashMap<String, (Relocatable, Option<Relocatable>)> {
-        [(
-            self.name.to_string(),
-            ((self.base, 0).into(), self.stop_ptr.clone()),
-        )]
-        .into_iter()
-        .collect()
+    fn get_memory_segment_addresses(&self) -> HashMap<String, (isize, Option<usize>)> {
+        [(self.name.to_string(), (self.base, self.stop_ptr))]
+            .into_iter()
+            .collect()
     }
 }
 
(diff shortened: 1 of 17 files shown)


src/vm/runners/builtin_runner/hash.rs
```diff
@@ -12,22 +13,24 @@ use crate::vm::vm_memory::memory_segments::MemorySegmentManager;
...
     _ratio: usize,
     cells_per_instance: usize,
     _n_input_cells: usize,
-    _stop_ptr: Option<Relocatable>,
     verified_addresses: Vec<Relocatable>,
 }
 
 impl HashBuiltinRunner {
-    pub fn new(ratio: usize) -> Self {
+    pub fn new(name: impl Into<String>, ratio: usize) -> Self {
```
fmoletta, line 25: This extra field is not needed, the name of the builtin is fixed and doesn't depend on external factors, we can make it into a constant (defined in each builtin runner's file) if its more convenient.

src/vm/runners/builtin_runner/hash.rs
```diff
@@ -98,6 +101,15 @@ impl BuiltinRunner for HashBuiltinRunner {
     fn as_any(&self) -> &dyn Any {
         self
     }
+
+    fn get_memory_segment_addresses(&self) -> HashMap<String, (Relocatable, Option<Relocatable>)> {
+        [(
+            self.name.to_string(),
```
fmoletta, line 107: We can use a constant here instead of adding an extra field

src/vm/runners/builtin_runner/hash.rs
azteca1998: In the reference implementation, it [could also implement](https://github.com/starkware-libs/cairo-lang/blob/54d7e92a703b3b5a1e07e9389608178129946efc/src/starkware/starknet/core/os/class_hash.py#L83) a `"custom_hasher"`. That's why it's a field right now.

Should I disregard other possible implementations?

azteca1998: I guess not right now since it seems to be hardcoded to `petersen` in `cairo-rs` for now.

fmoletta: Yes, we currently don't support custom hashers in the HashBuiltinRunner

fmoletta (approved):

Oppen:
I'd like to see the call site for this. It looks quite a bit like either the caller knows the names and needs only the `(base, stop_ptr)` pair to build the map itself or that the runners should all have those fields irrespective of particular type and be implemented only once.
In any case, it seems like they could just be returning the `(name, (base, stop_ptr))` tuple rather than constructing a hashmap of one element each.

azteca1998:
It's implemented like this on Python, but it's true for every builtin (that we have, at least) that the dict has only a single item.

Should I change it into a single item or an iterator just in case?

juanbono:
Please resolve the conflicts @azteca1998

juanbono (approved):

Oppen:
> Should I change it into a single item?

I'd suggest that, at least for the runners themselves.

Issue #523, opened 2022-10-30
Need to have finer custom hints


Bernardstanislas:
Custom hints given to the `BuiltinHintProcessor` must match an exact string in order to be called.

This is very nice for a first step, but we need to have a bit more flexibility on the way the hint is called in a Cairo contract.

A next iteration on the custom hints would be to have them match some code based on a regex rather than an exact string.

An even more optimized version would be to use peg as @unbalancedparentheses mentionned.

PR #524, opened 2022-10-30
Regex hint processor

Bernardstanislas:
# Regex hint processer

## Description

Linked to #523.

Implements a `RegexHintProcessor`, which only holds custom hints. The hints match the code based on a regex.

This processor does not have any builtin, using it can be done in conjonction with the `BuiltinHintProcessor` with a composite pattern for example. This is left to the user project of the vm.

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "0.7.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Oppen:
I would suggest this is better implemented as part of a downstream project.
Regardless of that, I _strongly_ suggest to move the regex compilation to the hint compilation step, as it tends to be a major performance bottleneck.

Bernardstanislas:
Oh I see, I'll add it to [our repo](https://github.com/onlydustxyz/cairo-foundry) then.

Thanks for the feedback, agreed with the regex compilation remark :)

PR #531, opened 2022-10-31
Add method `BultinRunner::get_used_diluted_check_units`

fmoletta:
Add method `BultinRunner::get_used_diluted_check_units()`
Add unit tests

### DIFF
--- a/src/vm/runners/builtin_runner/bitwise.rs
+++ b/src/vm/runners/builtin_runner/bitwise.rs
@@ -116,7 +116,7 @@ impl BitwiseBuiltinRunner {
     }
 
     pub fn get_used_diluted_check_units(&self, diluted_spacing: u32, diluted_n_bits: u32) -> usize {
-        let total_n_bits = self.total_n_bits;
+        let total_n_bits = self.bitwise_builtin.total_n_bits;
         let mut partition = Vec::with_capacity(total_n_bits as usize);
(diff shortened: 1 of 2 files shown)


src/vm/runners/builtin_runner/bitwise.rs
```diff
@@ -98,6 +98,24 @@ impl BitwiseBuiltinRunner {
...
+                };
+            }
+        }
+        let partition_lengh = partition.len();
+        let num_trimmed = partition
+            .into_iter()
+            .filter(|elem| elem + diluted_spacing * (diluted_n_bits - 1) + 1 > total_n_bits)
+            .count();
+        4 * partition_lengh + num_trimmed
+    }
```
Oppen, line 118: I think this merits a comment explaining what it does. It's certainly not obvious.

fmoletta: The only documentation for this method is "Returns the number of diluted check units used by the builtin." in the class' method definition, so Im not really sure what it does either

Oppen: There's a tiny bit more context in the callers of this function, apparently those "diluted check units" are extra zeros in between information carrying bits. I'm guessing some error correcting code gets injected in those at some later point.

Oppen (approved):
The code seems equivalent to that of Cairo. It's sad that there's not much information about the intent, but LGTM.

juanbono (approved):

PR #534, opened 2022-11-01
Make execution scopes methods generic

tdelabro:
All those getters are reimplemented multiple times, for each type.
That a perfect usecase for Rust generic types


fmoletta (approved):
Looks great! Thanks for the PR!

PR #536, opened 2022-11-01
Add method `BuiltinRunners::get_range_check_usage`

fmoletta:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/runners/builtin_runner/range_check.rs
+++ b/src/vm/runners/builtin_runner/range_check.rs
@@ -109,8 +109,7 @@ impl RangeCheckBuiltinRunner {
     }
 
     pub fn get_range_check_usage(&self, memory: &Memory) -> Option<(BigInt, BigInt)> {
-        let mut rc_min: Option<BigInt> = None;
-        let mut rc_max: Option<BigInt> = None;
+        let mut rc_bounds: Option<(BigInt, BigInt)> = None;
         let range_check_segment = memory.data.get(self.base as usize)?;
         for value in range_check_segment {
             //Split val into n_parts parts.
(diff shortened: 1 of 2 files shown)


Oppen (approved):

src/vm/runners/builtin_runner/range_check.rs
```diff
@@ -105,6 +107,33 @@ impl RangeCheckBuiltinRunner {
...
+                        rc_max = rc_min.clone();
+                    }
+                    Some(_) => {
+                        rc_min = Some(min(rc_min?, part_val.clone()));
+                        rc_max = Some(max(rc_max?, part_val));
+                    }
+                }
+            }
+        }
+        Some((rc_min?, rc_max?))
```
azteca1998, line 135: There's some avoidable `?` checks. Maybe this is more efficient?

```suggestion
        let mut rc_bounds: Option<(BigInt, BigInt)> = None;
        let range_check_segment = memory.data.get(self.base as usize)?;
        for value in range_check_segment {
            //Split val into n_parts parts.
            for _ in 0..self.n_parts {
                let part_val = value
                    .as_ref()?
                    .get_int_ref()
                    .ok()?
                    .mod_floor(&self.inner_rc_bound);
                rc_bounds = Some(match rc_bounds {
                    None => (part_val.clone(), part_val),
                    Some((rc_min, rc_max)) => {
                        let rc_min = min(rc_min, part_val.clone());
                        let rc_max = max(rc_max, part_val);

                        (rc_min, rc_max)
                    }
                });
            }
        }
        rc_bounds
```

azteca1998 (approved):

src/vm/runners/builtin_runner/range_check.rs
Oppen: We can probably do `min(&rc_min, &part_val)` to avoid clones.

azteca1998: I thought about that too, but that returns a reference which must be cloned, ending up with two `.clone()`.
Maybe with `Cow`? Though if it ends up working it won't be as pretty.

Oppen: No, you're right.

PR #538, opened 2022-11-02
Add WebAssembly example project and update docs.

azteca1998:
# Add WebAssembly example project and update docs

## Description

Added a WebAssembly example project and updated the docs.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


Oppen (approved):
Beautiful! Since this is an example, I suggest excluding it from coverage reporting. Otherwise, we'll need some tests.

azteca1998:
How does excluding something from coverage work?

Oppen:
Modify `.tarpaulin.toml` to look like this:
```toml
[cairo-rs]
exclude-files = ["wasm-examples"]

[report]
output-dir = "target/tarpaulin"
out = ["Html", "Xml"]
```

The features is not really documented, I had to dig the sources to find how to do it via the config file. But at least it works.

juanbono (approved):

PR #546, opened 2022-11-02
Initialize segments of every builtin and add included field

Jrigada:
# Initialize segments of every builtin and add included field

## Description

We had to make some changes to the security.rs tests because now for them to work as intended we need to have the BuiltinRunner::finalize_segments() implemented. That is currently in another PR and we will revert the changes to this tests after that is merged. 
https://github.com/lambdaclass/cairo-rs/pull/542

Description of the pull request changes and motivation.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #559, opened 2022-11-04
Add proof mode logic to CairoRunner::new()

fmoletta:
Depends on #540 
Add proof-mode logic to CairoRunner::new()
Fix confusion with builtin runner creation (BuiltinRunners are created only when the are in the layout, and are either included or the runner is in proof-mode. Previously, we made a change to create them if they are in the layout, leading to unused builtins being created and builtin bases being changed)
Revert test changes from #546


Jrigada (approved):

juanbono (approved):

azteca1998 (approved):

azteca1998 (approved):
Nice catch.

PR #562, opened 2022-11-07
Implement `Memory::relocate_memory`.

azteca1998:
# Implement `Memory::relocate_memory`

## Description

Implementation `Memory::relocate_memory`.
Fix `CairoRunner::end_run()`.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/vm_memory/memory.rs
+++ b/src/vm/vm_memory/memory.rs
@@ -171,30 +171,29 @@ impl Memory {
                 },
             ));
         for (addr, value) in data_iter {
+            // After the following check, addr.segment_index cannot be negative, therefore it is
+            // safe to cast to `usize`.
+            if addr.segment_index.is_negative() {
+                continue;
+            }
+
             let value = match value {
                 Some(x) => x,
                 None => continue,
             };
 
-            let mut new_addr: Relocatable = self
+            let new_addr: Relocatable = self
                 .relocate_value(&MaybeRelocatable::RelocatableValue(addr))
                 .into_owned()
                 .try_into()?;
             let new_value = self.relocate_value(&value).into_owned();
 
-            let target_memory = match new_addr.segment_index.is_negative() {
-                false => &mut self.data,
-                true => {
-                    new_addr.segment_index = -(new_addr.segment_index + 1);
-                    &mut self.temp_data
-                }
-            };
-
-            if new_addr.segment_index as usize >= target_memory.len() {
-                target_memory.resize(new_addr.segment_index as usize + 1, Vec::new());
+            if new_addr.segment_index as usize >= self.data.len() {
+                self.data
+                    .resize(new_addr.segment_index as usize + 1, Vec::new());
             }
 
-            let segment_data = &mut target_memory[new_addr.segment_index as usize];
+            let segment_data = &mut self.data[new_addr.segment_index as usize];
             if new_addr.offset as usize >= segment_data.len() {
                 segment_data.resize(new_addr.offset + 1, None);
             }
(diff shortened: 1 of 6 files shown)


azteca1998:
[Memory::relocate_memory() - end_run()](https://trello.com/c/iNKZz5Io/74-memoryrelocatememory-endrun)

src/vm/vm_memory/memory.rs
```diff
@@ -149,7 +224,7 @@ impl Memory {
             return Err(MemoryError::NonZeroOffset(src_ptr.offset));
         }
 
-        let segment_index = -src_ptr.segment_index as usize;
+        let segment_index = -(src_ptr.segment_index + 1) as usize;
```
fmoletta, line 227: Why did this change?

azteca1998: Before, the first temporary segment (-1) had key -1, not zero, which is unintuitive and may cause problems in the extreme case that there are `isize::MIN` temporary segments.

fmoletta: I would also recommend changing the test` relocate_value_mayberelocatable_temporary_segment_rules`, or creating a parallel test that uses add_relocation_rule() to add rules as opposed to initializing it directly

fmoletta: Regarding my first reply, I also think that having this extra logic makes relocation rules more confusing, and what we gain from it (being able to host an extra relocation_rule) doesn't quite make up for the rise in complexity (shifting relocation rules indexes to the left)

azteca1998: The snippet you quoted is from `add_relocation_rule()`, so it's already reflected in that method. Of course, it's also reflected in the added method (`Memory::relocate_memory()`). The idea is that this change is only about the internal representation and shouldn't affect anything outside `Memory`.

I completely agree about the test case. I've fixed it already.

My reasoning is that by considering the normal and temporary segments as separate lists that begin with index zero, it wouldn't make sense to have it begin at one at `HashMap` where the key contains exclusively temporary segment indices. In other words, I think it's better to keep it zero-based because `temp_data` is a zero-based list, and the key links to that structure's indices. If it's still confusing I can explicitly document it with a comment or revert it as it was before.

fmoletta: Adding comments  on all affected methods should be enough to prevent future confusion

Juan-M-V (approved):
Good work!

src/vm/vm_memory/memory.rs
azteca1998: Done. I've also made them private (`relocation_rules` and `validation_rules`) since they are neither accessed nor meant to be from the outside. This way the change is kept strictly internal.

src/vm/vm_memory/memory.rs
```diff
@@ -129,6 +133,81 @@ impl Memory {
...
+                .try_into()?;
+            let new_value = self.relocate_value(&value)?.into_owned();
+
+            let target_memory = match new_addr.segment_index.is_negative() {
+                false => &mut self.data,
+                true => {
+                    new_addr.segment_index = -(new_addr.segment_index + 1);
+                    &mut self.temp_data
+                }
+            };
```
Oppen, line 193:
```suggestion
            if new_addr.segment_index.is_negative() {
                new_addr.segment_index = -(new_addr.segment_index + 1);
            }
            let target_memory = &mut self.temp_data;
```

Oppen (changes requested):

src/vm/vm_memory/memory.rs
azteca1998: Doesn't this change the functionality? The idea was that since starknet had unified memory, then we should be able to work with both normal and temporary segments.

Oppen (approved):
LGTM

PR #565, opened 2022-11-07
Separate benchmark report upload in  a different job

fmoletta:
**Separate uploading benchmark and codecov reports on separate jobs.**
This PR aims to reduce CI time in case of unexpected failures by separating the uploading tasks in different jobs which can be restarted. Creating each report takes a considerable amount of time, and uploading these reports may fail due to external reasons. When this happens, we have to re-run the whole job, when we could just re-run the uploading of the results
This is the benchmark_pr workflow counterpart of #570

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -29,6 +29,22 @@ jobs:
       run: pip install ecdsa fastecdsa sympy cairo-lang==0.9.1
     - name: Run benchmark
       run: make benchmark-action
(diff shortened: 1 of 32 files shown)


Oppen (changes requested):
The perf report for PRs makes sense only to catch regressions early, which the alert should sort out. For showing users on the site only what they would actually use (i.e., releases and, since those aren't happening yet, commits to main) makes sense. This distinction was handled already by the split between merge and PR jobs.

PR #569, opened 2022-11-07
Finish `CairoRunner::end_run()`'s `proof_mode` section.

azteca1998:
# Finish `CairoRunner::end_run()`'s `proof_mode` section

## Description

Finish `CairoRunner::end_run()`'s `proof_mode` section.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


azteca1998:
[Expose runner.run_from_entrypoint() in cairo-rs-py](https://trello.com/c/OQoSXygP/78-expose-runnerrunfromentrypoint-in-cairo-rs-py)

azteca1998:
[Fix `CairoRunner::end_run()` with the `proof_mode` section.](https://trello.com/c/81b6ZQcf/81-fix-cairorunnerendrun-with-the-proofmode-section)

src/vm/runners/cairo_runner.rs
```diff
@@ -2959,18 +2984,47 @@ mod tests {
...
         assert!(!cairo_runner.run_ended);
     }
 
+    #[test]
+    fn end_run_proof_mode_insufficient_allocated_cells() {
+        let program = Program::from_file(
+            Path::new("cairo_programs/proof_programs/fibonacci.json"),
+            "main",
+        )
+        .unwrap();
```
Juan-M-V, line 3012: If you can, change this `unwrap` for `expect`

Juan-M-V (approved):

src/vm/runners/cairo_runner.rs
```diff
@@ -593,11 +594,26 @@ impl CairoRunner {
...
+                if let Err(e) = self.check_used_cells(vm) {
+                    if e == MemoryError::InsufficientAllocatedCells.into() {
+                        break;
+                    }
+                }
+
+                self.run_for_steps(1, vm, hint_processor)?;
+                self.run_until_next_power_of_2(vm, hint_processor)?;
+            }
         }
```
fmoletta, line 614: Is this behavior correct?, in the original vm the loop ends when check_used_cells returns True (our counterpart would be Ok()), but we end the loop when we get an InsufficientAllocatedCells error.

azteca1998: Yes, should be correct since `check_used_cells()` does catch `InsufficientAllocatedCells()` to return `False` which then breaks this loop.

Maybe it'd be better to move that special case into `check_used_cells()` as in Python?

fmoletta: What im worried about is that in python the loop runs while check_used_cells() returns False, and ends when it returns True `while not self.check_used_cells():`, but we are breaking the loop when it returns False (aka InsufficientAllocatedCells)

Oppen (commented):
Do the programs actually differ or is it only a compilation flag?
If the latter, just compiling the regular tests with different flags should be enough.

azteca1998:
Only the flags, but since it's using `wildcard` and there'll probably be more proof_programs in the future I've decided it'll probably be better to prepare it now.

Should I make a special (non-wildcard) rule instead of a `proof_programs` section?

src/vm/runners/cairo_runner.rs
azteca1998: Oh, you're right. That makes more sense. Thanks.

azteca1998: Fixed that and another bug where I forgot to return when an error different from that special case is returned.

src/vm/runners/cairo_runner.rs
```diff
@@ -2959,18 +2984,47 @@ mod tests {
...
+        .unwrap();
+
+        let hint_processor = BuiltinHintProcessor::new_empty();
+        let mut cairo_runner = cairo_runner!(program, "all", true);
+        let mut vm = vm!();
+
+        let end = cairo_runner.initialize(&mut vm).unwrap();
+        cairo_runner
+            .run_until_pc(end, &mut vm, &hint_processor)
+            .unwrap();
```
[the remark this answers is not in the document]
azteca1998: Done.

src/vm/runners/cairo_runner.rs
azteca1998: Done.

Oppen (commented):
Do the programs actually differ or is it only a compilation flag?
If the latter, just compiling the regular tests with different flags should be enough.

Oppen (commented):
In that case I think it's better to add the old cairo programs as symlinks to the new folder. Less risk of accidental divergence.

fmoletta (approved):

PR #570, opened 2022-11-07
Separate codecov-report upload in a different job

fmoletta:
Separate the upload of   codecov report on from other workflow tasks
This PR aims to reduce CI time in case of unexpected failures by separating the uploading task in  a different job which can be quickly re-run in case of failure. Creating each report takes a considerable amount of time, and uploading those reports may fail due to external reasons. When this happens, we have to re-run the whole job, when we could just re-run the uploading of the results.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,9 +38,23 @@ jobs:
       run: make clippy
     - name: Coverage
       run: make coverage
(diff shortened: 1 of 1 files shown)


Juan-M-V (approved):

Jrigada (approved):
LGTM

PR #576, opened 2022-11-08
Add relocate_segment hint implementation

pefontana:
# Add relocate_segment hint implementation

## Description
* Implement hints:
  * %{ memory.add_relocation_rule(src_ptr=ids.src_ptr, dest_ptr=ids.dest_ptr)%}
  * %{ids.temporary_array = segments.add_temp_segment()%} (Not in common library, just for testing purposes )

Description of the pull request changes and motivation.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added
- [x] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.

PR #577, opened 2022-11-08
Relocate pointer in temporary segments

pefontana:
# Relocate pointer in temporary segments

Add the feature to relocate temporary segments

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/types/relocatable.rs
+++ b/src/types/relocatable.rs
@@ -361,7 +361,7 @@ pub fn relocate_value(
         MaybeRelocatable::RelocatableValue(relocatable) => {
             let segment_index = if relocatable.segment_index.is_negative() {
                 relocation_rules
-                    .get(&(relocatable.segment_index.abs_diff(0)))
+                    .get(&(relocatable.segment_index.abs() as usize))
                     .ok_or(MemoryError::TemporarySegmentWithoutRelocationAddreess(
                         relocatable.segment_index,
                     ))?
--- a/src/types/relocatable.rs
+++ b/src/types/relocatable.rs
@@ -5,7 +5,7 @@ use crate::{
 use num_bigint::BigInt;
 use num_integer::Integer;
 use num_traits::{FromPrimitive, Signed, ToPrimitive};
-use std::ops::Add;
+use std::{collections::HashMap, ops::Add};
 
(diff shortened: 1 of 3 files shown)


src/types/relocatable.rs
```diff
@@ -354,14 +354,21 @@ impl<'a> Add<usize> for &'a Relocatable {
 pub fn relocate_value(
     value: MaybeRelocatable,
     relocation_table: &Vec<usize>,
+    relocation_rules: &HashMap<usize, Relocatable>,
```
Oppen, line 357: It doesn't look like we're ever using the `.offset` field, so maybe just store a `usize` for the index here?

src/types/relocatable.rs
```diff
@@ -354,14 +354,21 @@
...
     match value {
         MaybeRelocatable::Int(num) => Ok(num),
         MaybeRelocatable::RelocatableValue(relocatable) => {
-            let segment_index: usize = relocatable
-                .segment_index
-                .try_into()
-                .map_err(|_| MemoryError::AddressInTemporarySegment(relocatable.segment_index))?;
+            let segment_index = if relocatable.segment_index.is_negative() {
+                relocation_rules
+                    .get(&(relocatable.segment_index.abs_diff(0)))
```
Oppen, line 364: I think this is the same as `.segment_index.abs()`.

Oppen (commented):
LGTM, some minor stuff tho.

src/types/relocatable.rs
pefontana: Thanks for this one @Oppen. 
I found this a little bit strange, so I made some testing and found that we are having a bug in some edge cases:
If we relocate a temporary segment to an address with offset != 0, we are having some differences with the Python VM `relocate_segment(src_ptr=temporary_array, dest_ptr=array + 5)`
I think we need to make some modifications to the Memory::relocate_value and the Relocatable::relocate_value
I will tackle this issue as soon as we merge the #576

pefontana: https://github.com/lambdaclass/cairo-rs/issues/581

Oppen (approved):
LGTM

fmoletta (approved):

PR #582, opened 2022-11-09
Add check + Informative error for when there are builtins in the program that are not present in the layout

fmoletta:
When there are builtins used in the compiled program but the layout chosen doesn't include them, these builtins are not added to the VM, but no error is shown, leading to unpredictable errors later on.
This PR adds a check on CairoRunner::initialize_builtins() that checks this and returns an informative error message in that case

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -177,10 +177,12 @@ impl CairoRunner {
         let program_builtins: HashSet<&String> =
             self.program.builtins.iter().collect::<HashSet<&String>>();
         // Get the builtins that belong to the program but weren't inserted (those who dont belong to the instance)
-        let mut difference = program_builtins.difference(&inserted_builtins).peekable();
-        if difference.peek().is_some() {
+        if !program_builtins.is_subset(&inserted_builtins) {
             return Err(RunnerError::NoBuiltinForInstance(
-                difference.map(|x| (&(**x).clone()).clone()).collect(),
+                program_builtins
+                    .difference(&inserted_builtins)
+                    .map(|x| (&(**x).clone()).clone())
+                    .collect(),
                 self.layout._name.clone(),
             ));
         }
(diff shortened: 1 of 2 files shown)


Oppen (approved):
LGTM except from a minor comment. Feel free to ignore it.

SantiagoPittella (approved):

src/vm/runners/cairo_runner.rs
```diff
@@ -170,6 +170,25 @@ impl CairoRunner {
...
+        if !difference.is_empty() {
+            return Err(RunnerError::NoBuiltinForInstance(
+                difference
+                    .iter()
+                    .map(|x| (&(**x).clone()).clone())
+                    .collect(),
+                self.layout._name.clone(),
+            ));
+        }
+
```
Oppen, line 191: The following is more memory efficient as it avoids creation of a third `HashSet`:
```suggestion
        let difference = program_builtins
            .difference(&inserted_builtins)
            .peekable();
        if difference.peek().is_some() {
            return Err(RunnerError::NoBuiltinForInstance(
                difference
                    .map(|x| (&(**x).clone()).clone())
                    .collect(),
                self.layout._name.clone(),
            ));
        }

```

Actually we can do even better:
```suggestion
        if !inserted_builtins.is_subset(&program_builtins) {
            return Err(RunnerError::NoBuiltinForInstance(
                program_builtins
                    .difference(&inserted_builtins)
                    .map(|x| (&(**x).clone()).clone())
                    .collect(),
                self.layout._name.clone(),
            ));
        }

```

Oppen (approved):
LGTM except from a minor comment. Feel free to ignore it.

src/vm/runners/cairo_runner.rs
fmoletta: Added suggested change(971773), but swapped` inserted_builtins.is_subset(&program_builtins)` for `program_builtins.is_subset(&inserted_builtins)`, as we want to make sure all the builtins in the compiled program are part of (or subset of) the inserted builtins

Oppen: Right, I misplaced those.

PR #600, opened 2022-11-11
Rust cache tests

klaus993:


.github/workflows/bench_pull_request.yml
```diff
@@ -2,7 +2,7 @@ name: benchmark_pr
 
 on:
   pull_request:
-    branches: [ '*' ]
+    branches: [ 'asfd' ]
```
unbalancedparentheses, line 5: @klaus993 change this

Oppen:
Triggered manually with clean cache to compare timings. With cached deps it took 15m to run. Without it it takes 21m.

This is not currently caching the target directory, so it can improve a bit more. For that we'll need to make it build all targets so other workflows can benefit as well.

PS: sorry @klaus993 for destroying the other PR, I misunderstood the "tests" in the title and thought it was for testing the cache action :man_facepalming:

Juan-M-V (approved):

azteca1998 (approved):

PR #604, opened 2022-11-14
Mark program segment addresses as accessed in vm_initialize method

entropidelic:
In the Python implementation, the program segment addresses are set as accessed in the VM initialization. See https://github.com/starkware-libs/cairo-lang/blob/f951a669e264796d53bf7a3cfd23d5bc1f6e54b8/src/starkware/cairo/lang/vm/vm_core.py#L130

This PR adds this at the `initialize_vm()` method. It is necessary because otherwise those addresses are considered memory holes when running StarkNet contracts.

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -362,16 +362,16 @@ impl CairoRunner {
         }
 
         // Mark all addresses from the program segment as accessed
-        let mut initial_accessed_addresses: Vec<Relocatable> = Vec::new();
         let prog_segment_index = self
             .program_base
             .as_ref()
             .unwrap_or(&Relocatable::from((0, 0)))
             .segment_index;
 
-        for offset in 0..self.program.data.len() {
-            initial_accessed_addresses.push(Relocatable::from((prog_segment_index, offset)));
-        }
+        let initial_accessed_addresses = (0..self.program.data.len())
+            .map(|offset| Relocatable::from((prog_segment_index, offset)))
+            .collect();
+
         vm.accessed_addresses = Some(initial_accessed_addresses);
 
         vm.memory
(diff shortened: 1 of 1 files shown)


src/vm/runners/cairo_runner.rs
```diff
@@ -360,6 +360,20 @@ impl CairoRunner {
...
             builtin.add_validation_rule(&mut vm.memory)?;
         }
+
+        // Mark all addresses from the program segment as accessed
+        let mut initial_accessed_addresses: Vec<Relocatable> = Vec::new();
+        let prog_segment_index = self
+            .program_base
+            .as_ref()
+            .unwrap_or(&Relocatable::from((0, 0)))
+            .segment_index;
```
Oppen, line 370: Is this not guaranteed to still be segment `0`? If that's so, the special case is not necessary.

src/vm/runners/cairo_runner.rs
```diff
@@ -360,6 +360,20 @@
...
+
+        for offset in 0..self.program.data.len() {
+            initial_accessed_addresses.push(Relocatable::from((prog_segment_index, offset)));
+        }
```
Oppen, line 374: This looks like something that can be done with `.collect()` instead.

src/vm/runners/cairo_runner.rs
```diff
@@ -1328,6 +1342,31 @@
         assert_eq!(return_pc, Relocatable::from((1, 0)));
     }
 
+    #[test]
+    fn initialize_vm_program_segment_accessed_addrs() {
+        // This test checks that all addresses from the program segment are marked as accessed at VM initialization.
+        // The fibonacci program has 24 instructions, so there should be 24 accessed addresses,
+        // from (0, 0) to (0, 23).
```
Oppen, line 1349: What about a test that any other segment is not marked as accessed at this point?

Oppen (commented):
Two minor suggestions and a more important question.

src/vm/runners/cairo_runner.rs
entropidelic: I don't really know if there could be another program base, but at line 357 this function does
```rust
        vm._program_base = Some(MaybeRelocatable::from(
            self.program_base.as_ref().ok_or(RunnerError::NoProgBase)?,
        ));
```

so in reality, `program_base` won't be None in that point.

What I really intended is to use the program segment index instantiated in the CairoRunner.
Do you think it is better to hardcode a 0?

src/vm/runners/cairo_runner.rs
entropidelic: done!

src/vm/runners/cairo_runner.rs
Oppen: Not if we're not sure.

Oppen (approved):
LGTM

PR #614, opened 2022-11-16
Use mod_floor when inserting into dict

Juan-M-V:
When creating a dictionary with negative values, the felts no longer remain negative. This fixes a previously untested bug that was found in cairo-rs-py


juanbono (approved):
LGTM

src/hint_processor/builtin_hint_processor/dict_manager.rs
```diff
@@ -88,6 +90,12 @@ impl DictManager {
             ));
         };
 
+        for val in initial_dict.values_mut() {
+            if val.is_negative() {
+                *val = val.mod_floor(&vm.prime);
+            }
+        }
+
```
fmoletta, line 98: Is it necessary to mutate the initial_dict we receive? Can't we mutate our copy of the dictionary, as we are already cloning it afterwards

Juan-M-V: Resolved: https://github.com/lambdaclass/cairo-rs/pull/614/commits/9041c0cd93f048da6a252f3ec7aa0742d6321ffe

src/hint_processor/builtin_hint_processor/dict_manager.rs
```diff
@@ -88,9 +90,19 @@ impl DictManager {
...
+        let mut floored_initial = HashMap::new();
+        for (key, val) in initial_dict.into_iter() {
+            let floored_val = if val.is_negative() {
+                val.mod_floor(&vm.prime)
+            } else {
+                val
+            };
+            floored_initial.insert(key, floored_val);
+        }
+
```
fmoletta, line 102: Can we use map for this?

Juan-M-V: Resolved: https://github.com/lambdaclass/cairo-rs/pull/614/commits/99220f47aac9f24a5dd0f20b533d8a512f66b6e9

Oppen (approved):

PR #619, opened 2022-11-17
Handle new cairo syntax

karol-bisztyga:
## Description

Adding the support for new cairo-lang syntax for the compiled (JSON) contracts. I'm not fully aware of all possible changes but from what I observed the main difference is that some properties have been moved inside the `program` section.

So, in the old syntax we'd have:
```
{
  "prime": xxx,
  "data": [...],
  ...
}
```

But in the new syntax we have:
```
{
  "program": {
    "prime": xxx,
    "data": [...],
    ...
  },
  ...
}
```

The idea is to parse the JSON by looking for certain fields in the `program` section.

Issue: https://github.com/lambdaclass/cairo-rs/issues/620

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,10 @@ PROOF_TEST_FILES:=$(wildcard $(PROOF_TEST_DIR)/*.cairo)
 COMPILED_PROOF_TESTS:=$(patsubst $(PROOF_TEST_DIR)/%.cairo, $(PROOF_TEST_DIR)/%.json, $(PROOF_TEST_FILES))
 
 $(PROOF_TEST_DIR)/%.json: $(PROOF_TEST_DIR)/%.cairo
-	cairo-compile --proof_mode $< --output $@
+	starknet-compile --proof_mode $< --output $@
 
 $(TEST_DIR)/%.json: $(TEST_DIR)/%.cairo
-	cairo-compile --cairo_path="$(TEST_DIR):$(BENCH_DIR)" $< --output $@
+	starknet-compile --cairo_path="$(TEST_DIR):$(BENCH_DIR)" $< --output $@
 
 $(TEST_DIR)/%.rs.trace $(TEST_DIR)/%.rs.memory: $(TEST_DIR)/%.json build
 	./target/release/cairo-rs-run --layout all $< --trace_file $@ --memory_file $(@D)/$(*F).rs.memory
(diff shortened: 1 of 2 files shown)


karol-bisztyga (commented):
I wasn't sure what would be the best approach to do this. I could see two ways to go here:
- drop the support for the old syntax and replace it with the new syntax entirely
- just add the new syntax support and keep the old one

I decided to go with the latter so we can try to parse the JSON file first with the old model and then if that fails, fall back to the new one. I find it more secure and flexible, and probably more suitable for you.

I didn't write tests, for now, once the idea gets accepted I can add tests.

karol-bisztyga:
Please, let me know if docs should be updated.

src/types/program.rs
```diff
@@ -32,10 +32,17 @@ impl Program {
...
     ) -> Result<Program, ProgramError> {
-        deserialize_program(reader, entrypoint)
+        match deserialize_program(reader, entrypoint) {
+            Ok(program) => return Ok(program),
+            Err(e) => {
+                println!("{}", e);
+                println!("Failed to deserialize the program using the old cairo-lang syntax, falling back to the new syntax");
+            }
+        }
+        deserialize_program_new_syntax(reader, entrypoint)
```
karol-bisztyga, line 45: Please let me know whether this is okay. The goal is to fall back to the new syntax automatically and to keep the error from the old syntax deserialization at the same time.

I'm not sure about the output, but it feels like a good idea to let users know that one deserialization has failed, so I made it explicit.

juanbono:
Hi! First of all, thanks for your contribution :D !

To validate that this change doesn't break anything please change the cairo programs used for testing to the new syntax. You can use this PR as inspiration https://github.com/lambdaclass/cairo-rs/pull/420 . Or you can also use the cairo-migrate script to port them.
Also please update the [Makefile targets](https://github.com/lambdaclass/cairo-rs/blob/c7c3e4fca78d38da968e92c6586ec3a518697745/Makefile#L41) to install Cairo 0.10.2 instead of Cairo 0.9.1

karol-bisztyga:
Thank you for the response!

I updated the makefile (I changed the versions as well as `cairo-compile` => `starknet-compile`, please, let me know if it's ok) and ran the tests. The tests gave me the same output as before these changes.

Oppen:
> I updated the makefile (I changed the versions as well as cairo-compile => starknet-compile, please, let me know if it's ok) and ran the tests. The tests gave me the same output as before these changes.

This should update only the version of `cairo-compile`. Its output is still compatible with our code, while it's `starknet-compile` that produces a different bytecode. It's two different concerns that may be addressed in different PRs with different constraints.

Oppen:
This was implemented as part of a different PR for the programs. The bytecode itself didn't change, but the StarkNet one is supported by starknet_in_rust.

PR #625, opened 2022-11-18
Implement keccak builtin

SantiagoPittella:
# IMPLEMENT KECCAK BUILTIN

# TO DO
- [x] Fix clippy errors.
- [x] Add tests for deduce_memory_cell

# About cairo-rs-py
This PR will generate issues in `cairo-rs-py`. Fix is already done and I'm waiting for this being merged to open a PR.


src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+            cells_per_instance: 50,
+            stop_ptr: None,
+            verified_addresses: Vec::new(),
+            _included: included,
+            instances_per_component: 1,
+            state_rep: vec![
+                64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64, 64,
+                64, 64, 64, 64,
+            ],
+        }
```
fmoletta, line 44: Where did this come from?

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+    instances_per_component: u32,
+}
+
+impl KeccakBuiltinRunner {
+    pub(crate) fn new(instance_def: &KeccakInstanceDef, included: bool) -> Self {
+        KeccakBuiltinRunner {
+            base: 0,
+            ratio: instance_def._ratio,
+            n_input_cells: 25,
+            cells_per_instance: 50,
```
fmoletta, line 35: Where did these come from?

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+
+        for i in 0..self.n_input_cells {
+            present_in_memory.push(memory.data.contains(&vec![Some(MaybeRelocatable::from(
+                first_input_addr.clone() + i as usize,
+            ))]))
+        }
+
+        if present_in_memory.iter().all(|&x| x) {
+            return Ok(None);
+        }
```
fmoletta, line 103: Is this  block of code is trying to imitate the following functionality In python?
```
if not all(first_input_addr + i in memory for i in range(self.n_input_cells)):
 return
```
If so, the there are a couple of things that may be wrong here:
- In the python code, they return if any element is not true while we are returning if all the elements are true
- We are checking if our memory contains the addresses while in python they are checking if the addresses are in memory. While these two sound similar, they are not the same. In python, the memory consists of a dictionary, mapping addresses (keys) to memory values (values),  in our implementation we are using a marix (Vec of Vec), where the addresses are indicated by the indexes. In the python code, when they check for address in memory, they are checking that the memory dictionary contains the key address, while in our code we are checking if memory contains the memory value address, leading to very different results.

On a side note, we could carry out this check without creating a Vec to hold the every check's result. as we want to make sure that every address is in memory, we can just return when we first find an address that is not in memory.

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+
+        if let Some((i, bits)) = (0..self.state_rep.len()).zip(self.state_rep.clone()).next() {
```
fmoletta, line 105: This line is a bit confusing, could it be replaced by `for (i,  bits) in self.state_rep.iter().enumerate()` ?

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+            let value1 = memory
+                .get(&(first_input_addr.clone() + i))
```
fmoletta, line 107: Is this clone necessary?

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+                .map_err(RunnerError::FailedMemoryGet)?
+                .ok_or(RunnerError::NonRelocatableAddress)?;
+
+            let val = match value1.into_owned() {
+                MaybeRelocatable::Int(val) => val,
+                _ => return Err(RunnerError::BaseNotFinished),
```
fmoletta, line 113: Is this error type correct?

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+            };
+            if !(bigint!(0_i32) <= val && val < (bigint!(2_i32)).pow(bits as u32)) {
+                return Err(RunnerError::BaseNotFinished);
```
fmoletta, line 116: Same here

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,610 @@
...
+            .get_segment_size(
+                base.try_into()
+                    .map_err(|_| MemoryError::AddressInTemporarySegment(base))?,
+            )
+            .ok_or(MemoryError::MissingSegmentUsedSizes)?;
+
+        Ok((0..segment_size).map(|i| (base, i).into()).collect())
+    }
+}
+
```
fmoletta, line 237: The implementation of `get_used_diluted_check_units `is missing here, it should also me added to the method  `get_used_diluted_check_units` on the BuiltinRunner enum definition (mod.rs)

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,606 @@
...
+    instances_per_component: u32,
+}
+
+impl KeccakBuiltinRunner {
+    pub(crate) fn new(instance_def: &KeccakInstanceDef, included: bool) -> Self {
+        KeccakBuiltinRunner {
+            base: 0,
+            ratio: instance_def._ratio,
+            n_input_cells: 25,
+            cells_per_instance: 50,
```
fmoletta, line 35: Shouldn't this be `cells_per_instance: instance_def.cells_per_builtin()`

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -0,0 +1,606 @@
...
+    state_rep: Vec<u32>,
+    instances_per_component: u32,
+}
+
+impl KeccakBuiltinRunner {
+    pub(crate) fn new(instance_def: &KeccakInstanceDef, included: bool) -> Self {
+        KeccakBuiltinRunner {
+            base: 0,
+            ratio: instance_def._ratio,
+            n_input_cells: 25,
```
fmoletta, line 34: Shouldn't this be `n_input_cells: safe_div(instance_def.cells_per_builtin(), 2)`

fmoletta (approved):

Juan-M-V (approved):

PR #632, opened 2022-11-25
Change DiffAssertValues error message

entropidelic:
This change is needed so that some Protostar tests pass.

The test asserts for the error message and in the Python version, and the `res` and `dst` are flipped.


fmoletta (approved):
Nice!

PR #635, opened 2022-11-28
Add offsets as primitive

Oppen:
Offsets in instructions are specified to be signed 16 bits integers. We were keeping them as `BigInt`s and operating with those, heavily slowing down execution. By converting them to `isize` we save about 30% of execution time, up to 50% in some cases.


fmoletta (approved):
Great!

azteca1998 (approved):
Note that using `i16` for `off0`, `off1` and `off2` would be enough.

Oppen:
> Note that using `i16` for `off0`, `off1` and `off2` would be enough.

Yep. But doing it this way saves me a little typing due to the left hand argument being `isize` IIRC. I can try later with `i16`.

Oppen:
I'm running with `i16` to test. It seems to give a <1% improvement, but it's run only two of the benchmarks as of now.

Oppen:
L2 cache access seems to significantly degrade. Everything else reported by IAI seems to take about 0.1-0.3 percent less. I think `isize` will perform better.

PR #637, opened 2022-11-30
Make `compute_effective_sizes()` accessible from outside `cairo-rs`

azteca1998:
# Make `compute_effective_sizes()` accessible from outside `cairo-rs`.

## Description

Needed to be able to test `PySegmentManager::get_segment_used_size()` on `cairo-rs-py`.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/vm/vm_core.rs
+++ b/src/vm/vm_core.rs
@@ -857,6 +857,12 @@ impl VirtualMachine {
     ) -> Result<MaybeRelocatable, VirtualMachineError> {
         self.segments.gen_arg(arg, prime, &mut self.memory)
     }
+
+    /// Proxy to MemorySegmentManager::compute_effective_sizes() to make it accessible from outside
+    /// cairo-rs.
+    pub fn compute_effective_sizes(&mut self) -> &Vec<usize> {
+        self.segments.compute_effective_sizes(&self.memory)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


src/vm/vm_core.rs
```diff
@@ -3829,4 +3835,24 @@ mod tests {
...
+        vm.load_data(
+            &segment.into(),
+            vec![
+                mayberelocatable!(1),
+                mayberelocatable!(2),
+                mayberelocatable!(3),
+                mayberelocatable!(4),
+            ],
+        )
+        .expect("Could not load data into memory.");
```
SantiagoPittella, line 3854: Let's use `.unwrap()` here, in order in order to show the real panic error.

SantiagoPittella (approved):
LGTM, just a small comment. 🚀

src/vm/vm_core.rs
azteca1998: I'm not sure an `.unwrap()` would be better in this case. [Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fc8f63a2869da8eff7bcf720c77c8d7d)'s an example.

The difference is bettween getting
  - `thread 'main' panicked at 'called 'Result::unwrap()' on an 'Err' value: TestError(1234)', src/main.rs:12:50` with an unwrap, or
  - `thread 'main' panicked at 'An error occurred: TestError(1234)', src/main.rs:13:50` with an expect.

PR #640, opened 2022-12-02
add getter for trace attribute of VirtualMachine

raefko:
# add getter for trace attribute of VirtualMachine

## Description

Add a function to get the execution trace. We ([FuzzingLabs](https://github.com/FuzzingLabs)) need it for the fuzzer to do the coverage of the cairo program.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/vm_core.rs
+++ b/src/vm/vm_core.rs
@@ -108,6 +108,11 @@ impl VirtualMachine {
         }
     }
 
+    //Simple getter for trace attribute
+    pub fn get_trace(&self) -> Option<&Vec<TraceEntry>> {
+        return self.trace.as_ref();
+    }
+
     ///Returns the encoded instruction (the value at pc) and the immediate value (the value at pc + 1, if it exists in the memory).
     fn get_instruction_encoding(
         &self,


Oppen:
I suggest we modify [this](https://github.com/lambdaclass/cairo-rs/blob/main/src/cairo_run.rs#L66-L83) instead to take an `impl Write`. This is because we try to not expose internal containers, since those are usually the places where optimization happens along the way. The `write_binary_trace` function keeps a stable interface instead, and `Vec` implements the `Write` trait, so you can get an in-memory copy of the trace that way.

MegaRedHand:
As the change proposed by @Oppen was already added in [`cfeac09`](https://github.com/lambdaclass/cairo-rs/commit/cfeac09264bb5d2f598bc6d63157d5cb15befbb6#diff-d5253025fff4d14d6cb4d0aa47f9f3c1a72e6c7fceea740269e4ae2022ffd1f3R87), we're closing this. If that change hasn't fully solved the problem, feel free to open an issue!

PR #641, opened 2022-12-02
add deserialize program from string

raefko:
# add deserialize program from string

## Description

Add a function to deserialize a program from a string. We ([FuzzingLabs](https://github.com/FuzzingLabs)) need it for the fuzzer so it to optimize the I/O of the fuzzer

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/serde/deserialize_program.rs
+++ b/src/serde/deserialize_program.rs
@@ -279,6 +279,70 @@ pub fn deserialize_program_json(reader: impl Read) -> Result<ProgramJson, Progra
     Ok(program_json)
 }
 
+pub fn deserialize_program_json_from_string(json: &String) -> Result<ProgramJson, ProgramError> {
+    let program_json = serde_json::from_str(json)?;
+
+    Ok(program_json)
+}
+
+pub fn deserialize_program_from_string(
+    json: &String,
+    entrypoint: Option<&str>,
+) -> Result<Program, ProgramError> {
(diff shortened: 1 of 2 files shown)


Oppen:
This doesn't seem necessary. Slices implement `Read`, so you can just call `.as_bytes()` on the string and pass the result to the existing function. If you're more comfortable using a helper it may be acceptable, but please make it just a thin wrapper that pass `.as_bytes()`.

Oppen:
@raefko any news about this? Is the solution I proposed good enough or do we need to keep working on this?

raefko:
We benchmarked your solution, and our solution is faster (which is essential for fuzzing)

> This doesn't seem necessary. Slices implement `Read`, so you can just call `.as_bytes()` on the string and pass the result to the existing function. If you're more comfortable using a helper it may be acceptable, but please make it just a thin wrapper that pass `.as_bytes()`.

Oppen:
Have you tried preallocating with a bigger capacity?

Oppen:
Questions:
1. How much faster?
2. How was this benchmarked?
3. Where does the bottleneck lie in each case?

Oppen:
So, I double checked the changes and both functions differ only on the serde call. Let's extract that so we don't copy paste the whole thing.

Oppen:
Currently the VM reads the whole program and deserializes from a slice of bytes. This should also work for this use case, so I'm closing the PR. We can open it again if you have any objections.

PR #642, opened 2022-12-05
Fix CairoRunner::initialize_all_builtins to prioritize program builtins order

fmoletta:
Prerviously, we used the starkness initialization order by default, we should instead use the program builtins order and then complete the non-program builtins with the starkness order to mach the original implementation

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -222,33 +222,56 @@ impl CairoRunner {
     // Initialize all the builtins. Values used are the original one from the CairoFunctionRunner
     // Values extracted from here: https://github.com/starkware-libs/cairo-lang/blob/4fb83010ab77aa7ead0c9df4b0c05e030bc70b87/src/starkware/cairo/common/cairo_function_runner.py#L28
     fn initialize_all_builtins(&self, vm: &mut VirtualMachine) -> Result<(), RunnerError> {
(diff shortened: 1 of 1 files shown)


Oppen:
> we should instead use the program builtins order

Why? The code does look like it does what the PR describes, but it's not clear what makes it incorrect. Does it have something to do with ordering of memory segments for relocation?

fmoletta:
> > we should instead use the program builtins order
> 
> Why? The code does look like it does what the PR describes, but it's not clear what makes it incorrect. Does it have something to do with ordering of memory segments for relocation?

In the [original code for CairoFunctionRunner](https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/cairo/common/cairo_function_runner.py#L31), they first use the CairoRunner's init method, which creates the builtins according to the program builtins, and then they initialize all the builtins on top of that, this replaces the values of the builtins (and adds the missing ones), but it doesn't change their order. As the builtin order is different in the program than in this second builtin initialization phase, when we just used this new builtin order, the bases for some builtins ended up different, the test I added is an example I found when analyzing some tests in the starknet-devnet

Jrigada (approved):

Oppen (approved):

Juan-M-V (approved):

PR #643, opened 2022-12-05
Modify parser to handle more cases

mmsc2:
# Modify parser to handle more cases

* Resume: Modify program deserialization and Ids getters, to handle new reference values in the compiled JSON

While running kakarot we found some parsing errors because of missing cases for instruction formats and casts. 
The different possible formats and casts are as follow:

cast to pointers:  (offsets are i32 integer)
a) offset
b) register + offset
c) [register] + offset
d) [register + offset] + offset
e) [register + offset] + [register + offset] 
f) [register] + [register] 

cast to felts: (immediates are BigInts)
A) Immediate 
B) [register] + immediate
C) [register + offset] + immediate
D) [register + offset] + [register + offset] 
E) [register] + [register] 

Cases d, e, was not contemplated. 

The Pr introduces the Enum OffsetValue that handles all the possible combinations of operations that can be found in instructions. ` Reference` handles the cases when we have operations that involve registers, it stores the register and the offset that is added or subtracted, the bool field of `Reference` is set to `true` when we have a dereference operation like cases c to e and false otherwise (like case b). `Value`  field is used when we are casting to a pointer an the operation doesn't involve registers like case `a`  or when second offset is a number like cases b to d. `Immediate` is used when casting to felt and we are adding or subtracting a number like in cases A to C.
As this Enum encapsulates the behavior of `offset` field, we could simplify the structure of ValueAdress and fields register, immediate and inner_reference were deleted.

Currently, changes must be done in the `hint_processor_utils.rs` file to integrate the new structure into the hint processor.


Juan-M-V (approved):

src/hint_processor/builtin_hint_processor/hint_utils.rs
```diff
@@ -128,8 +122,8 @@ mod tests {
     fn get_ptr_from_var_name_immediate_value() {
         let mut vm = vm!();
         vm.memory = memory![((1, 0), (0, 0))];
-        let mut hint_ref = HintReference::new(0, 0, false, true);
-        hint_ref.immediate = Some(bigint!(2));
+        let mut hint_ref = HintReference::new(0, 0, true, false);
+        hint_ref.offset2 = OffsetValue::Value(2); // Value(offset2) Some(bigint!(2));
```
Juan-M-V, line 126: remove this comment

src/hint_processor/hint_processor_utils.rs
```diff
@@ -71,37 +67,38 @@ pub fn compute_addr_from_reference(
...
-            Ok(dereferenced_addr + bigint_to_usize(imm)?)
-        } else {
-            Ok(dereferenced_addr + hint_reference.offset2)
+        return Err(VirtualMachineError::NoRegisterInReference);
+    };
+
+    match &hint_reference.offset2 {
+        OffsetValue::Reference(register, offset, deref) => {
+            // Cant add two relocatable values
+            // So OffSet2 must be Bigint
```
Oppen, line 88: I'm not sure I understand this.

src/hint_processor/hint_processor_utils.rs
```diff
@@ -133,6 +130,39 @@ pub fn bigint_to_u32(bigint: &BigInt) -> Result<u32, VirtualMachineError> {
     bigint.to_u32().ok_or(VirtualMachineError::BigintToU32Fail)
 }
 
+fn get_offset_value_reference(
+    vm: &VirtualMachine,
+    hint_reference: &HintReference,
+    hint_ap_tracking: &ApTracking,
+    register: &Register,
+    offset: &i32,
+    deref: &bool,
```
Oppen, line 139: Pass these as values.

src/vm/runners/cairo_runner.rs
```diff
@@ -445,17 +446,21 @@ impl CairoRunner {
...
+                    ap_tracking_data: match (
+                        &reference.value_address.offset1,
+                        &reference.value_address.offset2,
+                    ) {
+                        (OffsetValue::Reference(Register::AP, _, _), _) => {
+                            Some(reference.ap_tracking_data.clone())
+                        }
+                        (_, OffsetValue::Reference(Register::AP, _, _)) => {
+                            Some(reference.ap_tracking_data.clone())
+                        }
```
Oppen, line 462:
```suggestion
                        (OffsetValue::Reference(Register::AP, _, _), _)|(_, OffsetValue::Reference(Register::AP, _, _)) => {
                            Some(reference.ap_tracking_data.clone())
                        },
```

Oppen (changes requested):

src/hint_processor/builtin_hint_processor/hint_utils.rs
pefontana: Done

src/vm/runners/cairo_runner.rs
pefontana: Done

src/hint_processor/hint_processor_utils.rs
pefontana: From what I read in the Python VM implementation, it is not possible to add two Relocatable values.
So, If the first term is a relocatable, the second term must be a BigInt

src/hint_processor/hint_processor_utils.rs
pefontana: Done!

Jrigada (approved):

PR #518, opened 2022-10-28
feat: enable the skip_instruction_exectution for hints

tdelabro:
the feature making hints able to disable next instruction executions was disabled

PR #645, opened 2022-12-08
feat: enable the skip_instruction_exectution for hints

Eikix:
# feat: enable the skip_instruction_execution for hints 

the feature making hints able to disable next instruction executions was disabled

## Description

New implementation for https://github.com/lambdaclass/cairo-rs/pull/518
#518 

Adds an integration test for skip_instruction

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added


Eikix:
Hey:), what is the `UnknownHint` error?

Oppen:
> Hey:), what is the UnknownHint error?

The builtin hint processor has a few hardcoded supported hints. Whenever you pass a hint that isn't natively implemented, it will return that error. This allows catching this special case if you want to extend the supported hints with an extra processor while still using the optimized version for the supported hints.

src/vm/vm_core.rs
```diff
@@ -480,21 +480,28 @@ impl VirtualMachine {
...
+                        attr.value.to_string(),
+                        Box::new(err),
+                    ),
+                    _ => err,
+                }
+            })?;
+        } else {
+            let pc = self.get_pc().clone();
+            let size = instruction.size();
+            println!("{:?}", instruction.opcode);
```
Oppen, line 501: We should remove this before merge.

Oppen (changes requested):
LGTM. Just remove the print and it's good to go.

Oppen:
There's one more thing I missed (which is the reason the tests fail): you'll need to implement that setting as a new hint in the builtin runner. You can avoid making the new function public as well when you do that.

Eikix:
> There's one more thing I missed (which is the reason the tests fail): you'll need to implement that setting as a new hint in the builtin runner. You can avoid making the new function public as well when you do that.

Thanks a lot for your reply! I'm on it

Eikix:
I proposed an updated version with what I understood from your review:)! 
Hope this is more what you had in mind. Thanks for your time

src/hint_processor/builtin_hint_processor/skip_next_instruction.rs
```diff
@@ -0,0 +1,28 @@
...
+
+Implements hint:
+%{ skip_next_instruction() %}
+*/
+pub fn skip_next_instruction(
+    vm: &mut VirtualMachine,
+    _exec_scopes: &mut ExecutionScopes,
+    _ids_data: &HashMap<String, HintReference>,
+    _ap_tracking: &ApTracking,
+    _constants: &HashMap<String, BigInt>,
```
Oppen, line 24: You can omit these.

tests/skip_instruction_test.rs
```diff
@@ -1,40 +1,24 @@
...
     let program = Program::from_file(
         Path::new("cairo_programs/test_skip_next_instruction.json"),
         Some("main"),
     )
     .expect("Failed to deserialize program");
 
-    let hint = Rc::new(HintFunc(Box::new(skip_hint)));
+    let hint = Rc::new(HintFunc(Box::new(skip_next_instruction)));
     let mut hint_processor = BuiltinHintProcessor::new_empty();
     hint_processor.add_hint(String::from("skip_next_instruction()"), hint);
```
Oppen, line 23: This is not needed. The `BuiltinHintProcessor` now knows how to handle it by itself.

Oppen (changes requested):
Two very minor changes and it's good to go!

Eikix:
Thanks a lot for your time!

Oppen (approved):
Excellent!

Oppen (approved):

Oppen:
Just for the record. Our common practice is squash and merge, so force pushing isn't strictly necessary. If you can push fixes individually it helps GitHub separate them to see what changed since the last review (I often do partial reviews with that feature to check requested changes before making a final one with everything together).

Juan-M-V (approved):

Juan-M-V:
Hi, thanks for the contribution!
I was looking at the rust/build check and saw that the cairo program `test_skip_next_instruction.cairo` does compile but it fails to run when using `cairo-run` with this error:
```cairo-run --layout all --program cairo_programs/test_skip_next_instruction.json --trace_file cairo_programs/test_skip_next_instruction.trace --memory_file cairo_programs/test_skip_next_instruction.memory
cairo_programs/test_skip_next_instruction.cairo:5:3: Error at pc=0:2:
Got an exception while executing a hint.
  %{ skip_next_instruction() %}
  ^***************************^
Traceback (most recent call last):
  File "cairo_programs/test_skip_next_instruction.cairo", line 5, in <module>
    %{ skip_next_instruction() %}
NameError: name 'skip_next_instruction' is not defined
```

I looked into it a bit and while I haven't found a solution, I did find some things that might be useful for fixing this.
It seems that while the function `skip_next_instruction()` doesn't exist, however, there is a field: `skip_instruction_execution` inside the VM that does what you want to do in the test. I tried doing something similar to what is done here https://github.com/starkware-libs/cairo-lang/blob/d61255f32a7011e9014e1204471103c719cfd5cb/src/starkware/cairo/lang/vm/vm_test.py#L461 but unfortunately it didn't work either, the `vm` object doesn't seem to be in scope, at least not like its seen in the test.

Would you mind taking a look and correcting this?

Eikix:
> Hi, thanks for the contribution! I was looking at the rust/build check and saw that the cairo program `test_skip_next_instruction.cairo` does compile but it fails to run when using `cairo-run` with this error:
> 
> ```
> cairo_programs/test_skip_next_instruction.cairo:5:3: Error at pc=0:2:
> Got an exception while executing a hint.
>   %{ skip_next_instruction() %}
>   ^***************************^
> Traceback (most recent call last):
>   File "cairo_programs/test_skip_next_instruction.cairo", line 5, in <module>
>     %{ skip_next_instruction() %}
> NameError: name 'skip_next_instruction' is not defined
> ```
> 
> I looked into it a bit and while I haven't found a solution, I did find some things that might be useful for fixing this. It seems that while the function `skip_next_instruction()` doesn't exist, however, there is a field: `skip_instruction_execution` inside the VM that does what you want to do in the test. I tried doing something similar to what is done here https://github.com/starkware-libs/cairo-lang/blob/d61255f32a7011e9014e1204471103c719cfd5cb/src/starkware/cairo/lang/vm/vm_test.py#L461 but unfortunately it didn't work either, the `vm` object doesn't seem to be in scope, at least not like its seen in the test.
> 
> Would you mind taking a look and correcting this?

Hey, thanks a lot for your time:)! 

What do you mean to say `It seems that while the function skip_next_instruction() doesn't exist`? 


My understanding of how hints work rn in the VM is:

1. declare a hint code in `hint_code` file:
```pub(crate) const SKIP_NEXT_INSTRUCTION: &str = "skip_next_instruction()";```

this means that the hint processor will understand the text `skip_next_instruction()` inside `%{ %}` hint braces.

2. then, in the `builtin_hint_processor` file, you match the hint_code to the specific implementation of the hint
```
hint_code::SKIP_NEXT_INSTRUCTION => skip_next_instruction(vm),
```

3. the mapped function to the hint code is `skip_next_instruction(vm)`
it is implemented in `skip_next_instruction`
as:
```
pub fn skip_next_instruction(vm: &mut VirtualMachine) -> Result<(), VirtualMachineError> {
    vm.skip_next_instruction_execution();
    Ok(())
}
```

**Comment by Juan-M-V** [comment]
> > Hi, thanks for the contribution! I was looking at the rust/build check and saw that the cairo program `test_skip_next_instruction.cairo` does compile but it fails to run when using `cairo-run` with this error:
> > ```
> > cairo_programs/test_skip_next_instruction.cairo:5:3: Error at pc=0:2:
> > Got an exception while executing a hint.
> >   %{ skip_next_instruction() %}
> >   ^***************************^
> > Traceback (most recent call last):
> >   File "cairo_programs/test_skip_next_instruction.cairo", line 5, in <module>
> >     %{ skip_next_instruction() %}
> > NameError: name 'skip_next_instruction' is not defined
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I looked into it a bit and while I haven't found a solution, I did find some things that might be useful for fixing this. It seems that while the function `skip_next_instruction()` doesn't exist, however, there is a field: `skip_instruction_execution` inside the VM that does what you want to do in the test. I tried doing something similar to what is done here https://github.com/starkware-libs/cairo-lang/blob/d61255f32a7011e9014e1204471103c719cfd5cb/src/starkware/cairo/lang/vm/vm_test.py#L461 but unfortunately it didn't work either, the `vm` object doesn't seem to be in scope, at least not like its seen in the test.
> > Would you mind taking a look and correcting this?
> 
> Hey, thanks a lot for your time:)!
> 
> What do you mean to say `It seems that while the function skip_next_instruction() doesn't exist`?
> 
> My understanding of how hints work rn in the VM is:
> 
>     1. declare a hint code in `hint_code` file:
>        `pub(crate) const SKIP_NEXT_INSTRUCTION: &str = "skip_next_instruction()";`
> 
> 
> this means that the hint processor will understand the text `skip_next_instruction()` inside `%{ %}` hint braces.
> 
>     2. then, in the `builtin_hint_processor` file, you match the hint_code to the specific implementation of the hint
> 
> 
> ```
> hint_code::SKIP_NEXT_INSTRUCTION => skip_next_instruction(vm),
> ```
> 
>     3. the mapped function to the hint code is `skip_next_instruction(vm)`
>        it is implemented in `skip_next_instruction`
>        as:
> 
> 
> ```
> pub fn skip_next_instruction(vm: &mut VirtualMachine) -> Result<(), VirtualMachineError> {
>     vm.skip_next_instruction_execution();
>     Ok(())
> }
> ```
> 
> Meaning that when the runner encounters my hint in `test_skip_next_instruction.cairo` line 1: ` %{ skip_next_instruction() %}` It will go up to point 3.

Oh, sorry, I didn't explain myself properly. I meant that the function doesn't exist in the original python implementation. The part that is failing to run is not cairo_rs, but cairo-lang from strakware

**Comment by Eikix** [comment]
Ohh thanks for the explanation, what are some steps I could take to fix this?

**Comment by Oppen** [comment]
The issue here would be that you need to change the hint to:
```
%{
    vm.skip_instruction_execution = True
%}
```

The problem I see is that you typically will want some flexibility to decide whether to skip it, right?

**Comment by Eikix** [comment]
The issue here is that the hint must work both in Rust and in Python? 

So I have to change the hint text from `skip_next_instruction()` to `vm.skip_instruction_execution = True` to satisfy the Python-version of the VM?

File: src/hint_processor/builtin_hint_processor/skip_next_instruction.rs [hunk 4]
```diff
@@ -0,0 +1,11 @@
+use crate::vm::{errors::vm_errors::VirtualMachineError, vm_core::VirtualMachine};
+
+/*
+This hint doesn't belong to the Cairo common library
+It's only added for testing proposes
+*/
+
+pub fn skip_next_instruction(vm: &mut VirtualMachine) -> Result<(), VirtualMachineError> {
```
Oppen, line 8: This could be `pub(crate)`.

Juan-M-V (approved):

Eikix:
I'm not able to reproduce why tests fail locally.

image

fmoletta:
> I'm not able to reproduce locally why tests fail.
> 
> In the CI: image
> 
> Locally: image
> 
> Both using make test -j

Hello! we merged a PR recently that changed how the hint_processor is used, we now receive a mutable reference to it instead of a normal reference. The CI uses `main` as a base to run the tests, so while your local branch may be working fine, once the changes are applied to the main branch it starts failing. You should pull from the main branch, update your branch, and then you can fix the errors (should be adding a couple `mut`s here and there). Hope this is useful!

Eikix:
> > I'm not able to reproduce locally why tests fail.
> > In the CI: image
> > Locally: image
> > Both using make test -j
> 
> Hello! we merged a PR recently that changed how the hint_processor is used, we now receive a mutable reference to it instead of a normal reference. The CI uses `main` as a base to run the tests, so while your local branch may be working fine, once the changes are applied to the main branch it starts failing. You should pull from the main branch, update your branch, and then you can fix the errors (should be adding a couple `mut`s here and there). Hope this is useful!

Great ! Thanks:)! Done

src/hint_processor/builtin_hint_processor/skip_next_instruction.rs
Eikix: i'm thinking this could be exposed for maximum composability with other test-oriented hints

Eikix:
Hi:)! I'll resolve the conflicts,
Also, I posted a message in the starknet-rs telegram chat that I reproduce hereafter: 

```
hi all:)! 
i think this PR (approved) is closed to the finish line https://github.com/lambdaclass/cairo-rs/pull/645

TL;DR: some features (i.e. test runner features) i’d like to add into cairo-rs are intrinsically not retrocompatible with the python vm so this PR finds a way to devise a workaround way to add these features 

there is one thing i’d like to check with you, it is about VM features that do not have retrocompatibility with the Python VM.
An example of that is the skip_next_instruction() hint, and other hints aimed at making Cairo an easily testable language (with lifecycle hooks available and so on to be able to build a test runner).

I’d like to check the following flow with you:
- whenever someone PRs for a functionality that has no retrocompat with the python VM, they put it in a feature (in the case of language test utils, it’ll be in the feature test_utils.
Then, the repository test suites tests has the test_utils feature enabled but the cross VM benchmark has it disabled.

- Then, for .cairo files that test the test_utils features, their name is of format <my_test_filename>.noretrocompat.cairo so that the .sh script does not include them in cross VM tests.
```

Eikix:
What is the status on this PR:)?

Oppen:
Hi @Eikix, sorry for the delays. I think the procedure in your last comment is sensible and the way to go.
I'll do a last round of review but I think it should be mergeable now.

Eikix:
> Hmmm, it doesn't build. It must have been broken by the toolchain updates. There's another issue we need to handle internally where the benchmark failed to build but the pipeline didn't (it did technically fail but later than it should have). I'll raise this one in an issue.

Thanks! Tried to rebase and fix issues due to rebase

Makefile
```diff
@@ -102,7 +105,7 @@ cairo_trace: $(CAIRO_TRACE) $(CAIRO_MEM)
 cairo-rs_trace: $(CAIRO_RS_TRACE) $(CAIRO_RS_MEM)
 
 test: $(COMPILED_PROOF_TESTS) $(COMPILED_TESTS) $(COMPILED_BAD_TESTS)
-	cargo test --workspace
+	cargo test  --workspace --features test_utils
```
Oppen, line 108:
```suggestion
	cargo test --workspace --features test_utils
```

Oppen (approved):

PR #653, opened 2022-12-21
Fix bug in  ASSERT_EQ case in opcode_assertions 

fmoletta:
In the ASSERT_EQ case of opcode_assertions, values are only compared if they are int values, which lead to assertions between relocatables to be skipped.
This PR fixes the method so that relocatables are also compared and adds a test for this case


Oppen (approved):
Beautiful catch!

Jrigada (approved):

PR #654, opened 2022-12-22
Change execute_hint to receive mutable self reference

Jrigada:
# Change execute_hint to receive a mutable self reference

## Description
This change is in line with what the sequencer needs, to be able to make changes to the hint_processor while executing hints.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ let mut vm = VirtualMachine::new(
 
 let mut cairo_runner = CairoRunner::new(&$program, "all", false);
 
-let hint_processor = BuiltinHintProcessor::new_empty();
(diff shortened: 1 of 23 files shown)


Oppen (approved):
The code itself is perfect as it is and solves a real, immediate and urgent problem.
However, I'd like you to file a tech debt issue because mutable hint processors may lead to very hard to debug errors during runtime, since it allows a behavior analogous to self-modifying code. I'd like to discuss (later) if we can find a better solution than making the processor itself mutable.

Juan-M-V (approved):

PR #657, opened 2022-12-22
Add traceback to VmException

fmoletta:
- Add the `traceback` field to `VmException`
- Add logic necessary to construct the traceback
- Add tests for new/changed methods
- Add changelog
The following will be added in a later PR:
- Improving the function `get_error_attr_value` by adding the method `substitute_error_message_references`.

This PR will not add/process the following information:
- hints field of debug_info.inst_location
- input_file contents
- notes
This may be added in the future in terms of priority

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,7 @@
-2022-12-28
-* Add traceback to VmException
-    * PR: [#657](https://github.com/lambdaclass/cairo-rs/pull/657)
-    * Main functionality changes: `VmException` now contains a traceback in the form of a String which lists the locations (consisting of filename, line, column and pc) of the calls leading to the error.
+* Add traceback to VmException [#657](https://github.com/lambdaclass/cairo-rs/pull/657)
     * Public API changes: 
         * `traceback` field added to `VmException` struct
-        * `VmException::from_vm_error()` signature changed from `pub fn from_vm_error(runner: &CairoRunner, error: VirtualMachineError, pc: usize) -> Self` to `pub fn from_vm_error(runner: &CairoRunner, vm: &VirtualMachine, error: VirtualMachineError) -> Self`
-        * `get_location()` signature changed from `pub fn get_location(pc: &usize, runner: &CairoRunner) -> Option<Location>` to `pub fn get_location(pc: usize, runner: &CairoRunner) -> Option<Location>`
-        * `decode_instruction()` signature changed from `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<BigInt>) -> Result<instruction::Instruction, VirtualMachineError>` to `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<&BigInt>) -> Result<instruction::Instruction, VirtualMachineError>`
-        * Minor changes in `VmExcepion` field's string format to mirror their cairo-lang conterparts.
+        * `pub fn from_vm_error(runner: &CairoRunner, error: VirtualMachineError, pc: usize) -> Self` is now `pub fn from_vm_error(runner: &CairoRunner, vm: &VirtualMachine, error: VirtualMachineError) -> Self`
+        * `pub fn get_location(pc: &usize, runner: &CairoRunner) -> Option<Location>` is now `pub fn get_location(pc: usize, runner: &CairoRunner) -> Option<Location>`
+        * `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<BigInt>) -> Result<instruction::Instruction, VirtualMachineError>` is now `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<&BigInt>) -> Result<instruction::Instruction, VirtualMachineError>`
+        * `VmExcepion` field's string format now mirror their cairo-lang conterparts.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 2022-12-28
 * Add traceback to VmException
-    * PR: #657
+    * PR: [#657](https://github.com/lambdaclass/cairo-rs/pull/657)
     * Main functionality changes: `VmException` now contains a traceback in the form of a String which lists the locations (consisting of filename, line, column and pc) of the calls leading to the error.
     * Public API changes: 
         * `traceback` field added to `VmException` struct
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,10 @@
+2022-12-28
+* Add traceback to VmException
+    * PR: #657
+    * Main functionality changes: `VmException` now contains a traceback in the form of a String which lists the locations (consisting of filename, line, column and pc) of the calls leading to the error.
+    * Public API changes: 
+        * `traceback` field added to `VmException` struct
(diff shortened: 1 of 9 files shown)


src/types/instruction.rs
```diff
@@ -78,3 +81,38 @@ impl Instruction {
         }
     }
 }
+
+// Returns True if the given instruction looks like a call instruction.
+pub(crate) fn is_call_instruction(encoded_instruction: &BigInt, imm: Option<BigInt>) -> bool {
+    let encoded_i64_instruction: i64 = match encoded_instruction.to_i64() {
+        Some(num) => num,
+        None => return false,
+    };
```
Oppen, line 90: I think the following should work (tho not an important change, it's neat information):
```suggestion
    let encoded_i64_instruction: i64 = encoded_instruction.to_i64()
        .ok_or(false);
```

Oppen: That's because `Result<T, U>` can have any type `U` as its other variant.

src/types/instruction.rs
```diff
@@ -78,3 +81,38 @@ impl Instruction {
         }
     }
 }
+
+// Returns True if the given instruction looks like a call instruction.
+pub(crate) fn is_call_instruction(encoded_instruction: &BigInt, imm: Option<BigInt>) -> bool {
```
Oppen, line 86: Why is `encoded_instruction` a `BigInt`?

src/vm/errors/vm_exception.rs
```diff
@@ -11,42 +14,69 @@ pub struct VmException {
...
+        runner: &CairoRunner,
+        vm: &VirtualMachine,
+        error: VirtualMachineError,
+    ) -> Self {
+        let error_attr_value = get_error_attr_value(vm.run_context.pc.offset, runner);
         VmException {
-            pc,
-            inst_location: get_location(&pc, runner),
+            pc: vm.run_context.pc.offset,
+            inst_location: get_location(vm.run_context.pc.offset, runner),
```
Oppen, line 29:
```suggestion
        let pc = vm.run_context.pc.offset;
        let error_attr_value = get_error_attr_value(pc, runner);
        VmException {
            pc,
            inst_location: get_location(pc, runner),
```

src/vm/errors/vm_exception.rs
```diff
@@ -11,42 +14,69 @@ pub struct VmException {
...
 }
 
-pub fn get_location(pc: &usize, runner: &CairoRunner) -> Option<Location> {
+pub fn get_location(pc: usize, runner: &CairoRunner) -> Option<Location> {
     runner
         .program
         .instruction_locations
         .as_ref()?
-        .get(pc)
+        .get(&pc)
```
Oppen, line 52: Why does `get` take a reference here?

src/types/instruction.rs
Oppen: Maybe use an `Option<&BigInt>` for the immediate?

Oppen (approved):
Few minor comments but should be good to ship.

src/vm/errors/vm_exception.rs
fmoletta: Its due to the `HashMap::get()` signature

src/vm/errors/vm_exception.rs
fmoletta: Done!

src/types/instruction.rs
fmoletta: `encoded_instruction` was left as a BigInt in order to reduce the complexity of `get_traceback_entries`. `is_call_instruction` is only used inside `get_traceback_entries`., which is on the longer side, so it was a bit more convenient to have the type conversion inside the simpler method `is_call_instruction`

fmoletta: Added reference to `immediate `argument in `is_call_instruction` and `decode_instruction`

src/types/instruction.rs
fmoletta: Couldn't get it to work

Jrigada (approved):

src/types/instruction.rs
Oppen: Because I was wrong and didn't notice the types mismatch :facepalm:

CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+2022-12-28
+* Add traceback to VmException
+    * PR: [#657](https://github.com/lambdaclass/cairo-rs/pull/657)
+    * Main functionality changes: `VmException` now contains a traceback in the form of a String which lists the locations (consisting of filename, line, column and pc) of the calls leading to the error.
```
Oppen, line 4: This can be more concise:
```suggestion
* Add traceback to VmException [#657](https://github.com/lambdaclass/cairo-rs/pull/657)
```

CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
...
+    * Public API changes: 
+        * `traceback` field added to `VmException` struct
+        * `VmException::from_vm_error()` signature changed from `pub fn from_vm_error(runner: &CairoRunner, error: VirtualMachineError, pc: usize) -> Self` to `pub fn from_vm_error(runner: &CairoRunner, vm: &VirtualMachine, error: VirtualMachineError) -> Self`
+        * `get_location()` signature changed from `pub fn get_location(pc: &usize, runner: &CairoRunner) -> Option<Location>` to `pub fn get_location(pc: usize, runner: &CairoRunner) -> Option<Location>`
+        * `decode_instruction()` signature changed from `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<BigInt>) -> Result<instruction::Instruction, VirtualMachineError>` to `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<&BigInt>) -> Result<instruction::Instruction, VirtualMachineError>`
+        * Minor changes in `VmExcepion` field's string format to mirror their cairo-lang conterparts.
```
Oppen, line 10: These may only mention the before and after, like:
```suggestion
        * `traceback` field added to `VmException` struct
        * `pub fn from_vm_error(runner: &CairoRunner, error: VirtualMachineError, pc: usize) -> Self` is now `pub fn from_vm_error(runner: &CairoRunner, vm: &VirtualMachine, error: VirtualMachineError) -> Self`
        * `pub fn get_location(pc: &usize, runner: &CairoRunner) -> Option<Location>` is now `pub fn get_location(pc: usize, runner: &CairoRunner) -> Option<Location>`
        * `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<BigInt>) -> Result<instruction::Instruction, VirtualMachineError>` is now `pub fn decode_instruction(encoded_instr: i64, mut imm: Option<&BigInt>) -> Result<instruction::Instruction, VirtualMachineError>`
        * `VmExcepion` field's string format now mirror their cairo-lang conterparts.
```

In any case, I think it's better to signal a given release breaks the API instead and keep a migration guide for the details. We also don't need dates IMO. Users are expected to use releases and the changelogs are only for them, so that's how we should group the changes. Besides, the information given by pointing to the PR is a superset of the one given by the date, making it redundant.

CHANGELOG.md
fmoletta: Sounds good!

CHANGELOG.md
fmoletta: Done!

PR #682, opened 2023-01-03
SECP optimizations

Oppen:
- Instead of calling `div_floor` followed by `mod_floor` with the same arguments in a loop in `math_utils::igcdex`, call `div_mod_floor` once;
- Make substractions in the `igcdex` loop in-place for `x` and `y` to avoid creating temporary objects;
- Make the `assert_eq` that checks the input to `igcdex` and the prime are actually coprime run only on debug (testing) builds;
- Make the `assert` calls that validate inputs to `ec_double` and `line_slope` run only on debug builds;
- Square by product rather than `modpow`.

These changes make the SECP integration benchmark 12% faster when running 50 iterations, with 10 warmup runs and 100 effective runs on the benchmark server. (TODO: recheck cumulative change since latest commits)
Other benchmark programs seem completely unaffected.


fmoletta (approved):
Nice Work!

Jrigada (approved):

unbalancedparentheses:
Great work

PR #684, opened 2023-01-03
Bitwise optimizations

Oppen:
- Count bits rather than creating a shifted bound;
- Simplify index deduction;
- Capture by reference.

Local benchmark shows 19% speed improvement in keccak program (confusingly).


fmoletta (approved):
Great Work!

Jrigada (approved):

PR #685, opened 2023-01-03
Optimize bit-width bounds checks

Oppen:
Converts some constants created by shifts into either of:
- Construction from `u128::MAX`;
- Comparison based on the `.bits()` method instead.

This shaves 2-7% according to local benchmarks.
Update: some get to 15-18% of speed improvement.


fmoletta (approved):

Oppen:
After resolving the merge conflicts it stopped presenting any significant gains, so I'm closing it.

PR #687, opened 2023-01-04
Change the Dictionary used in dict hints to store `MaybeRelocatable` instead of `BigInt`

fmoletta:
Solves #679


Jrigada (approved):
LGTM!

PR #690, opened 2023-01-05
Handle extra case in is_zero function

Jrigada:
# Handle extra case in is_zero function

## Description

Handle extra case when asking if a relocatable is zero so that it can return false according to the python implementation. 

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/vm/vm_core.rs
+++ b/src/vm/vm_core.rs
@@ -248,7 +248,13 @@ impl VirtualMachine {
     fn is_zero(addr: &MaybeRelocatable) -> Result<bool, VirtualMachineError> {
         match addr {
             MaybeRelocatable::Int(num) => Ok(num.is_zero()),
-            MaybeRelocatable::RelocatableValue(_rel_value) => Err(VirtualMachineError::PureValue),
+            MaybeRelocatable::RelocatableValue(rel_value) => {
+                if rel_value.offset > 0 {
+                    return Ok(false);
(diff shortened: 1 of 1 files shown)


src/vm/vm_core.rs
```diff
@@ -248,7 +248,13 @@ impl VirtualMachine {
...
         match addr {
             MaybeRelocatable::Int(num) => Ok(num.is_zero()),
-            MaybeRelocatable::RelocatableValue(_rel_value) => Err(VirtualMachineError::PureValue),
+            MaybeRelocatable::RelocatableValue(rel_value) => {
+                if rel_value.offset > 0 {
+                    return Ok(false);
+                } else {
+                    Err(VirtualMachineError::PureValue)
+                }
+            }
```
Oppen, line 257: Nit (feel free to ignore):
```suggestion
            MaybeRelocatable::RelocatableValue(rel_value) if rel_value.offset > 0 => Ok(false),
            _ => Err(VirtualMachineError::PureValue),
            }
```

Oppen: In any case, at a glance it seems like the expected logic would be:
```suggestion
            MaybeRelocatable::RelocatableValue(rel_value) => Ok(rel_value.offset.is_zero()),
```

fmoletta (approved):

PR #698, opened 2023-01-09
Use Euclid to compute the inverse faster

Oppen:
Also update changelog

### DIFF
--- a/felt/src/bigint_felt.rs
+++ b/felt/src/bigint_felt.rs
@@ -427,7 +427,7 @@ impl Div for FeltBigInt {
     fn div(self, rhs: Self) -> Self::Output {
         let x = rhs
             .0
-            .to_bigint()
+            .to_bigint() // Always succeeds for BitUint -> BigInt
             .unwrap()
             .extended_gcd(&CAIRO_SIGNED_PRIME)
             .x;
@@ -442,7 +442,7 @@ impl<'a> Div for &'a FeltBigInt {
     fn div(self, rhs: Self) -> Self::Output {
         let x = rhs
             .0
-            .to_bigint()
+            .to_bigint() // Always succeeds for BitUint -> BigInt
             .unwrap()
             .extended_gcd(&CAIRO_SIGNED_PRIME)
             .x;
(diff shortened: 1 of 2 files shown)


felt/src/bigint_felt.rs
```diff
@@ -422,28 +422,46 @@ impl<'a> Pow<u32> for &'a FeltBigInt {
...
+    // In Felts `x / y` needs to be expressed as `x * y^-1`
+    #[allow(clippy::suspicious_arithmetic_impl)]
     fn div(self, rhs: Self) -> Self::Output {
-        let mut x = rhs.0.modpow(&(&*CAIRO_PRIME - 2usize), &CAIRO_PRIME);
-        x *= self.0;
-        FeltBigInt::from(x)
+        let x = rhs
+            .0
+            .to_bigint()
+            .unwrap()
```
juanbono, line 431: It would be nice to have a comment here about why it's ok to use unwrap() here

felt/src/bigint_felt.rs
```diff
@@ -422,28 +422,46 @@ impl<'a> Pow<u32> for &'a FeltBigInt {
...
+    // In Felts `x / y` needs to be expressed as `x * y^-1`
+    #[allow(clippy::suspicious_arithmetic_impl)]
     fn div(self, rhs: Self) -> Self::Output {
-        let mut x = rhs.0.modpow(&(&*CAIRO_PRIME - 2usize), &CAIRO_PRIME);
-        x *= &self.0;
-        FeltBigInt::from(x)
+        let x = rhs
+            .0
+            .to_bigint()
+            .unwrap()
```
juanbono, line 446: Same as the previous comment

juanbono (approved):
LGTM. Also I left some comments

azteca1998 (approved):
I was going to suggest using `CAIRO_PRIME` to avoid both conversions and `.unwrap()`, but the algorithm doesn't seem to work properly without negative numbers available.

Oppen:
> I was going to suggest using `CAIRO_PRIME` to avoid both conversions and `.unwrap()`, but the algorithm doesn't seem to work properly without negative numbers available.

There's a bug in `BigUint` for the extended GCD algorithm where they fail to consider at least one of the Bézout coefficients for two coprimes will be negative, so there's a substraction that ends up in a `panic` due to having to return unsigned.

PR #750, opened 2023-01-19
Abstract big integers into a new type

fmoletta:
Creates new type `BigNum` used to abstract big signed integers
Replace all instances of `BigInt` with `BigNum`
WARNING: Still needs to be tested for performance


big-num/src/bigint_num.rs
```diff
@@ -0,0 +1,694 @@
...
+        value.into()
+    }
+}
+
+impl Add for BigIntNum {
+    type Output = Self;
+    fn add(mut self, rhs: Self) -> Self {
+        self.0 += rhs.0;
+        self
+    }
```
Oppen, line 133: `add` is not supposed to modify `self`.

big-num/src/bigint_num.rs
```diff
@@ -0,0 +1,694 @@
...
+        iter.fold(BigIntNum::zero(), |mut acc, x| {
+            acc += x;
+            acc
+        })
+    }
+}
+
+impl Neg for BigIntNum {
+    type Output = BigIntNum;
+    fn neg(self) -> Self::Output {
```
Oppen, line 199: `neg` should always negate its input.

big-num/src/bigint_num.rs
```diff
@@ -0,0 +1,694 @@
...
+            BigIntNum(-self.0.clone())
+        } else {
+            self.clone()
+        }
+    }
+}
+
+impl Sub for BigIntNum {
+    type Output = Self;
+    fn sub(mut self, rhs: Self) -> Self::Output {
```
Oppen, line 221: `sub` shouldn't mutate.

big-num/src/bigint_num.rs
```diff
@@ -0,0 +1,694 @@
...
+    fn div_floor(&self, other: &Self) -> Self {
+        BigIntNum(self.0.div_floor(&other.0))
+    }
+
+    fn div_rem(&self, other: &Self) -> (Self, Self) {
+        div_rem(self, other)
+    }
+
+    fn divides(&self, other: &Self) -> bool {
+        self.0.divides(&other.0)
```
Oppen, line 413: `divides` is deprecated. Use `other.0.is_multiple_of(&self.0)` instead.

Oppen (changes requested):

Oppen (approved):

PR #763, opened 2023-01-22
Fix offsets in builtin security check.

alonh5:
# Fix offsets in builtin security check

## Description

Count felt values in the offsets.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


Oppen (approved):
Looks correct. I compared with the Python code and it seems we mixed up our count due to still filtering by offset despite not being iterating the whole memory, so instead we filtered not offsets that contained a value but offsets where the value was an address too.

Oppen:
Just to clarify, @fmoletta said she wanted to make a second round just in case, so I'm not gonna merge it directly to give her the time to do so :)

PR #764, opened 2023-01-23
Refactor method `BuiltinRunner::run_security_checks`

fmoletta:
- Refactor method `BuiltinRunner::run_security_checks` to be clearer and reduce inconsistencies with source code
- Add convenience methods `BuiltinRunner::input_cells`, `BuiltinRunner::cells_per_instance`,  `BuiltinRunner::name` & `VirtualMachine::verify_auto_deductions_for_addr`
- Add tests for new methods and adapt existing tests for the modified method

Notes:
After the refactor of `BuiltinRunner::run_security_checks`, some tests had to be changed as the functionality changed. In order to check that each change was valid, I compared the test to a modified version of the [source implementation](https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/cairo/lang/vm/builtin_runner.py#L257) using the test's hardcoded values.
An assumption was made in order to simplify the code: builtin's memory segments don't contain trailing None values.
The reason for this assumption is that trailing none values don't occur naturally in the Vm .None values are a product of  inserting elements(in which case None values are always before an existing element), and there is no public interface that would allow inserting a None value separately.
If the memory was  tampered with, and there were trailing None values, the security checks would fail.
This is an internal refactor without changes to the public interface

### DIFF
--- a/src/vm/runners/builtin_runner/mod.rs
+++ b/src/vm/runners/builtin_runner/mod.rs
@@ -1090,6 +1090,64 @@ mod tests {
     }
 
     #[test]
+    fn run_security_ec_op_check_memory_empty() {
+        let mut ec_op_builtin = EcOpBuiltinRunner::new(&EcOpInstanceDef::default(), true);
+
+        ec_op_builtin.cells_per_instance = 5;
+        ec_op_builtin.n_input_cells = 7;
+
+        let builtin: BuiltinRunner = ec_op_builtin.into();
+
+        let mut vm = vm!();
+        // The values stored in memory are not relevant for this test
+        vm.memory.data = vec![vec![]];
+
+        assert_eq!(builtin.run_security_checks(&mut vm), Ok(()),);
+    }
+
+    #[test]
+    fn run_security_ec_op_check_memory_1_element() {
+        let mut ec_op_builtin = EcOpBuiltinRunner::new(&EcOpInstanceDef::default(), true);
+
+        ec_op_builtin.cells_per_instance = 5;
+        ec_op_builtin.n_input_cells = 7;
+
+        let builtin: BuiltinRunner = ec_op_builtin.into();
+
+        let mut vm = vm!();
+        // The values stored in memory are not relevant for this test
+        vm.memory.data = vec![vec![mayberelocatable!(0).into()]];
+
+        assert_eq!(
+            builtin.run_security_checks(&mut vm),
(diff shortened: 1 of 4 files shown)


src/vm/runners/builtin_runner/mod.rs
```diff
@@ -257,108 +258,102 @@ impl BuiltinRunner {
...
+        let cells_per_instance = self.cells_per_instance() as usize;
+        let n_input_cells = self.n_input_cells() as usize;
+        let builtin_segment_index = self
+            .base()
+            .to_usize()
+            .ok_or(VirtualMachineError::NegBuiltinBase)?;
+        // If the builtin's segment is empty, there are no security checks to run
+        let builtin_segment = match vm.memory.data.get(builtin_segment_index) {
+            Some(segment) => segment,
+            None => return Ok(()),
```
Oppen, line 310: Is `segment.is_empty()` a possibility? If it is then it should also `Return Ok(())`, right?
```suggestion
        // If the builtin's segment is empty, there are no security checks to run
        let builtin_segment = match vm.memory.data.get(builtin_segment_index) {
            Some(segment) if !segment.is_empty() => segment,
            _ => return Ok(()),
```

src/vm/vm_core.rs
```diff
@@ -666,8 +666,8 @@ impl VirtualMachine {
     }
 
     ///Makes sure that all assigned memory cells are consistent with their auto deduction rules.
-    pub fn verify_auto_deductions(&mut self) -> Result<(), VirtualMachineError> {
-        for (name, builtin) in self.builtin_runners.iter_mut() {
+    pub fn verify_auto_deductions(&self) -> Result<(), VirtualMachineError> {
```
Oppen, line 669: I'm confused here. Isn't the `deduce_memory_cell` supposed to mutate `self.memory`?

Oppen (approved):
Beautiful code!

src/vm/runners/builtin_runner/mod.rs
fmoletta: Good call

src/vm/vm_core.rs
fmoletta: The values obtained from `deduce_memory_cell` are inserted into memory in `compute_operands`, not here

Oppen (approved):

src/vm/vm_core.rs
Oppen: That makes sense, thanks!

Jrigada (approved):

PR #768, opened 2023-01-23
Refactor `verify_secure_runner`

fmoletta:
- Refactor `verify_secure_runner`
- Simplify `CairoRunner::get_segment_info` & `BuiltinRunner::get_memory_addresses`
- Apply a quickfix for builtin's stop_ptr not being updated when calling `BuiltinRunner::final_stack` inside `CairoRunner::read_return_values` (Note: significant refactor needed to fix in a clean way, may affect cairo_rs_py)
- Add tests for affected methods

TODO in a next PR: integrate `verify_secure_runner` into `cairo_run`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Refactor `Refactor verfify_secure_runner` [#768](https://github.com/lambdaclass/cairo-rs/pull/768)
     Public Api changes:
     * Remove builtin name from the return value of `BuiltinRunner::get_memory_segment_addresses`
-    * Simplify the return value of `CairoRunner::get_builtin_segments_info` to `HashMap<isize, usize>`
+    * Simplify the return value of `CairoRunner::get_builtin_segments_info` to `Vec<(usize, usize)>`
     * CairoRunner::read_return_values now receives a mutable reference to VirtualMachine
     Bugfixes:
     * CairoRunner::read_return_values now updates the `stop_ptr` of each builtin after calling `BuiltinRunner::final_stack`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 #### Upcoming Changes
 
+* Use CairoArg enum instead of Any in CairoRunner::run_from_entrypoint [#686](https://github.com/lambdaclass/cairo-rs/pull/686)
+    * Public Api changes:
+        * Remove `Result` from `MaybeRelocatable::mod_floor`, it now returns a `MaybeRelocatable` 
+        * Add struct `CairoArg`
+        * Change `arg` argument of `CairoRunner::run_from_entrypoint` from `Vec<&dyn Any>` to `&[&CairoArg]`
+        * Remove argument `typed_args` from `CairoRunner::run_from_entrypoint`
+        * Remove no longer used method `gen_typed_arg` from `VirtualMachine` & `MemorySegmentManager`
+        * Add methods `MemorySegmentManager::gen_cairo_arg` & `MemorySegmentManager::write_simple_args` as typed counterparts to `MemorySegmentManager::gen_arg` & `MemorySegmentManager::write_arg`
+        
 #### [0.1.1] - 2023-01-11
 
 * Add input file contents to traceback [#666](https://github.com/lambdaclass/cairo-rs/pull/666/files)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 #### Upcoming Changes
 
+* Refactor `Refactor verfify_secure_runner` [#768](https://github.com/lambdaclass/cairo-rs/pull/768)
+    Public Api changes:
+    * Remove builtin name from the return value of `BuiltinRunner::get_memory_segment_addresses`
+    * Simplify the return value of `CairoRunner::get_builtin_segments_info` to `HashMap<isize, usize>`
+    * CairoRunner::read_return_values now receives a mutable reference to VirtualMachine
+    Bugfixes:
+    * CairoRunner::read_return_values now updates the `stop_ptr` of each builtin after calling `BuiltinRunner::final_stack`
+
 * Use CairoArg enum instead of Any in CairoRunner::run_from_entrypoint [#686](https://github.com/lambdaclass/cairo-rs/pull/686)
     * Public Api changes:
         * Remove `Result` from `MaybeRelocatable::mod_floor`, it now returns a `MaybeRelocatable` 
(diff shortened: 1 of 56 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 #### Upcoming Changes
 
+* Refactor `Refactor verfify_secure_runner` [#768](https://github.com/lambdaclass/cairo-rs/pull/768)
```
juanbono, line 5: typo `verify`

fmoletta: Oh no!

fmoletta: Fixed

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 #### Upcoming Changes
 
+* Refactor `Refactor verify_secure_runner` [#768](https://github.com/lambdaclass/cairo-rs/pull/768)
+    Public Api changes:
+    * Remove builtin name from the return value of `BuiltinRunner::get_memory_segment_addresses`
+    * Simplify the return value of `CairoRunner::get_builtin_segments_info` to `HashMap<isize, usize>`
+    * CairoRunner::read_return_values now receives a mutable reference to VirtualMachine
+    Bugfixes:
+    * CairoRunner::read_return_values now updates the `stop_ptr` of each builtin after calling `BuiltinRunner::final_stack`
```
Oppen, line 11: This could be a bit confusing. When I call a method called `read_something` I don't expect it to change stage.

src/vm/security.rs
```diff
@@ -1,99 +1,88 @@
...
     };
-
-    let builtin_segment_by_index = builtin_segments
-        .iter()
-        .map(|(seg_name, seg_info)| (seg_info.index, (seg_name, seg_info)))
-        .collect::<HashMap<_, _>>();
-
-    let memory_iter = vm
+    // Check builtin segment out of bounds.
+    for (index, stop_ptr) in builtins_segment_info {
```
Oppen, line 32: I'm confused about this segment info. Is the index allowed to be negative for builtins? Otherwise, why don't we just use `usize` and try to catch the error earlier?

src/vm/security.rs
```diff
@@ -1,99 +1,88 @@
...
     vm: &mut VirtualMachine,
 ) -> Result<(), VirtualMachineError> {
-    let program_base = runner
-        .program_base
-        .as_ref()
-        .ok_or(RunnerError::NoProgBase)?;
-
-    let builtin_segments = match verify_builtins {
+    let builtins_segment_info = match verify_builtins {
         true => runner.get_builtin_segments_info(vm)?,
```
Oppen, line 28: Since we're breaking this interface, I think this could probably return a `Vec<(index, stop_ptr)>`. Is there any scenario where we want random lookup?

src/vm/security.rs
```diff
@@ -1,99 +1,88 @@
...
-                return Err(RunnerError::FailedMemoryGet(MemoryError::NumOutOfBounds).into());
+        .get(program_segment_index)
+        .map(|segment| segment.len());
+    // + 1 here accounts for maximum segment offset being segment.len() -1
+    if program_segment_size >= Some(runner.program.data.len() + 1) {
+        return Err(VirtualMachineError::OutOfBoundsProgramSegmentAccess);
+    }
+    // Check that the addresses in memory are valid
+    // This means that every temporary address has been properly relocated to a real address
+    // Asumption: If temporary memory is empty, this means no temporary memory addresses were generated and all addresses in memory are real
```
Oppen, line 58: What if the address was stored but not accessed? Then (I think) it wouldn't create any values in `temp_data`, nor would it fail, but it would still store an address to a temporary segment. Is this reasoning right?

Jrigada (approved):
Just some small typos! I made the suggestions

CHANGELOG.md
fmoletta: The reason for this is the following:
`BuiltinRunner::final_stack`, in the source code updates the value of the builtin's stop_ptr. When this method was implemented, they needed the method to receive an inmutable reference (i dont know the reason why). So they instead made the method return the stop_ptr as a second return value and added the method `BuiltinRunner::stop_ptr` to update the stop_ptr after calling `BuiltinRunner::final_stack`. But this didnt happen, leading to the stop_ptr never being written and therefore verify_secure_runner always returned with error due to stop_ptr being None.
The correct way to fix this would be to correct `BuiltinRunner::final_stack`, but this requires changes in cairo_rs_py too, thats why I did a quick fix for the moment.

src/vm/security.rs
fmoletta: Good call

src/vm/security.rs
fmoletta: Good call

src/vm/security.rs
fmoletta: `RelocatableValue` and `MaybeRelocatable` are not exposed during hints, neither in the original nor in our cairo_rs_py version, so there is no way to "create" a relocatable value. The only way to obtain a relocatable value with negative index is to create a temporary segment, and when this happens, a new segment will be created in the temp_data, leading to it being not empty.
I made this assumption as having temporary data is not something that happens when running normal cairo programs, and this function is ran after every standard (aka no special flags) run, so being able to forgo iterating the whole memory will present an advantage performace-wise.

src/vm/errors/vm_errors.rs
```diff
@@ -135,6 +135,12 @@ pub enum VirtualMachineError {
     Hint(usize, Box<HintError>),
     #[error("Unexpected Failure")]
     Unexpected,
+    #[error("Out of bounds access to builtin segment")]
+    OutOfBoundsBuiltinSegmentAccess,
+    #[error("Out of bounds access to program segment")]
+    OutOfBoundsProgramSegmentAccess,
     #[error("Negative builtin base")]
     NegBuiltinBase,
+    #[error("Security Error: Invalid Memory Value: tenporary address not relocated: {0}")]
```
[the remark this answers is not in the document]
fmoletta: Good call

src/vm/runners/builtin_runner/mod.rs
```diff
@@ -1020,7 +1010,7 @@ mod tests {
             true,
         ));
         let vm = vm!();
-        // Unsed builtin shouldnt fail security checks
+        // Unused builtin shouldnt fail security checks
```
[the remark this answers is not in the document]
fmoletta: Good catch

CHANGELOG.md
Oppen: Can we point the quick fix with a `FIXME` comment?

fmoletta: sure

Oppen (approved):

PR #769, opened 2023-01-24
WIP - installation scripts

martinacantaro:
# Installation scripts

## Description

Goal: to make it easier to build the repo.
These scripts will install all dependencies for:
- Debian
- Ubuntu
- Mac OS

## Checklist
- [x] Linked to Github Issue https://github.com/lambdaclass/cairo-rs/issues/736
- [-] Unit tests added
- [-] Integration tests added.
- [-] This change requires new documentation.
  - [-] Documentation has been added/updated.

### DIFF
--- a/install-scripts/install-debian.sh
+++ b/install-scripts/install-debian.sh
@@ -1,5 +1,3 @@
-# This installation has been tested on Debian 11 XFCE
-
 # Install curl in order to install Rust and Cargo
 # Install make, necessary for installing python 3.9 with pyenv
 
--- a/install-scripts/install-debian.sh
+++ b/install-scripts/install-debian.sh
@@ -1,0 +1,53 @@
+# This installation has been tested on Debian 11 XFCE
+
+# Install curl in order to install Rust and Cargo
+sudo apt install curl
+
+# Install Rust and Cargo
+curl https://sh.rustup.rs -sSf | sh
+
+# TODO Restart console? (It's necessary in order to ensure PATH is configured correctly by rustup)
+
+# Make sure Rust has been installed correctly
+rustc --version
+
+# Install pyenv dependencies
+sudo apt-get install -y make build-essential libssl-dev zlib1g-dev libbz2-dev 
+libreadline-dev libsqlite3-dev wget llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev
+
+# Install pyenv
+curl -L https://raw.githubusercontent.com/pyenv/pyenv-installer/master/bin/pyenv-installer | bash
+
+# Install make, necessary for installing python 3.9 with pyenv
+sudo apt-get install -y make
+
+# TODO Add the following lines at the end of your .bashrc file if you use bash or .zshrc file if you use zsh
+
+"export PYENV_ROOT="$HOME/.pyenv"
+export PATH="$PYENV_ROOT/bin:$PATH"
+if command -v pyenv 1>/dev/null 2>&1; then
+ eval "$(pyenv init -)"
+fi"
+
+# TODO Restart shell ??
+
+# Make sure pyenv has been installed correctly
+pyenv -v
+
+# Installing python 3.9 with pyenv
+pyenv install 3.9
+
+# Setting python 3.9 as the default local version
+pyenv local 3.9 # is this needed given that we then create a virtual environment?
+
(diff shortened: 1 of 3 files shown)


install-scripts/install-debian.sh
```diff
@@ -0,0 +1,40 @@
+# This installation has been tested on Debian 11 XFCE
+
+# Install curl in order to install Rust and Cargo
+# Install make, necessary for installing python 3.9 with pyenv
+
+sudo apt install -y curl 
+                    \ make
```
klaus993, line 7: The backslash `\` must go in the previous line, for example:

```shell
sudo apt install -y curl \
                               make
```
its function is to escape the newline so the shell interprets it as a continuous command => `sudo apt install -y curl make`

install-scripts/install-debian.sh
```diff
@@ -0,0 +1,40 @@
+# This installation has been tested on Debian 11 XFCE
```
klaus993, line 1: Recommendations on some lines to add at the very beggining of the script (they are useful in almost any bash script you may develop):
```shell
#!/usr/bin/env bash
set -ex
```

* `#!/usr/bin/env bash` tells the operating system to run this script with bash, this is a common convention to be able to run the script directly by doing `./install-debian.sh` (the file must have execute permissions for this)
* the `set` command  is used to modify the different attributes and parameters of the shell environment, in this case:
  * the `-e` flag makes the script end/fail after a command inside the script fails (otherwise after a failure the following commands would be executed anyways)
  * the `-x` flag displays the command being executed, it's useful to debug the script as it's running

install-scripts/install-debian.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Install curl in order to install Rust and Cargo
+# Install make, necessary for installing python 3.9 with pyenv
+
+sudo apt install -y curl 
+                    \ make
+
+# Install Rust and Cargo
+curl https://sh.rustup.rs -sSf | sh
```
klaus993, line 10: For automation, we need all installations to be non-interactive.

Does Rust installation have any interactive steps in which you have to say yes?

klaus993: Also: do we have to source some files to have `rustc` available in our PATH?

install-scripts/install-debian.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Installing python 3.9 with pyenv
+pyenv install 3.9
+
+# Setting python 3.9 as the default local version
+pyenv local 3.9 # is this needed given that we then create a virtual environment?
+
+# Create and enter a virtual environment
+python3.9 -m venv ~/cairo_venv
+source ~/cairo_venv/bin/activate
```
klaus993, line 32: Is this file (`~/cairo_venv/bin/activate`) somewhere in the repo? It seems this is pointing to some local file in your system

install-scripts/install-debian.sh
```diff
@@ -0,0 +1,40 @@
...
+curl https://sh.rustup.rs -sSf | sh
+
+# Make sure Rust has been installed correctly
+rustc --version
+
+# Install pyenv dependencies
+sudo apt-get install -y make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev
+
+# Install pyenv
+curl https://pyenv.run | bash
```
klaus993, line 19: Same as with rust: do we have to source some files to have `pyenv` available in our PATH?

install-scripts/install-debian.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Make sure pyenv has been installed correctly
+pyenv -v
+
+# Installing python 3.9 with pyenv
+pyenv install 3.9
+
+# Setting python 3.9 as the default local version
+pyenv local 3.9 # is this needed given that we then create a virtual environment?
```
klaus993, line 28: > is this needed given that we then create a virtual environment?
Yes, if you look at your next line, you are using python to create a virtual environment, so you need python first

install-scripts/install-debian.sh
martinacantaro: #!/usr/bin/env bash tells the operating system to run this script with bash

Does it execute this line despite it being commented out? :O

PR #771, opened 2023-01-24
Add `secure_run` flag + integrate `verify_secure_runner` into `cairo-run`

fmoletta:
* Add command_line argument `secure_run`
* Integrate `verify_secure_runner` into `cairo_run`
* Remove asserts from `ec_op` programs in `cairo_programs/bad_programs`
* Adapt tests using the programs mentioned above to use `cairo_run` when checking for failures, and check error type
* Fix bug in `EcOpBuiltinRunner::deduce_memory_cell` that skipped the point_on_curve check on the second point
* Fix the values returned when `PoinNotOnCurve` error is returned
* Misc: Fix typo in error string

Depends on: #768 
Notes: An #[ allow::clippy_too_many_arguments] was added in this PR, and an issue #772 was created to solve it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,10 @@
     * `EcOpBuiltinRunner::deduce_memory_cell` now checks that both points are on the curve instead of only the first one
     * `EcOpBuiltinRunner::deduce_memory_cell` now returns the values of the point coordinates instead of the indices when a `PointNotOnCurve` error is returned
 
-* Refactor `Refactor verfify_secure_runner` [#768](https://github.com/lambdaclass/cairo-rs/pull/768)
+* Refactor `Refactor verify_secure_runner` [#768](https://github.com/lambdaclass/cairo-rs/pull/768)
     Public Api changes:
     * Remove builtin name from the return value of `BuiltinRunner::get_memory_segment_addresses`
-    * Simplify the return value of `CairoRunner::get_builtin_segments_info` to `HashMap<isize, usize>`
+    * Simplify the return value of `CairoRunner::get_builtin_segments_info` to `Vec<(usize, usize)>`
     * CairoRunner::read_return_values now receives a mutable reference to VirtualMachine
     Bugfixes:
     * CairoRunner::read_return_values now updates the `stop_ptr` of each builtin after calling `BuiltinRunner::final_stack`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 * Add secure_run flag + integrate verify_secure_runner into cairo-run [#771](https://github.com/lambdaclass/cairo-rs/pull/777)
     Public Api changes:
     * Add command_line argument `secure_run`
-    * Add argument `secure_run: Oprion<bool>` to `cairo_run`
-    * `verify_secure_runner` is now called iside `cairo-run` when `secure_run` is set to true or when it not set and the run is not on `proof_mode`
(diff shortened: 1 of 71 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 ## Cairo-VM Changelog
 
 #### Upcoming Changes
+* Add secure_run flag + integrate verify_secure_runner into cairo-run [#771](https://github.com/lambdaclass/cairo-rs/pull/777)
+    Public Api changes:
+    * Add command_line argument `secure_run`
+    * Add argument `secure_run: Oprion<bool>` to `cairo_run`
```
Oppen, line 7: Typo:
```suggestion
    * Add argument `secure_run: Option<bool>` to `cairo_run`
```

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
...
+    * `verify_secure_runner` is now called iside `cairo-run` when `secure_run` is set to true or when it not set and the run is not on `proof_mode`
```
Oppen, line 8: Typo:
```suggestion
    * `verify_secure_runner` is now called inside `cairo-run` when `secure_run` is set to true or when it not set and the run is not on `proof_mode`
```

src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -1032,38 +1037,50 @@ mod tests {
...
+            "main",
+            false,
+            false,
+            "all",
+            false,
+            None,
+            &mut BuiltinHintProcessor::new_empty(),
+        );
+        assert!(result.is_err());
+        // We need to check this way because CairoRunError doenst implement Debug
```
Oppen, line 1053: Does anything stop us from deriving `Debug`?

Oppen: Typo:
```suggestion
        // We need to check this way because CairoRunError doesn't implement Debug
```

src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -1032,38 +1037,50 @@ mod tests {
...
```
Oppen, line 1077: Typo:
```suggestion
        // We need to check this way because CairoRunError doesn't implement Debug
```

Jrigada (approved):

CHANGELOG.md
fmoletta: Good catch!

CHANGELOG.md
fmoletta: Good catch!

src/vm/runners/builtin_runner/ec_op.rs
fmoletta: Good catch!

fmoletta: I think its because ProgramError uses serde's errors which dont implement Debug, we could probably implement some mapping for them (in another PR)

src/vm/runners/builtin_runner/ec_op.rs
fmoletta: Good catch!

src/vm/runners/builtin_runner/ec_op.rs
fmoletta: This is true for PartialEq, not Debug, Ill fix the comment

Oppen (approved):

PR #773, opened 2023-01-24
feat: add missing 'cairo_type' for Identifier

ptisserand:
In cairo compiled program, funtion returned values types is stored in `cairo_type` key.

# Add `cairo_type` for Identifier

## Description

This pull request add missing `cairo_type` field for `serde::deserialize_program::Identifier`

In a cairo compiled program, each function have an identifier with key `{symbol}.Return`  which contains function output parameter signature in `cairo_type` key.

For instance a cairo function such as:
```
func multi{}() -> (a: felt, b: felt) {
    return (42,7,);
}
```
will have the following identifier in cairo compiled program JSON output:
```
        "__main__.multi.Return": {
            "cairo_type": "(a: felt, b: felt)",
            "type": "type_definition"
        },
```

Using this `cairo_type` information, we can compute function returns size and ensure that mocked value have the same size.

[cairo-foundry related issue](https://github.com/open-dust/cairo-foundry/issues/98)

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/serde/deserialize_program.rs
+++ b/src/serde/deserialize_program.rs
@@ -1279,147 +1279,31 @@ mod tests {
 
     #[test]
     fn deserialize_program_with_type_definition() {
-        let valid_json = r#"{
-            "prime": "0x800000000000011000000000000000000000000000000000000000000000001",
-            "attributes": [],
-            "debug_info": {
-                "instruction_locations": {}
-            },  
-            "builtins": [],
-            "data": [
-            ],
-            "identifiers": {
-                "__main__.a": {
-                    "decorators": [],
-                    "pc": 0,
-                    "type": "function"
-                },
-                "__main__.a.Args": {
-                    "full_name": "__main__.a.Args",
-                    "members": {},
-                    "size": 0,
-                    "type": "struct"
-                },
-                "__main__.a.ImplicitArgs": {
-                    "full_name": "__main__.a.ImplicitArgs",
-                    "members": {},
-                    "size": 0,
-                    "type": "struct"
-                },
-                "__main__.a.Return": {
-                    "cairo_type": "(b: felt)",
-                    "type": "type_definition"
-                },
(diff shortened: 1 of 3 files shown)


Oppen (commented):
Hi @ptisserand, thanks for your PR. It looks quite good, and it's actually a feature we wanted for some time but couldn't allocate time for it, so kudos!

I have only one question about the test program. Is there any chance it could be included as source code instead? We typically reserve the manually compiled folder to programs that for one reason or another need to be exactly that compiled version (say, for example, it's one specific build that triggered a deserializer bug).

ptisserand:
Hi @Oppen 
 
> I have only one question about the test program. Is there any chance it could be included as source code instead? We typically reserve the manually compiled folder to programs that for one reason or another need to be exactly that compiled version (say, for example, it's one specific build that triggered a deserializer bug).

Ok I understand.
Should I add all the compiled program as source code for testing purpose or only some part ?

Oppen:
> Hi @Oppen
> 
> > I have only one question about the test program. Is there any chance it could be included as source code instead? We typically reserve the manually compiled folder to programs that for one reason or another need to be exactly that compiled version (say, for example, it's one specific build that triggered a deserializer bug).
> 
> Ok I understand. Should I add all the compiled program as source code for testing purpose or only some part ?

Ideally a minimal test case for the feature, but I think the whole program is acceptable as well. So, whatever takes you the least extra work is OK for me.

ptisserand:
I have rebased my modification on main and changed test to use JSON extract in source code instead of external file.

fmoletta (approved):

Oppen:
I think there was a misunderstanding. What I meant when I said we'd rather have the sources was that we use the `.cairo` file rather than the `.json` file if possible.

Oppen (approved):

Issue #751, opened 2023-01-19
Extensible VirtualMachineError


tdelabro:
[We](https://github.com/open-dust/cairo-foundry) are using cairo-rs to build a cairo testing suite. 
We are extending the VM with custom behaviors using hints (and other stuff).
We would like to be able to have our custom logic return specific custom VirtualMachineError.

Right now the only way to do so is to use the `VirtualMachineError::CustomHint(String)` variant.
Problem: 
- it only take a String, no custom error types
- for our "other stuff" that are not hints it's a smell to use something called `CustomHint`, yet it's the only thing we have now.

Our ideal implementation would be something like this:
`VirtualMachineError::Other(Box<dyn Error>)` or `VirtualMachineError::Other(anyhow::Error)` (if you don't mind the dependency)

If the lambdaclass team agrees, we can come up with a PR for this.

unbalancedparentheses:
Hey this is great. Let me check with the team but I'm pretty we would love to have this.

tdelabro:
@unbalancedparentheses you got time to talk about it?

hurrikaanig:
Hey, I have talked with @tdelabro, I will be working on this :)

PR #776, opened 2023-01-25
Make `ProgramError` & `CairoRunError` implement `PartialEq`

fmoletta:
This PR aims to make the CairoRunError more user-friendly by making sure it implements PartialEq

### DIFF
--- a/src/types/errors/program_errors.rs
+++ b/src/types/errors/program_errors.rs
@@ -1,19 +1,41 @@
+use std::{fmt::Debug, io};
+
 use felt::PRIME_STR;
-use std::io;
 use thiserror::Error;
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq)]
 pub enum ProgramError {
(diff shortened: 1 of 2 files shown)


Oppen (approved):

Jrigada (approved):

tdelabro:
Hey,
Just saw this one and it will conflict with #751 
The reason is `Box<dyn Error>` does not implement `PartialEq` and we cannot use `Box<dyn Error + PartialEq>` because lot of error types don't implement `PartialEq`.
You will find multiple online threads about this specific issue and how most projects opt for not implementing `PartialEq` on their error types. Instead, they use `assert_matches` for tests.
 
@hurrikaanig  is coming up with a PR that will remove PartialEq on the different Errors types in the project and migrate all tests from `assert!` to `assert_matches!`

So please don't merge  :)

Oppen:
@tdelabro this sounds like a sensible request. @fmoletta do you think we can close this one?

PR #777, opened 2023-01-25
Cache all Cairo programs in Rust workflow

Oppen:
Lowers build times from <30' to about 5' for the Rust workflow by caching compiled cairo programs and the results of running them with cairo-lang, as they should not change unless we change the source files.

Future improvements:
- Single cache for all workflows.
- Parallel creation.


fmoletta (approved):
I love it!

juanbono (approved):

PR #784, opened 2023-01-26
Refactor  `Memory::relocate_memory`

fmoletta:
- Refactor method 
- Add more test cases
- Add more cairo programs to check relocation of temporary memory (these programs dont contain assertions, they are meant to be used by the memory comparison script)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 #### Upcoming Changes
 
+* Refactor `Memory::relocate memory` [#784](https://github.com/lambdaclass/cairo-rs/pull/784)
```
Oppen, line 5:
```suggestion
* Refactor `Memory::relocate_memory` [#784](https://github.com/lambdaclass/cairo-rs/pull/784)
```

src/vm/vm_memory/memory.rs
```diff
@@ -111,73 +110,57 @@ impl Memory {
...
+        }
+        // Move relocated temporary memory into the real memory
+        for index in (0..self.temp_data.len()).rev() {
+            if let Some(base_addr) = self.relocation_rules.get(&index) {
+                let data_segment = self.temp_data.remove(index);
+                // Insert the to-be relocated segment into the real memory
+                let mut addr = *base_addr;
+                if let Some(s) = self.data.get_mut(addr.segment_index as usize) {
+                    s.reserve_exact(data_segment.len())
+                }
```
Oppen, line 154: Won't `Memory::insert()` take care of reserving enough space? Specially since we're inserting reversed, the first insertion should be enough for that.

Oppen (approved):
LGTM but some minor comments.

src/vm/vm_memory/memory.rs
fmoletta: I added the reserve in order to resize the memory segment to fit all of the elements in the temporary segment before inserting its elements one by one (as to minimize resizing)

Oppen: Yes, but the question is whether it'll be inserting by the end, essentially resizing only once anyway.

fmoletta: We are inserting elements one by one incrementing de address with each insertion, so not really

src/vm/vm_memory/memory.rs
```diff
@@ -111,73 +110,57 @@ impl Memory {
         Ok(None)
     }
 
+    // Version of Memory.relocate_value() that doesnt requiere a self reference
```
Jrigada, line 113:
```suggestion
    // Version of Memory.relocate_value() that doesn't require a self reference
```

src/vm/vm_memory/memory.rs
```diff
@@ -111,73 +110,57 @@ impl Memory {
...
+            if let Some(base_addr) = self.relocation_rules.get(&index) {
+                let data_segment = self.temp_data.remove(index);
+                // Insert the to-be relocated segment into the real memory
+                let mut addr = *base_addr;
+                if let Some(s) = self.data.get_mut(addr.segment_index as usize) {
+                    s.reserve_exact(data_segment.len())
+                }
+                for elem in data_segment {
+                    if let Some(value) = elem {
+                        // Rely on Memory::insert to chatch memory insconsistencies
```
Jrigada, line 157:
```suggestion
                        // Rely on Memory::insert to catch memory inconsistencies
```

Jrigada (approved):
Just some minor typos!

PR #789, opened 2023-01-27
Test install actions

klaus993:


test-install-scripts/debian.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
martinacantaro, line 1: Can we delete these tests?

fmoletta (approved):
Great Work!

martinacantaro:
Closes https://github.com/lambdaclass/cairo-rs/issues/736

Oppen (approved):

PR #793, opened 2023-01-27
Refactor  `SignatureBuiltinRunner::add_relocation_rule` + add tests for signature.rs

fmoletta:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/vm/runners/builtin_runner/signature.rs
```diff
@@ -99,57 +99,47 @@ impl SignatureBuiltinRunner {
...
-                let mem_addr_less = if address.offset > 0 {
-                    memory.get(
-                        &address
-                            .sub_usize(1)
-                            .map_err(|_| MemoryError::NumOutOfBounds)?,
-                    )
-                } else {
-                    Ok(None)
+            move |memory: &Memory, addr: &Relocatable| -> Result<Vec<Relocatable>, MemoryError> {
+                let cell_index = addr.offset.mod_floor(&(cells_per_instance as usize));
```
Oppen, line 103:
```suggestion
                let cell_index = addr.offset % cells_per_instance as usize;
```

src/vm/runners/builtin_runner/signature.rs
```diff
@@ -99,57 +99,47 @@ impl SignatureBuiltinRunner {
...
+
+                let (pubkey_addr, message_addr) = match cell_index {
+                    0 => (*addr, addr + 1),
+                    1 => match addr.sub_usize(1) {
+                        Ok(prev_addr) => (prev_addr, *addr),
+                        Err(_) => return Ok(vec![]),
```
Oppen, line 109: Is this correct? The original code returns an error in this case.

fmoletta: The original code does this at the beginning: 
```
if addr.offset % CELLS_PER_SIGNATURE == 0 and addr + 1 in memory:
                pubkey_addr = addr
                msg_addr = addr + 1
            elif addr.offset % CELLS_PER_SIGNATURE == 1 and addr - 1 in memory:
                pubkey_addr = addr - 1
                msg_addr = addr
            else:
                return set()`
```
In the case that addr.offset % CELLS_PER_SIGNATURE (in our case `cell_index`, the matched variable) is 1 and addr - 1 is not in memory (this would be true for addresses with negative offsets, which is the case prevented by addr.sub_usize()), we fall in the `else` clause and return an empty set. So yes, this case is correct.

Oppen: Ok. Note with "the original code" I meant the Rust one before the refactor. Typically we call something a refactor if behavior doesn't change, so I found this to be confusing.

Oppen (approved):

src/vm/runners/builtin_runner/signature.rs
fmoletta: The previous code also does something similar: 
```
let mem_addr_less = if address.offset > 0 {
                    memory.get(
                        &address
                            .sub_usize(1)
                            .map_err(|_| MemoryError::NumOutOfBounds)?,
                    )
                } else {
                    Ok(None)
                };`
```
Here, this error is unreachable as the offset is checked before, it will store Ok(None)
And then the following code will return `Ok(vec![])`, as it wont match the `(1, _, Ok(Some(_element))) `clause.
let (pubkey_addr, msg_addr) = match (address_offset, mem_addr_sum, mem_addr_less) {
                    (0, Ok(Some(_element)), _) => {
                        let pubkey_addr = address;
                        let msg_addr = address + 1_i32;
                        (*pubkey_addr, msg_addr)
                    }
                    (1, _, Ok(Some(_element))) if address.offset > 0 => {
                        let pubkey_addr = address
                            .sub_usize(1)
                            .map_err(|_| MemoryError::EffectiveSizesNotCalled)?;
                        let msg_addr = address;
                        (pubkey_addr, *msg_addr)
                    }
                    _ => return Ok(Vec::new()),
                };

Jrigada (approved):

PR #794, opened 2023-01-28
Separate release build target, use dev for tests

Oppen:
This should speed up testing workflows significantly.


Oppen:
Turned out to be counter-productive, so off it goes.

PR #795, opened 2023-01-28
Switch from Tarpaulin to llvm-cov

Oppen:
llvm-cov is both more accurate and portable (works natively in MacOS) than tarpaulin.
It also allows doing a single run for coverage and it can be used to report coverage from the memory comparison programs (will be addressed later).


Jrigada (approved):

juanbono (approved):
LGTM

martinacantaro:
Installation and usage: https://github.com/taiki-e/cargo-llvm-cov

PR #797, opened 2023-01-31
Parallelize end-to-end tests in CI

Oppen:
End-to-end tests may take a long time, specially when the cache is invalidated. This attempts to exploit parallelism in the runners.
An exception had to be added for running `cairo-run` sequentially when proof mode was enabled due to running more than one concurrently triggered the OOM killer, failing the pipeline.
The cache for all cairo-related artifacts was disabled for testing.


fmoletta (approved):
Looks good!

PR #805, opened 2023-02-01
Check capacity of vector before resizing

Jrigada:
# Check capacity of vector before resizing

## Description
PR to solve the issue reported by Patrick from Fuzzing Labs
https://github.com/lambdaclass/cairo-rs/issues/798

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/vm_memory/memory.rs
+++ b/src/vm/vm_memory/memory.rs
@@ -7,7 +7,7 @@ use felt::Felt;
 use std::{
     borrow::Cow,
     collections::{HashMap, HashSet},
-    mem::swap,
+    mem::{self, swap},
 };
 
 pub struct ValidationRule(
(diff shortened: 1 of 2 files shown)


src/vm/vm_memory/memory.rs
```diff
@@ -63,7 +63,13 @@ impl Memory {
 
         //Check if the element is inserted next to the last one on the segment
         //Forgoing this check would allow data to be inserted in a different index
+
         if segment.len() <= value_offset {
+            if (segment.capacity() + value_offset + 1) * mem::size_of::<MaybeRelocatable>()
```
Oppen, line 68: This will overflow if the condition is true. I'm personally not sure it's valuable to catch this case, and certainly I don't think this is the proper place to fix it. Instead, we should make sure offsets in relocatables don't exceed 47 bits, which is how big they're specified to be.

fmoletta: + 1

PR #810, opened 2023-02-02
Fix codecov and include e2e tests

Oppen:
This fixes codecov.io uploads by collapsing the two steps again, since using the token it no longer fails on upload and the culprit was the repo wasn't checked out in the second workflow.
It also changes the compare runs so they are built with llvm-cov support and are included in the coverage report.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,0 @@
-fixes:
-  - "/home/runner/work/cairo-rs/::"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,3 @@
 fixes:
   - "lcov.info/lcov.info::"
+  - "/home/runner/work/cairo-rs/cairo-rs/::"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+fixes:
+  - "lcov.info/lcov.info::"
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,3 @@
 fixes:
   - "lcov.info/lcov.info::"
-  - "/home/runner/work/cairo-rs/cairo-rs/::"
+  - "/home/runner/work/cairo-rs/::"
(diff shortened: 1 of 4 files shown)


codecov.yml
```diff
@@ -0,0 +1,2 @@
+fixes:
```
oboulant, line 1: There is a `codecov.yml` file also in : [.github/codecov.yml](https://github.com/lambdaclass/cairo-rs/blob/main/.github/codecov.yml). Maybe it can be deleted its content moved to this top level `codecov.yml` alongside your addition ?

Oppen: Yeah, I'm not sure which is preferred or whether both are being read. All I know for sure is codecov does report my `fixes` field, so it is reading that one for sure. It's not needed anyway I think, it's more a leftover from attempts to fix it following the wrong clue.

Oppen: Hey, thanks for pointing that out! It actually works without the new file :)

oboulant: Seems like it yes ! 

I wonder how it will play out whenever a file covered by tests is impacted in the PR. Indeed, in the file uploaded to codecov, I still see some `/home/runner/work/cairo-rs/` (for instance [here](https://app.codecov.io/github/lambdaclass/cairo-rs/commit/1a3b2a54e813cf2e342e111d251d6165c52dc003)). At the end, maybe codecov is perfectly fine with it and all the gymnastic with the `fixes` is not needed.

Jrigada (approved):

oboulant (approved):

codecov.yml
Oppen: Yes, it was likely not needed. I could try removing it.

Oppen: Oh, I see I did 🤦 
It was a long week.

Juan-M-V (approved):

PR #815, opened 2023-02-02
ci(codecov): upload codecov-report rather than lcov.info

oboulant:
# Upload Codecov reports

Related issue : https://github.com/lambdaclass/cairo-rs/issues/695 

## Description

Codecov does not see the coverage report

This PR modifies the GHA called `rust`. It uploads to codecov the file `codecov-report` rather than `lcov.info`. Moreover, it removes the step `Retrieve cached report` in job `upload-codecov` (the artifact is normally accessible by the `upload-codecov` job without downloading it. 

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -83,14 +83,9 @@ jobs:
     needs: build
     runs-on: ubuntu-20.04
     steps:
-    - name: Retrieve cached report
-      uses: actions/download-artifact@master
-      with:
-        name: codecov-report
-        path: lcov.info
     - name: Upload coverage to codecov.io
       uses: codecov/codecov-action@v3
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
-        files: lcov.info
+        files: ./codecov-report
         fail_ci_if_error:     true


.github/workflows/rust.yml
```diff
@@ -83,14 +83,9 @@ jobs:
...
-      uses: actions/download-artifact@master
-      with:
-        name: codecov-report
-        path: lcov.info
     - name: Upload coverage to codecov.io
       uses: codecov/codecov-action@v3
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
-        files: lcov.info
+        files: ./codecov-report
```
Oppen, line 90: Have you checked this file exists?

Oppen:
Hi @oboulant. There's already a PR that fixes this. I'm sorry we stepped on each other's toes.

oboulant:
> 

Oh, sorry about that @Oppen ! Could you point me the the PR ? Now I am curious about the fix :) 

Initially wrongly thought it was a matter of artefact naming between `build` and `upload-codecov` jobs ! 

Thought I found interesting clues looking at https://docs.codecov.com/docs/fixing-paths but does not seem to do the trick either !

oboulant:
Found it : https://github.com/lambdaclass/cairo-rs/pull/810 

Closing this PR !

Oppen:
Yeah, it's #810
The clue in codecov.io was misleading, that's why I took so much effort to fix. The real issue was that at some point upload started failing, because of that it caused huge delays in re-running the job just for that, so we started uploading an artifact and downloading it. The error was we never checked out the repo in the second part of the job 🤦 
So, now that we use the token, the upload no longer fails, so the easy fix was to just upload as part of the main job.

Oppen:
TBF it's my bad, I was working on something else in that PR but needed the codecov upload working to actually confirm my changes made sense, so I fixed it out of annoyance in there.

PR #817, opened 2023-02-02
Adding proptests for feltbigint 

martinacantaro:
# Adding proptests for feltbigint 

## Description

Testing addition: unit test for adding two zeroes and proptests for adding big numbers.

Closes https://github.com/lambdaclass/cairo-rs/issues/803

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added
  - [x] Documentation has been added/updated. Tests are documented in the comments


Juan-M-V (approved):

Juan-M-V (approved):

felt/src/bigint_felt.rs
```diff
@@ -483,6 +483,9 @@ impl<const PH: u128, const PL: u128> Div for FeltBigInt<PH, PL> {
     // In Felts `x / y` needs to be expressed as `x * y^-1`
     #[allow(clippy::suspicious_arithmetic_impl)]
     fn div(self, rhs: Self) -> Self::Output {
+        if rhs.to_i8().unwrap() == 0 {
+            panic!("Can't divide Felt by zero")
```
Jrigada, line 487: I think here we should throw an Error instead of panic. We can ask @Oppen what is the correct way to handle these things.

Oppen: I think it was mentioned in Slack, `panic!` is the only possible way to handle it in the operator itself, as it's expected to be chainable and thus can't return a `Result`. `checked_div` returns a `Result` for when things are expected to handle failure.
That said, this will crash for an unrelated reason: `to_i8().unwrap()` will crash for values over `127` or under `-128` (the latter should never happen tho). Use `.is_zero()` instead.

Oppen (commented):
The tests seem repetitive. It may be time to add a macro for this common test or something. Otherwise, just address the comments we left with @Jrigada and it's good to go.

felt/src/bigint_felt.rs
Juan-M-V: Resolved: https://github.com/lambdaclass/cairo-rs/pull/817/commits/0d286f661b3b9a74eaf64b875925b7f4ee03fa6e

felt/src/bigint_felt.rs
```diff
@@ -1125,29 +1167,7 @@ mod tests {
...
-            let p:BigUint = BigUint::parse_bytes(CAIRO_PRIME.to_string().as_bytes(), 16).unwrap();
-            x -= y;
-            let as_uint = &x.to_biguint();
-            prop_assert!(as_uint < &p, "{}", as_uint);
-        }
-
-        #[test]
-        // Property-based test that ensures that vectors of three of values that are randomly generated each time tests are run, that performing an iterative sum returns a result that is inside of the range [0, p]. The test is performed 100 times each run.
+        // Property-based test that ensures, vectors of three of values that are randomly generated each time tests are run, that performing an iterative sum returns a result that is inside of the range [0, p]. The test is performed 100 times each run.
             fn sum_bigint_felt_within_field(ref x in "([1-9][0-9]*)", ref y in "([1-9][0-9]*)", ref z in "([1-9][0-9]*)") {
```
Oppen, line 1171: Indentation seems messed up here.

felt/src/bigint_felt.rs
```diff
@@ -1158,17 +1178,5 @@ mod tests {
...
-        // Property-based test that ensures that the remainder of a division between two random bigint felts returns a result that is inside of the range [0, p]. The test is performed 100 times each run.
-        fn rem_bigint_felt_within_field(ref x in "([1-9][0-9]*)", ref y in "([1-9][0-9]*)") {
-            let x = FeltBigInt::<FIELD_HIGH, FIELD_LOW>::parse_bytes(x.as_bytes(), 10).unwrap();
-            let y = FeltBigInt::<FIELD_HIGH, FIELD_LOW>::parse_bytes(y.as_bytes(), 10).unwrap();
-            let p:BigUint = BigUint::parse_bytes(CAIRO_PRIME.to_string().as_bytes(), 16).unwrap();
-
-            let result = x % y;
-            let as_uint = result.to_biguint();
-            prop_assert!(&as_uint < &p, "{}", as_uint);
-        }
```
Oppen, line 1172: Maybe test it's always `0` instead of removing the test?

felt/src/bigint_felt.rs
Juan-M-V: Resolved: https://github.com/lambdaclass/cairo-rs/pull/817/commits/497b276b2132b1c7059e913e70c0618f948ce35d

Oppen (approved):

PR #829, opened 2023-02-09
Test from_bytes_be for bigint felt

martinacantaro:
# Test from_bytes_be for bigint felt

## Description

Test the from_bytes_be function implementation for FeltBigInts

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added
  - [x] Documentation has been added/updated. The test is documented in the code comments


felt/src/bigint_felt.rs
```diff
@@ -820,25 +820,67 @@ impl fmt::Display for ParseFeltError {
...
+    }
+
+    #[test]
+    // Tests that the result of performing a division between two zeros results in zero.
+    fn div_zeros() {
+        let a = FeltBigInt::<FIELD_HIGH, FIELD_LOW>::new(0);
+        let b = FeltBigInt::new(0);
+        let c = FeltBigInt::new(0);
+
+        assert_eq!(&a / &b, c);
```
Juan-M-V, line 882: There is a bug in the test and in the operation, dividing by 0 should panic

martinacantaro: fixed

felt/src/bigint_felt.rs
```diff
@@ -483,6 +483,9 @@ impl<const PH: u128, const PL: u128> Div for FeltBigInt<PH, PL> {
     // In Felts `x / y` needs to be expressed as `x * y^-1`
     #[allow(clippy::suspicious_arithmetic_impl)]
     fn div(self, rhs: Self) -> Self::Output {
+        if rhs.to_i8().unwrap() == 0 {
+            panic!("Can't divide Felt by zero")
+        }
```
fmoletta, line 488: Should't we apply this same logic to the other `Div` implementations below this one?
(lines 499 and 514)

fmoletta: Also, it would be useful to have some documentation for these implementations indicating possible panics

Oppen: Use `is_zero()` here. Otherwise it will crash with valid inputs (`> 127`).

Oppen: I agree with everything @fmoletta said. We should implement `checked_div` as well so users have the option of handling the error instead, too.

Oppen (changes requested):

Juan-M-V:
I've made a bit of a mess as I've corrected the things you mentioned plus making codecov happy. All the changes to Div are split between this PR and https://github.com/lambdaclass/cairo-rs/pull/817. If we can get any of the two merged I'll finish up correcting whichever is left

felt/src/bigint_felt.rs
Juan-M-V: https://github.com/lambdaclass/cairo-rs/pull/829/commits/97ffb059e3f750830436c3eab8161513eab193d9 should resolve all this

martinacantaro:
Ready for another review :)

Oppen (approved):

PR #832, opened 2023-02-13
Move `secure_run` flag to `CairoRunConfig` struct

fmoletta:


src/cairo_run.rs
```diff
@@ -24,6 +24,7 @@ pub struct CairoRunConfig<'a> {
     pub print_output: bool,
     pub layout: &'a str,
     pub proof_mode: bool,
+    pub secure_run: Option<bool>,
```
Oppen, line 27:
```suggestion
    pub secure_run: bool,
```

Oppen (approved):

src/cairo_run.rs
fmoletta: This is the original format. 
The logic for determining its value is
if Some(bool) -> bool
if None -> if proof_mode: false else true
This logic cant be applied with a simple boolean value

Juan-M-V (approved):

PR #836, opened 2023-02-14
feat: integrate CodSpeed

art049:
# Integrate CodSpeed

## Description

This adds the integration with CodSpeed to report performance data from the criterion benches continuously. Thanks, @Eikix for the idea 😀.

Instead of using criterion.rs directly, I swap it with the compatibility crate for CodSpeed, if you're interested in more details about that, there are some more details [here](https://docs.codspeed.io/benchmarks/rust).
I also added a new GitHub Action workflow that will run the instrumented benches and upload the results to CodSpeed. This will report performance changes on any pull requests once merged.

Thrilled to have your feedback on the tool since you're relying a lot on benchmarks from what I saw !

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
     - name: Install Rust 1.66.1
       uses: actions-rs/toolchain@v1
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,0 +1,31 @@
+name: codspeed
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ '*' ]
+
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Install Rust 1.66.1
+      uses: actions-rs/toolchain@v1
+      with:
+          toolchain: 1.66.1
+          override: true
+    - name: Cache Rust installation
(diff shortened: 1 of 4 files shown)


Eikix:
Hi @HermanObst! This is the PR I told you about, it's aimed at having faster and granular performance assessment at the PR level:)!

art049:
Okay, it seems to work now 🎉
While having a look I came across a little benchmark naming issue that I fixed in #838.
It'd be perfect to merge this other PR first so that the names are good from the beginning!

Oppen:
Hi @art049. I'm sad to tell you we're rejecting this PR for a few reasons, the most important being that we'd rather not tie up our code to a proprietary service. It would possibly be alright if the workflow touched only the GitHub actions, but touching the benchmark code is a bit of a red flag, specially considering that, IIUC the service's website, the new code will refuse to run benchmarks locally, outside of the service's servers, which is unacceptable for the project. We already have a solution in place for reliable benchmarks in the CI based on IAI. Admittedly it still needs a bit of work for reporting, but it's reliable and transparent.
As a side-note, the timings reported make me think the benchmarks are failing, but as currently implemented this is opaque in the report: those constant-time microsecond timings are consistent with errors rather than timings of runs calibrated to take seconds. It's probably looking for the files in the wrong place, or not checking them in the first place.
All that said, the team and I appreciate that you put time an effort to this PR and look forward for future contributions.

art049:
Hello @Oppen, thanks for the feedback.
Just to clarify, the compatibility library is transparent and the existing benchmarks are untouched(i.e. you can still run criterion.rs without any issues).
For the timings, I guess it's only a scaling factor issue. But anyways, if at some point you're willing to use the solution, don't hesitate to ping me.

PR #844, opened 2023-02-15
Fix `KeccakBuiltinRunner::deduce_memory_cell`

fmoletta:
The method should write the output values of keccak into the memory
As a result of this fix, multiple method signatures changed due to the added mutability requirement
Depends on #842


fmoletta:
This fix is currently invalid

PR #851, opened 2023-02-16
Get rid of uninformative `FailedToComputeOperands` error

fmoletta:
Currently, when retrieving ids values from memory, all internal memory errors get replaced by the `FailedToComputeOperands` error, which doesnt give us much information about what went wrong.
This PR aims to remove this error, and replace it with variants that state which is the identifier that couldnt be computed.
Also, methods reflecting internal computation of ids values and addresses will have their return values changed to Option<>, as their errors arent and wont be used

### DIFF
--- a/src/vm/errors/hint_errors.rs
+++ b/src/vm/errors/hint_errors.rs
@@ -10,8 +10,12 @@ use super::{exec_scope_errors::ExecScopeError, vm_errors::VirtualMachineError};
 pub enum HintError {
     #[error("HintProcessor failed retrieve the compiled data necessary for hint execution")]
     WrongHintData,
-    #[error("Failed to get ids for hint execution")]
-    FailedToGetIds,
+    #[error("Unknown identifier {0}")]
+    UnknownIdentifier(String),
+    #[error("Expected ids.{0} to be an Integer value")]
+    IdentifierNotInteger(String),
+    #[error("Expected ids.{0} to be a Relocatable value")]
+    IdentifierNotRelocatable(String),
     #[error("Tried to compute an address but there was no register in the reference.")]
     NoRegisterInReference,
     #[error("Custom Hint Error: {0}")]


src/hint_processor/builtin_hint_processor/secp/bigint_utils.rs
```diff
@@ -7,14 +7,55 @@ use crate::{
...
 use felt::Felt;
-use std::collections::HashMap;
+use std::{borrow::Cow, collections::HashMap};
+
+#[derive(Debug, PartialEq)]
+pub(crate) struct BigInt3<'a> {
+    pub d0: Cow<'a, Felt>,
+    pub d1: Cow<'a, Felt>,
+    pub d2: Cow<'a, Felt>,
+}
```
Oppen, line 27: Why `Cow<_>`?

src/vm/errors/hint_errors.rs
```diff
@@ -10,10 +10,18 @@ use super::{exec_scope_errors::ExecScopeError, vm_errors::VirtualMachineError};
...
-    #[error("Failed to get ids for hint execution")]
-    FailedToGetIds,
-    #[error("Tried to compute an address but there was no register in the reference.")]
-    NoRegisterInReference,
+    #[error("Unknown identifier {0}")]
+    UnknownIdentifier(String),
+    #[error("Expected ids.{0} to be an Integer value")]
+    IdentifierNotInteger(String),
+    #[error("Expected ids.{0} to be a Relocatable value")]
+    IdentifierNotRelocatable(String),
```
Oppen, line 18: It may be a good idea (or not, I'm not really sure) to include the resolved address.

Oppen (approved):
LGTM but see comments.

src/hint_processor/builtin_hint_processor/secp/bigint_utils.rs
fmoletta: I wanted to use references when possible instead of cloning values, and using references wasn't possible

src/vm/errors/hint_errors.rs
fmoletta: On it!

fmoletta: Done

juanbono (approved):

PR #852, opened 2023-02-17
Simplify `Memory::get return` type to `Option`

fmoletta:
Change Memory::get & Memory::get_maybe to return an Option instead of Result<Option
Change `MaybeReloctable::get_int_ref` & `MaybeReloctable::get_relocatable` to return Option instead of Result (previous error had incorrect info)
Fix errors used by the range_check builtin (add info & change name) + remove use of these errors outside of range_check methods
Add error `UnknownMemoryCell`
Refactor a hint to use `get_integer_range` instead of using get_range and the checking for None or non-integer valuesl


Oppen (approved):

juanbono (approved):
LGTM

PR #853, opened 2023-02-17
Fix `CairoRunner::write_output` so that it prints missing and relocatable values

fmoletta:
Currently our vm only prints values in the output segment if they are integer values, and fails if they are not.
This PR aims to fix this behaviour and also print missing and relocatable values as the original vm does.
Expample program:
```
%builtins output

func main{output_ptr: felt*}() {
    //Memory Gap + Relocatable value
    assert [output_ptr + 1] = cast(output_ptr, felt);
    let output_ptr = output_ptr + 2;
    return ();
}
```
Example program output:
```
Program Output: 
<missing>
2:0
```
Depends on #852

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -2822,6 +2822,49 @@ mod tests {
     }
 
     #[test]
+    /*Program used:
+    %builtins output
+
+    func main{output_ptr: felt*}() {
+        //Memory Gap + Relocatable value
+        assert [output_ptr + 1] = cast(output_ptr, felt);
+        let output_ptr = output_ptr + 2;
+        return ();
+    }*/
+    fn write_output_from_program_gap_relocatable_output() {
+        //Initialization Phase
+        let program = program!(
+            builtins = vec![OUTPUT_BUILTIN_NAME],
+            data = vec_data!(
+                (4612671187288162301),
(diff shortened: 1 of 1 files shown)


src/vm/runners/cairo_runner.rs
```diff
@@ -895,13 +895,18 @@ impl CairoRunner {
...
                 .segments
                 .memory
-                .get_integer((segment_index as isize, i).into())
-                .map_err(|_| RunnerError::MemoryGet((segment_index as isize, i).into()))?
-                .to_bigint();
-            writeln!(stdout, "{value}").map_err(|_| RunnerError::WriteFail)?;
+                .get(&Relocatable::from((segment_index as isize, i)))
+            {
+                Some(val) => match val.as_ref() {
+                    MaybeRelocatable::Int(num) => format!("{}", num.to_bigint()),
```
Oppen, line 904: For a later refactor: it could be a better idea to implement `Display` for `Felt` instead.

Oppen (approved):

src/vm/runners/cairo_runner.rs
fmoletta: It is implemented, the reason for using `to_bigint()` is to change the range of the felt (from [0,P] to [-P/2, P/2})

juanbono (approved):

PR #854, opened 2023-02-17
Move memory-related errors to `MemoryError`

fmoletta:
- Remove unused error variants
- Move memory-related error variants to MemoryError
- Change memory getters to return `MemoryError` instead of `VirtualMachineError`
- Change all memory-related errors in hint from  HintError::Internal(VmError::... to HintError::Memory(MemoryError::...
Depends on #852 
Will probably collide with #851, should be merged afterwards


juanbono (approved):
LGTM

PR #857, opened 2023-02-23
Integrate accessed_addresses logic into memory structure

fmoletta:
This PR aims to mitigate the performance loss of having to keep a growing structure to keep track of accessed addresses and instead integrate this logic into the memory structure itself
Notable functionality differences:
- Cant mark as accessed addresses that dont exist in memory
- Marking the program segment as accessed is now done during `initialize_state` instead of `initialize_vm`

Other changes in this PR:
- Remove unused field `_program_base` in `VirtualMachine`
- Fix `memory` macro to accept temporary addresses

Average performace gain (Using programs in bench folder): 5%


Oppen (approved):

Oppen:
> Cant mark as accessed addresses that dont exist in memory

I think this can be fixed by moving the `Option` inside the `MemoryCell`, so it can be both marked as accessed and its value be `None`.

fmoletta:
> > Cant mark as accessed addresses that dont exist in memory
> 
> I think this can be fixed by moving the `Option` inside the `MemoryCell`, so it can be both marked as accessed and its value be `None`.

It doesn't really make sense for an address that doesn't exist to me mark as accessed

Oppen:
> It doesn't really make sense for an address that doesn't exist to me mark as accessed

I agree.

Jrigada (approved):

Issue #823, opened 2023-02-06
Add tests for num_traits for Felts


martinacantaro:
Add tests for num_traits for Felts:
- [x] Bounded
- [x] FromPrimitive
- [ ] Num
- [ ] One
- [x] Pow
- [ ] Signed
- [x] ToPrimitive
- [ ] Zero

PatStiles:
Willing to take this on!

oboulant:
Sorry @PatStiles : did not reloaded the page before pushing tests for the Zero trait ! 
Please, feel free to take the rest of the traits and even propose a better alternative to what I wrote in here : https://github.com/lambdaclass/cairo-rs/pull/831

haroune-mohammedi:
@PatStiles are you working on this ? If not, I'm willing to step up and take it

PR #858, opened 2023-02-24
Test untested functions and traits of  `Felt`

haroune-mohammedi:
## Description
Issue: #823

We added three tests for `num_traits::One`:

- [x] `one_value` to test `One::one`
- [x] `is_one` to test `One::is_zero`
- [x] `one_multiplicative_identity` to test the multiplicative identity:

```
x * 1 = x       ∀ x
1 * x = x       ∀ x
```
- [x] `zero_additive_identity` to test the additive identity: of `num_traits::Zero`:
``` 
x + 0 = x       ∀ x
0 + x = x       ∀ x
```

Tests for `num_traits::Signed` implementation
   
- [x] `felt_is_never_negative`
- [x] `non_zero_felt_is_always_positive`
- [x]  `non_zero_felt_signum_is_always_one`
- [x] `signum_of_zero_is_zero`
- [x]  `sub_abs`
- [x]  `abs`

Tests for `Integer`:
- [x] `divides_doesnt_panic`
- [x] `gcd_doesnt_panic`
- [x] `is_even_always_true`
- [x] `is_odd_always_false`

Tests for `FromPrimitive` and `ToPrimitive`:
- [x] `from_u64_and_to_u64_primitive`
- [x] `from_i64_and_to_i64_primitive`

Tests for:
- [x] `modpow_in_range`
- [x] `sqrt_in_range`
- [x] `sqrt_is_inv_square`

Tests for `Add`,  `AddAssign`,  `Add<u32>`, `Sub<u32>`, `Sub<usize>`:

- [x] `add_u32_in_range`
- [x] `add_u32_is_inv_sub`
- [x] `sub_u32_in_range`
- [x] `sub_u32_is_inv_add`
- [x] `sub_uszie_in_range`
- [x] `sub_usize_is_inv_add`
- [x] `add_in_range`
- [x] `add_is_inv_sub`
- [x] `add_assign_in_range`

We tested the reference variant for 

- [x] `sub_assign_in_range`
- [x] `neg_in_range`
- [x] `pow_in_range`

As a result of writing tests, we discoverd bugs and fixed the implementation of `is_even`, `is_odd` and `divides`

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/felt/src/lib.rs
+++ b/felt/src/lib.rs
@@ -1074,7 +1074,7 @@ mod test {
     }
 
     #[test]
-    // Checks that the result of perfforming a bit and operation between zeroes is zero
+    // Checks that the result of performing a bit and operation between zeroes is zero
     fn bit_and_zeros_in_range() {
         let x = Felt::new(0);
         let y = Felt::new(0);
@@ -1130,5 +1130,34 @@ mod test {
         let felt_non_zero = Felt::new(3);
         assert!(felt_zero.is_zero());
         assert!(!felt_non_zero.is_zero())
+    }
+
+    #[test]
+    fn one_value() {
+        let one = BigUint::one();
+        let felt_one = Felt::one().to_biguint();
+        assert_eq!(one, felt_one)
+    }
+
+    #[test]
+    fn is_one() {
+        let felt_one = Felt::one();
+        let felt_non_one = Felt::new(8);
+        assert!(felt_one.is_one());
+        assert!(!felt_non_one.is_one())
+    }
+
+    /// Tests the multiplicative identity of the implementation of One trait for felts
+    ///
+    /// ```{.text}
+    /// a * 1 = a       ∀ a
+    /// 1 * a = a       ∀ a
+    /// ```
+    #[test]
+    fn one_multiplicative_identity() {
+        let one = Felt::one();
+        let felt = Felt::new(99);
+        assert_eq!(felt, &felt * &one);
+        assert_eq!(felt, &one * &felt)
     }
 }
--- a/felt/src/lib.rs
+++ b/felt/src/lib.rs
@@ -1088,7 +1088,6 @@ mod test {
             let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
             let y = Felt::parse_bytes(y.as_bytes(), 10).unwrap();
             prop_assert!(x.is_multiple_of(&y));
-            // prop_assert!(x.divides(&y));
         }
 
         #[test]
@@ -1146,6 +1145,7 @@ mod test {
             prop_assert_eq!(&x, &(&x * &one));
             prop_assert_eq!(&x, &(&one * &x));
         }
+
         #[test]
         fn non_zero_felt_is_always_positive(ref x in FELT_NON_ZERO_PATTERN) {
             let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
@@ -1157,7 +1157,6 @@ mod test {
             let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
             prop_assert!(!x.is_negative())
         }
-
 
         #[test]
         fn non_zero_felt_signum_is_always_one(ref x in FELT_NON_ZERO_PATTERN) {
--- a/felt/src/lib.rs
+++ b/felt/src/lib.rs
@@ -822,6 +822,9 @@ mod test {
     use super::*;
     use proptest::prelude::*;
 
+    const FELT_PATTERN: &str = "(0|[1-9][0-9]*)";
+    const FELT_NON_ZERO_PATTERN: &str = "[1-9][0-9]*";
+
     proptest! {
         #[test]
         #[allow(deprecated)]
@@ -1068,7 +1071,6 @@ mod test {
             prop_assert_eq!(&x, &(&zero + &x));
         }
 
-
         /// Tests the multiplicative identity of the implementation of One trait for felts
         ///
         /// ```{.text}
@@ -1081,6 +1083,41 @@ mod test {
             let one = Felt::one();
             prop_assert_eq!(&x, &(&x * &one));
             prop_assert_eq!(&x, &(&one * &x));
+        }
+        #[test]
+        fn non_zero_felt_is_always_positive(ref x in FELT_NON_ZERO_PATTERN) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert!(x.is_positive())
+        }
+
+        #[test]
+        fn felt_is_never_negative(ref x in FELT_PATTERN) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert!(!x.is_negative())
+        }
+
+
+        #[test]
+        fn non_zero_felt_signum_is_always_one(ref x in FELT_NON_ZERO_PATTERN) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let one = Felt::one();
+            prop_assert_eq!(x.signum(), one)
+        }
+
+        #[test]
+        fn sub_abs(ref x in FELT_PATTERN, ref y in FELT_PATTERN) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let y = Felt::parse_bytes(y.as_bytes(), 10).unwrap();
+
+            let expected_abs_sub = if x > y {&x - &y} else {&y - &x};
+
+            prop_assert_eq!(&x.abs_sub(&y), &expected_abs_sub)
+        }
+
+        #[test]
+        fn abs(ref x in FELT_PATTERN) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert_eq!(&x, &x.abs())
         }
     }
 
@@ -1174,5 +1211,11 @@ mod test {
         let felt_non_one = Felt::new(8);
         assert!(felt_one.is_one());
         assert!(!felt_non_one.is_one())
+    }
+
+    #[test]
+    fn signum_of_zero_is_zero() {
+        let zero = Felt::zero();
+        assert_eq!(&zero.signum(), &zero)
     }
 }
--- a/felt/src/lib.rs
+++ b/felt/src/lib.rs
@@ -874,10 +874,15 @@ mod test {
         // Property-based test that ensures, for 100 {x} and {y} values that are randomly generated each time tests are run, that a subtraction with assignment between two felts {x} and {y} and doesn't fall outside the range [0, p]. The values of {x} and {y} can be either [0] or a very large number.
         fn sub_assign_in_range(ref x in "(0|[1-9][0-9]*)", ref y in "(0|[1-9][0-9]*)") {
             let mut x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
-            let y = &Felt::parse_bytes(y.as_bytes(), 10).unwrap();
+            let y = Felt::parse_bytes(y.as_bytes(), 10).unwrap();
             let p = &BigUint::parse_bytes(PRIME_STR[2..].as_bytes(), 16).unwrap();
 
-            x -= y;
+            x -= y.clone();
+            let as_uint = &x.to_biguint();
+            prop_assert!(as_uint < p, "{}", as_uint);
+
+            // test reference variant
+            x -= &y;
             let as_uint = &x.to_biguint();
             prop_assert!(as_uint < p, "{}", as_uint);
         }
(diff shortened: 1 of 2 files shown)


fmoletta (approved):
LGTM

Jrigada (approved):
Thanks for the PR, LGTM

haroune-mohammedi:
I have some suggestions that involves changes to the tests I haven't wrote so I figured I should ask you guys first

- Use `Felt::max_value` instead of parsing `PRIME_STR` inside all the functions
- use ref implementation to implement owned ones instead of implementing the same thing twice
- Why are we only testing for in_range, why don't we test the result values of the operations like: add, neg, sub ..., 
- Why do we need to implement the same operation for `Felt`, `&Felt`, `u32`, `usize` ... check https://docs.rs/impl_ops/0.1.1/impl_ops/index.html that could be used to get rid of the boilerplate

haroune-mohammedi:
I've added tests for `Felt` and updated the description

@fmoletta @Jrigada would you please re-review and approve the workflows, also is there any way I can run the workflows without waiting approval in my fork ?

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
             let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
             let y = Felt::parse_bytes(y.as_bytes(), 10).unwrap();
             prop_assert!(x.is_multiple_of(&y));
+            // prop_assert!(x.divides(&y));
```
Oppen, line 1091: Let's remove this.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+        }
+
+        #[test]
+        fn divides_doesnt_panic(ref x in "(0|[1-9][0-9]*)", ref y in "(0|[1-9][0-9]*)") {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let y = Felt::parse_bytes(y.as_bytes(), 10).unwrap();
+            prop_assert!(x.divides(&y));
```
Oppen, line 1098: This test will fail for `y == 0`. I'm not 100% what should happen for `x == 0` if this is not returning a `Result<_>`, but we're passing it in the test as well.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+            let y = Felt::parse_bytes(y.as_bytes(), 10).unwrap();
+            let gcd1 = x.gcd(&y);
+            let gcd2 = y.gcd(&x);
+            prop_assert_eq!(gcd1, gcd2);
+        }
+
+        #[test]
+        fn is_even(ref x in "(0|[1-9][0-9]*)") {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert_eq!(x.is_even(), x.to_biguint().is_even());
```
Oppen, line 1113: This will fail for a bunch of even uints. Take `CAIRO_PRIME + 1` for example. That's even before taking the modulus, but it becomes `1` afterwards.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+        }
+
+        #[test]
+        fn is_odd(ref x in "(0|[1-9][0-9]*)") {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert_eq!(x.is_odd(), x.to_biguint().is_odd());
```
Oppen, line 1119: The same happens here, `CAIRO_PRIME + 2` is odd, but `CAIRO_PRIME + 2 mod CAIRO_PRIME` is `2` which is even.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+        /// x * 1 = x       ∀ x
+        /// 1 * x = x       ∀ x
+        /// ```
+        #[test]
+        fn one_multiplicative_identity(ref x in "(0|[1-9][0-9]*)") {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let one = Felt::one();
+            prop_assert_eq!(&x, &(&x * &one));
+            prop_assert_eq!(&x, &(&one * &x));
+        }
```
Oppen, line 1148: Nit: add a newline.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+            prop_assert!(!x.is_negative())
+        }
+
+
+        #[test]
+        fn non_zero_felt_signum_is_always_one(ref x in FELT_NON_ZERO_PATTERN) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let one = Felt::one();
+            prop_assert_eq!(x.signum(), one)
+        }
```
Oppen, line 1167: IIRC, the `is_positive`, `signum` and `is_negative` for type `Felt` are meant for the signed representation, which reads a field element as the range `(-CAIRO_PRIME/2, CAIRO_PRIME/2)`, where `x > CAIRO_PRIME/2` represents a negative field element. This would make these assumptions wrong.

Oppen: There was the intent to make a separate type for singed field elements to avoid this confusion, I don't exactly remember if it was applied (I'll check later today and come back with a definitive answer).

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+            let expected_abs_sub = if x > y {&x - &y} else {&y - &x};
+
+            prop_assert_eq!(&x.abs_sub(&y), &expected_abs_sub)
+        }
+
+        #[test]
+        fn abs(ref x in FELT_PATTERN) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert_eq!(&x, &x.abs())
+        }
```
Oppen, line 1183: For the same reason as above, I'm not sure if this is true.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let p = BigUint::parse_bytes(PRIME_STR[2..].as_bytes(), 16).unwrap();
+            let x_add_y = (x + y).to_biguint();
+            prop_assert!(x_add_y < p, "{}", x_add_y);
+        }
+
+        #[test]
+        fn add_u32_is_inv_sub(ref x in FELT_PATTERN, y in any::<u32>()) {
+            let x = &Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let expected_y = (x.clone() + y - x).to_u32().unwrap();
```
Oppen, line 1223: `expected_y` does not necessarily fit in a `uint32`. `Felt(CAIRO_PRIME - 1) - u32::MAX` is still greater than `u32::MAX`.

Oppen: You should convert `y` to `BigUint` before the comparison instead.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+            prop_assert_eq!(expected_y, y, "{}", expected_y);
+        }
+
+        #[test]
+        fn sub_u32_in_range(ref x in FELT_PATTERN, y in any::<u32>()) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let p = BigUint::parse_bytes(PRIME_STR[2..].as_bytes(), 16).unwrap();
```
Oppen, line 1230: Don't we already have this prime parsed?

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+        #[test]
+        fn sub_u32_is_inv_add(ref x in FELT_NON_ZERO_PATTERN, y in any::<u32>()) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert_eq!(x.clone() - y + y, x)
+        }
+
+        #[test]
+        fn sub_uszie_in_range(ref x in FELT_PATTERN, y in any::<usize>()) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let p = BigUint::parse_bytes(PRIME_STR[2..].as_bytes(), 16).unwrap();
```
Oppen, line 1244: Ditto.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+        fn sub_usize_is_inv_add(ref x in FELT_PATTERN, y in any::<usize>()) {
+            let x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            prop_assert_eq!(x.clone() - y + y, x)
+        }
+
+        #[test]
+        fn add_in_range(ref x in "(0|[1-9][0-9]*)", ref y in "(0|[1-9][0-9]*)") {
+            let x = &Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let y = &Felt::parse_bytes(y.as_bytes(), 10).unwrap();
+            let p = &BigUint::parse_bytes(PRIME_STR[2..].as_bytes(), 16).unwrap();
```
Oppen, line 1259: Ditto.

felt/src/lib.rs
```diff
@@ -1052,6 +1088,204 @@ mod test {
...
+
+            let expected_y = x + y - x;
+            prop_assert_eq!(&expected_y, y, "{}", y);
+        }
+
+        #[test]
+        fn add_assign_in_range(ref x in "(0|[1-9][0-9]*)", ref y in "(0|[1-9][0-9]*)") {
+            let mut x = Felt::parse_bytes(x.as_bytes(), 10).unwrap();
+            let y = Felt::parse_bytes(y.as_bytes(), 10).unwrap();
+            let p = &BigUint::parse_bytes(PRIME_STR[2..].as_bytes(), 16).unwrap();
```
Oppen, line 1279: Ditto.

Oppen (changes requested):
Hello @haroune-mohammedi, thanks for your contribution, it looks good.
However, there are a few changes I need you to make before we can merge it.

felt/src/lib.rs
Oppen: Effectively it still works as I detailed in the first comment.

felt/src/lib.rs
haroune-mohammedi: It won't fail, `is_multiple_of` and `divides` always return `true`, see [`bigint_felt.rs#L615`](https://github.com/Quadratic-Labs/cairo-rs/blob/ef74e80fe4f8f1156233418e3a2c381db5b14047/felt/src/bigint_felt.rs#L615)

felt/src/lib.rs
haroune-mohammedi: Same answer as `is_even`

felt/src/lib.rs
haroune-mohammedi: what should be done about them ?

felt/src/lib.rs
haroune-mohammedi: Why wouldn't it fit, we're subtracting `x` from `x`, we should be left with exactly `y` or am I missing something here ?

felt/src/lib.rs
haroune-mohammedi: it's being parsed in every function, I wanted to get it out or even use `Felt::max_value` instead but I didn't want to change before asking (I asked in the comments about it)

felt/src/lib.rs
Oppen: Sorry, I thought you were comparing with the `BigUint` created from the string, not from the `Felt` itself.

felt/src/lib.rs
Oppen: Ugh, didn't see the `- x` 🤦

felt/src/lib.rs
Oppen: That is odd. `0` shouldn't divide anything.

felt/src/lib.rs
Oppen: I'd say the practical thing is to split each test in two and use a `prop_assume!` to fix the range. One test checks for negatives, one test checks for positives. The `0` case can be made a normal unit test.

haroune-mohammedi: But that's not what is implemented in `Signed`, do you mean we have to change the implementation and make felts where `x > CAIRO_PRIME/2` negative ? What about having a separate `SignedFelt` type, isn't it a better design than messing with the implementation of `Felt` ?

Oppen: Yes, the idea was to eventually refactor into a `SignedFelt` type that you could build from `Felt`. But for now it is that way. It should be documented and tested as that until we do the refactor. We're aware as it is right now it's unintuitive, but we prioritized fixing quickly fast over having it be beautiful, with our sight on refactoring it afterwards.
If you'd rather have it that way __now__ so tests can be more obvious, we can either see if someone on our side has time to do the refactor now (I think everyone has a task currently, so this is possible but unlikely), or maybe we can create an issue and you can assign it to yourself if you wish. The task would be to track where we actually care about sign (it's a few hints IIRC) as well as create the base shim.

Oppen (approved):

PR #859, opened 2023-02-24
Check for overflows in relocatable operations

fmoletta:
This PR does the following:
- Add checks for possible overflows when adding to a relocatable's offset (which can panic if the addition exceeds usize::MAX)
- Move Relocatable operations to trait implementations

Depends on #855

### DIFF
--- a/src/types/relocatable.rs
+++ b/src/types/relocatable.rs
@@ -707,11 +707,13 @@ mod tests {
     }
 
     #[test]
-    #[should_panic]
     fn relocatable_add_i32_with_overflow() {
         let reloc = relocatable!(1, 1);
 
-        let _panic = reloc + (-3);
+        assert_eq!(
+            reloc + (-3),
+            Err(MathError::RelocatableSubNegOffset(relocatable!(1, 1), 3))
+        );
     }
 
     #[test]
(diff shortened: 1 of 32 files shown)


src/types/relocatable.rs
```diff
@@ -88,43 +88,82 @@ impl Display for Relocatable {
...
-        } else {
-            relocatable!(
-                self.segment_index,
-                self.offset - other.unsigned_abs() as usize
-            )
+/// Adds a MaybeRelocatable to self
+/// Cant add two relocatable values
+impl Add<&MaybeRelocatable> for Relocatable {
+    type Output = Result<Relocatable, MathError>;
+    fn add(self, other: &MaybeRelocatable) -> Result<Relocatable, MathError> {
```
Oppen, line 135: Typical (equivalent) signature:
```suggestion
    type Output = Result<Relocatable, MathError>;
    fn add(self, other: &MaybeRelocatable) -> Self::Output {
```

I'm not 100% sure returning a `Result<_>` from an operator is what anybody would expect, as typically they'd use `checked_*` to handle errors and let it `panic!` otherwise. Still, this is infinitely better than what we had, so I'm OK with it.

Oppen (approved):

Jrigada (approved):
Much nicer

PR #860, opened 2023-02-24
Use `Relocatable` instead of `&MaybeRelocatable` in `load_data`

fmoletta:
Change the ptr argument type in `MemorySegmentManager::load_data` to Relocatable


Jrigada (approved):

Oppen (approved):
Loving it!

PR #873, opened 2023-03-02
Update keccak-builtin

fmoletta:
Solves #869


src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -245,10 +253,39 @@ impl KeccakBuiltinRunner {
...
+
+    fn keccak_f(input_message: &[u8]) -> Result<Vec<u8>, RunnerError> {
+        let bigint = BigUint::from_bytes_le(input_message);
+        let w = 64;
+        let keccak_input = (0..25)
+            .map(|i| {
+                ((&bigint >> (i * w)) & (BigUint::from(2_u8).pow(w) - BigUint::one())).to_u64()
+            })
+            .collect::<Option<Vec<u64>>>()
+            .ok_or(RunnerError::KeccakInputCellsNotU64)?;
```
Oppen, line 272:
```suggestion
        let bigint = BigUint::from_bytes_le(input_message);
        let w = 64;
        let mut keccak_input = bigint.to_u64_digits();
        keccak_input.resize(25, 0);
```

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -245,10 +253,39 @@ impl KeccakBuiltinRunner {
...
+        // This unwrap wont fail as keccak_input is created from a (0..25) range
+        let mut keccak_input: [u64; 25] = keccak_input.try_into().unwrap();
+        keccak::f1600(&mut keccak_input);
+        Ok(keccak_input
+            .iter()
+            .enumerate()
+            .map(|(i, x)| BigUint::from(*x) << (i * w as usize))
+            .sum::<BigUint>()
+            .to_bytes_le())
+    }
```
Oppen, line 282:
```suggestion
        Ok(keccak_input
            .iter()
            .flat_map(|x| x.to_le_bytes())
            .collect())
    }
```

Oppen (approved):

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -74,52 +77,57 @@ impl KeccakBuiltinRunner {
...
         }
-
-        let first_input_addr = address
-            .sub_usize(index)
-            .map_err(|_| RunnerError::KeccakNoFirstInput)?;
-
-        if self.verified_addresses.contains(&first_input_addr) {
-            return Ok(None);
+        if let Some(felt) = self.cache.borrow().get(&address) {
+            return Ok(Some(felt.into()));
```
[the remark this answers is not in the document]
fmoletta: Its not redundandt in this case. The keccak takes 8 input values and outputs 8 values. Only one of the 8 output values is returned (when the memory cell is deduced), but all of them are saved to the cache. If we didn't save them in the builtin's cache, we would need to re-compute the keccak for each of the other seven output cells on further calls.

src/vm/runners/builtin_runner/keccak.rs
fmoletta: Added!

src/vm/runners/builtin_runner/keccak.rs
fmoletta: Added!

Jrigada (approved):
LGTM

PR #876, opened 2023-03-06
ec_op changes 0.11 release

Jrigada:
# ec_op changes 0.11 release

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -40,6 +44,7 @@ impl EcOpBuiltinRunner {
             stop_ptr: None,
             included,
             instances_per_component: 1,
+            cache: Rc::new(RefCell::new(HashMap::new())),
```
Oppen, line 47: I think `RefCell<HashMap<_>>` should suffice, we only need the interior mutability, right?

Oppen: Regardless, this seems to duplicate the check that already happens in the call sites of the validation rules. AFAIR it's the memory module that has a `HashMap<Relocatable>` with all the validated addresses to avoid calling the rule more than once.

Oppen (approved):

Issue #770, opened 2023-01-24
Compatibility with no_std


LucasLvy:
**Is your feature request related to a problem? Please describe.**

For now it's not compatible with [no_std](https://docs.rust-embedded.org/book/intro/no-std.html) so it's impossible to run it on bare metal hardware 

**Describe the solution you'd like**

Here is the simple way to do it:
This is a pretty typical large-scale refactoring task. Rust does not have a rich IDE experience (unlike Java) but CLI tools will help you. I'd suggest to:

* add an unconditional #![no_std] to your library. this disables the std prelude, keeping the non-std one. Then, fix all errors by adding use std::... if only present in std or use alloc::... or use core::... if also present there. Yes, the tag just disables the prelude, it does not make std inaccessible from your crate. Do a git commit.

* call rg use std:: or rg std:: and use the output to build a GNU find invocation that calls sed -i on all files with stuff like s/std::boxed/alloc::boxed. So you don't edit any files, but work on the find invocation. If there is a compile error or something, you can just do a git reset --hard to get back to where you started to try an improved replacement command. do another commit.

* The invocation might not have found everything e.g. a use std::{boxed, ... might break the regex and it's hard to do regex based replacement for that. But those cases will be in the minority, unless that was your preferred style, so you can change them manually.

* now you can add an std feature and add #![cfg(...)]s for the stuff that didn't catch. For hashmaps, you can put put a cfg-decided use hashbrown::HashMap/use std::collections::HashMap into lib.rs and then use that via use crate::HashMap. Here also you can use the auto-replacement trick from above.

what I hope is having something automatically choose std::vec and alloc::vec everytime depending on whether the no_std feature is on or off.

For stuff that's in alloc/core, just always use the more portable variant. For hash maps, use libs.rs crate-local reexports.

source: https://www.reddit.com/r/rust/comments/10f3nvn/how_do_you_idiomatically_convert_libs_to_no_std/

**Additional context**

Here is an example on how to make it switchable between no_std and regular rust
https://github.com/DioxusLabs/taffy/blob/main/src/sys.rs

PR #892, opened 2023-03-10
Extend IAI benchmark to include all serializations and proof mode

Oppen:
IAI benchmarks were modeled after the Criterion benchmark. Those failed to consider the serialization times, which led to #770 causing unexpected slowdowns not being caught before merge.
This PR fixes that by including writing memory and trace files (to `/dev/null`) to consider that processing cost as well.
There's still a pending concern about syscalls. I'm not 100% sure Valgrind can take blocking operations into account for the estimation.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


fmoletta (approved):
LGTM

Jrigada (approved):

PR #893, opened 2023-03-10
WASM32 support with regression fix

Oppen:
This reverts commit 23d941fd587354ea6ea82efd397baeed5c4f24bf and applies the following changes on top to fix a performance regression caused by the previously reverted commit:
- Re-add support for mimalloc to the new subproject that creates the runnable;
- Use a larger buffer for writing the memory and execution traces (5MiB and 3MiB respectively, so a whole number of fields fit between flushes);
- Extend IAI benchmarks a bit to consider these writes.

Co-authored-by: @tdelabro


fmoletta (approved):
LGTM

igaray (approved):

juanbono (approved):

tdelabro:
:fire:

PR #896, opened 2023-03-13
feat: impl Deserialize on Program by changing builtins field type to enum

### LINKED ISSUE #895: Derive Deserialize on Program
```rust
#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
pub struct Program {
    pub builtins: Vec<&'static str>,
    ...
}
```

Recently the `builtins` field of the `Program` struct changed from `Vec<String>` to `Vec<&'static str>`. This makes sense because it's always populated with values coming from this function:

```rust
impl BuiltinName {
    pub fn name(&self) -> &'static str {
        match self {
            BuiltinName::output => OUTPUT_BUILTIN_NAME,
            BuiltinName::range_check => RANGE_CHECK_BUILTIN_NAME,
            BuiltinName::pedersen => HASH_BUILTIN_NAME,
            BuiltinName::ecdsa => SIGNATURE_BUILTIN_NAME,
            BuiltinName::keccak => KECCAK_BUILTIN_NAME,
            BuiltinName::bitwise => BITWISE_BUILTIN_NAME,
            BuiltinName::ec_op => EC_OP_BUILTIN_NAME,
        }
    }
}
```

In starknet_in_rust, it is used in the struct `ContractClass` upon which the trait Deserialize is derived. It was not a problem when the builtins were `Vec<String>` but it becomes one when they are references with static lifetime, because if we just derive `Deserialize`, it is expected that the whole object from which we are reading is static too.
We could introduce a named lifetime instead of using `'static`, but it would make it impossible to use on an object that does not outlive the resulting struct, which is already happening in a lot of places:
```rust
trait Storage {
 fn get_value(&self, key: &StorageKey) -> Option<Vec<u8>>;

 fn get_contract_state(&self, key: &[u8; 32]) -> Result<ContractState, StorageError> {
        let ser_contract_state = self
            .get_value(&(Prefix::ContractState, *key))
            .ok_or(StorageError::ErrorFetchingData)?;

        let contract_state: ContractState = serde_json::from_slice(&ser_contract_state)?;
        Ok(contract_state)
    }
}
```

We can solve this in different ways.

The first one is to use a custom function to deserialize this field rather that the default one shipped by the trait derivation. It may look like this:
```rust
#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct Program {
    #[serde(deserialize_with = "deserialize_builtins")]
    pub builtins: Vec<&'static str>,
    ...
}


fn deserialize_builtins<'de, D>(data: D) -> Result<Vec<&'static str>, D::Error>
where
    D: Deserializer<'de>,
{
    struct BuiltinsVisitor;

    impl<'de> Visitor<'de> for BuiltinsVisitor {
        type Value = Vec<&'static str>;

        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
            formatter.write_str("a string containing json data")
        }

        fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
        where
            E: serde::de::Error,
        {
            let builtins =
                serde_json::from_str(v)
                    .map_err(E::custom)
                    .map(|builtins: Vec<String>| {
                        builtins
                            .iter()
                            .map(|builtin| match builtin.as_str() {
                                OUTPUT_BUILTIN_NAME => Ok(OUTPUT_BUILTIN_NAME),
                                RANGE_CHECK_BUILTIN_NAME => Ok(RANGE_CHECK_BUILTIN_NAME),
                                HASH_BUILTIN_NAME => Ok(HASH_BUILTIN_NAME),
                                SIGNATURE_BUILTIN_NAME => Ok(SIGNATURE_BUILTIN_NAME),
                                KECCAK_BUILTIN_NAME => Ok(KECCAK_BUILTIN_NAME),
                                BITWISE_BUILTIN_NAME => Ok(BITWISE_BUILTIN_NAME),
                                EC_OP_BUILTIN_NAME => Ok(EC_OP_BUILTIN_NAME),
                                _ => return Err(E::custom("unknown builtin")),
                            })
                            .collect::<Result<Vec<&'static str>, E>>()
                    })??;

            Ok(builtins)
        }
    }

    data.deserialize_any(BuiltinsVisitor)
}
```    
It should work, but look rather silly. Not gonna lie.

Another way would be to use an enum like `BuiltinName`, rather than a raw string. It will have better performance, be a more expressive type and solve our lifetime problems.
Here is a commit that does more or less this on my fork: https://github.com/open-dust/cairo-rs/commit/4f1a2096f4d6001ad1dd6ab85f4ad4dc65c233df
The only thing I'm not sure is about supporting other custom built-ins. Is it something that is needed? I'm under the impression it is needed because there is an `add_additional_hash_builtin` in the codebase, which does exactly that. My opinion is that we can still make it possible by adding an escape variant to the enum, something like `Other(String)`. 

In the same vein, we can just revert to `Vec<String>`, but I think having an enum brings additional value.

I hope it gave you a clear view of what is the problem, and that my ideas of solutions can help. I will be glad to help further, maybe making the PR once we agreed on the best way to do this.

tdelabro:
close #895

### DIFF
--- a/src/types/program.rs
+++ b/src/types/program.rs
@@ -1,21 +1,21 @@
 use crate::{
     serde::deserialize_program::{
-        deserialize_program, Attribute, HintParams, Identifier, InstructionLocation,
+        deserialize_program, Attribute, BuiltinName, HintParams, Identifier, InstructionLocation,
         ReferenceManager,
     },
     types::{errors::program_errors::ProgramError, relocatable::MaybeRelocatable},
 };
 use felt::{Felt, PRIME_STR};
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use std::{
     fs::File,
     io::{BufReader, Read},
     {collections::HashMap, path::Path},
 };
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Program {
-    pub builtins: Vec<&'static str>,
+    pub builtins: Vec<BuiltinName>,
     pub prime: String,
     pub data: Vec<MaybeRelocatable>,
     pub constants: HashMap<String, Felt>,
@@ -33,7 +33,7 @@ pub struct Program {
 impl Program {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
-        builtins: Vec<&'static str>,
+        builtins: Vec<BuiltinName>,
         prime: String,
         data: Vec<MaybeRelocatable>,
         main: Option<usize>,
@@ -121,7 +121,7 @@ mod tests {
             references: Vec::new(),
         };
 
-        let builtins: Vec<&'static str> = Vec::new();
+        let builtins: Vec<BuiltinName> = Vec::new();
         let data: Vec<MaybeRelocatable> = vec![
             mayberelocatable!(5189976364521848832),
             mayberelocatable!(1000),
@@ -156,7 +156,7 @@ mod tests {
             references: Vec::new(),
         };
 
-        let builtins: Vec<&'static str> = Vec::new();
+        let builtins: Vec<BuiltinName> = Vec::new();
 
         let data: Vec<MaybeRelocatable> = vec![
             mayberelocatable!(5189976364521848832),
@@ -225,7 +225,7 @@ mod tests {
             references: Vec::new(),
         };
 
-        let builtins: Vec<&'static str> = Vec::new();
+        let builtins: Vec<BuiltinName> = Vec::new();
 
         let data: Vec<MaybeRelocatable> = vec![
             mayberelocatable!(5189976364521848832),
@@ -285,7 +285,7 @@ mod tests {
         )
         .expect("Failed to deserialize program");
 
-        let builtins: Vec<&'static str> = Vec::new();
+        let builtins: Vec<BuiltinName> = Vec::new();
         let data: Vec<MaybeRelocatable> = vec![
             mayberelocatable!(5189976364521848832),
             mayberelocatable!(1000),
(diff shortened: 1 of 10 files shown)


tdelabro:
@fmoletta I think this is what you described. wdyt?

src/types/program.rs
```diff
@@ -1,21 +1,21 @@
...
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use std::{
     fs::File,
     io::{BufReader, Read},
     {collections::HashMap, path::Path},
 };
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
fmoletta, line 16: Is deriving Deserialize necessary here?

fmoletta (approved):
LGTM, left a small comment

src/types/program.rs
tdelabro: Well, it's the whole reason this problem was raised.
`Program` is used in starknet_in_rust inside the struct `ContractClass`, which does implement `Deserialize`. So it has to implement it too.
It has been removed in cairo-rs when you changed `Vec<String>` for `Vec<&'static str>`, probably because some errors arise, but I will be needed when starknet_in_rust upgrade its dependency to the last version of cairo-rs

fmoletta: Oh, my mistake

Oppen (commented):
Minor comment, but otherwise LGTM.

src/vm/runners/cairo_runner.rs
```diff
@@ -183,57 +180,61 @@ impl CairoRunner {
...
-        let program_builtins: HashSet<&&str> =
-            self.program.builtins.iter().collect::<HashSet<&&str>>();
+            .map(|x| x.0)
+            .collect::<HashSet<&str>>();
+        let program_builtins = self
+            .program
+            .builtins
+            .iter()
+            .map(|builtin_name| builtin_name.name())
+            .collect::<HashSet<&str>>();
```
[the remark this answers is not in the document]
tdelabro: we are calling
` !program_builtins.is_subset(&inserted_builtins)` so both hashset should have the same key, and `inserted_builtins` type is be &'static str

Oppen (approved):

PR #897, opened 2023-03-14
Use mimalloc in benchmarks

Oppen:
Benchmarks (criterion and iai) were running with mimalloc disabled. This should make them run faster.


fmoletta (approved):
lgtm

PR #906, opened 2023-03-16
Optimize trace relocation

fmoletta:
This PR does the following:

- Use only register offsets in `TraceEntry ` struct
- Simplify `CairoRunner::relocate_trace`
- Overwrite the trace with its relocated version instead of creating a separate struct
- Move `relocate_trace` method to VM
- Add method `VM::get_relocated_trace`
These changes are valid as pc, ap, and fp segment indexes will always be 0,1,1 respectively (the last pc can have a different index but it wont be written to the trace)


Oppen (approved):
Simply beautiful ❤️

src/vm/runners/cairo_runner.rs
```diff
@@ -756,23 +756,27 @@ impl CairoRunner {
...
+                .iter()
+                .map(|entry| RelocatedTraceEntry {
+                    pc: entry.pc_off + 1,
+                    ap: entry.ap_off + segment_1_base,
+                    fp: entry.fp_off + segment_1_base,
+                })
+                .collect(),
+        );
         Ok(())
     }
```
Oppen, line 781: Now that it's `[usize]` we could probably do the relocation in-place.

fmoletta: Added in-place relocation [524413d)

src/vm/runners/builtin_runner/keccak.rs
```diff
@@ -376,36 +376,36 @@ mod tests {
...
-        let mut vm = vm!();
+    //     let mut vm = vm!();
 
-        vm.segments.segment_used_sizes = Some(vec![0]);
-        let program = Program::from_bytes(
-            include_bytes!("../../../../cairo_programs/_keccak.json"),
-            Some("main"),
-        )
-        .unwrap();
+    //     vm.segments.segment_used_sizes = Some(vec![0]);
```
juanbono, line 387: Why this test is commented?

fmoletta: Oh, my mistake, Thanks for catching it!

PR #909, opened 2023-03-17
Optimize memory comparison for set hints

Oppen:
Implements a new `memcmp` function in `Memory` for comparing two ranges. This allows linear searches and the like for arbitrarily big structures without needing to copy a range of memory.
The improvement is 2.20x for set_integration_benchmark.


juanbono (approved):

juanbono:
LGTM but check codecov suggestions

fmoletta (approved):
Great!

PR #910, opened 2023-03-17
Only relocate memory if `memory_file` is passed

fmoletta:
This PR aims to avoid relocating the cairo memory if its not going to be outputted by the vm. If using the `cairo_run`, this can be customized via the `CairoRunConfig`

### DIFF
--- a/src/vm/runners/cairo_runner.rs
+++ b/src/vm/runners/cairo_runner.rs
@@ -776,16 +776,25 @@ impl CairoRunner {
         Ok(())
     }
 
-    pub fn relocate(&mut self, vm: &mut VirtualMachine) -> Result<(), TraceError> {
+    pub fn relocate(
+        &mut self,
+        vm: &mut VirtualMachine,
+        relocate_mem: bool,
+    ) -> Result<(), TraceError> {
         vm.segments.compute_effective_sizes();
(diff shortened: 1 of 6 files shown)


src/vm/runners/cairo_runner.rs
```diff
@@ -776,16 +776,25 @@ impl CairoRunner {
         Ok(())
     }
 
-    pub fn relocate(&mut self, vm: &mut VirtualMachine) -> Result<(), TraceError> {
+    pub fn relocate(
+        &mut self,
+        vm: &mut VirtualMachine,
+        relocate_mem: bool,
```
Oppen, line 782: It's probably better to just check the condition before calling this.

fmoletta: It also relocates the trace, thats why

Oppen: Would it be too bad to split those?

fmoletta: We can do it, but it would mean replacing one function call with 3 (compute_effective_sizes - relocate_trace - relocate-memory), and making the user responsible for calling compute_effective_sizes (which will cause the other two to fail if not called)

Oppen (approved):

PR #879, opened 2023-03-07
Add dynamic layout

fmoletta:
Solves #878

PR #913, opened 2023-03-20
[0.11] Add `SegmentArenaBuiltinRunner`

fmoletta:
Depends on #879


src/types/instance_definitions/bitwise_instance_def.rs
```diff
@@ -3,19 +3,19 @@ pub(crate) const INPUT_CELLS_PER_BITWISE: u32 = 2;
 
 #[derive(Clone, Debug, PartialEq)]
 pub(crate) struct BitwiseInstanceDef {
-    pub(crate) ratio: u32,
+    pub(crate) ratio: Option<u32>,
```
Oppen, line 6: I leave the choice to you because it's a compromise in ergonomics, but it may be a good idea to make this `Option<NonZeroU32>` instead. The same would be for the constructor.

Oppen (approved):

juanbono (approved):
LGTM!

juanbono (changes requested):
Please fix build

PR #914, opened 2023-03-20
Implement AddAssign for ExecutionResources and fix union

Oppen:
Support the `+=` operator for `ExecutionResources`.
It also avoids requiring ownership in `Add` and fixes a bug where it would take the intersection rather than the union.


fmoletta (approved):
LGTM

juanbono (changes requested):
can you also change it for the rest of the operations? (Mul, Add, Sub)

juanbono (approved):
LGTM. It would be nice to avoid moving self on all the operations

PR #916, opened 2023-03-21
Several fixes to `ExecutionResources` operators

Oppen:
- Implements `SubAssign` and `MulAssign`;
- Uses union rather than intersection for `Sub`;
- Uses references rather than ownership transfer.


fmoletta (approved):
LGTM

PR #917, opened 2023-03-22
Use `$crate` instead of name literal on `felt_str` macro

MegaRedHand:
# Use `$crate` instead of name literal on `felt_str` macro

## Description

When using a name different than `felt` on crate import, we can't use the `felt_str` macro without an additional `use felt;` statement.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


fmoletta (approved):
LGTM

PR #918, opened 2023-03-22
Add requirements.txt

fmoletta:
Notes: iai benchmark workflow breaks as there is no requirements.txt in last main commit

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -26,7 +26,7 @@ jobs:
       with:
         python-version: '3.9'
     - name: Install test dependencies
-      run: make python-requirements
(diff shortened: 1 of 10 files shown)


Oppen (approved):
Excellent!

Oppen:
> Notes: iai benchmark workflow breaks as there is no requirements.txt in last main commit

I think a fix for this is to delay one merge the change for the base branch. This way, it will build now (using the hardcoded version as is for main), then in a different PR we start using the requirements.txt and because it already exists in main it doesn't break.

Oppen (approved):

juanbono (approved):

PR #919, opened 2023-03-23
Fix cairo constants being used in hints that import python constants

fmoletta:
Some hints from cairo_scept import python constants, and have intead been using cairo constants


Oppen (approved):
Neat!

PR #921, opened 2023-03-23
Remove builtin names from `VirtualMachine.builtin_runners`

fmoletta:
Motive: Simplify code & have a cleaner/simpler interface
Reasoning: builtin runners are now enums instead of trait objectst, so we can know exectly which builtin we are using without needing to see its name. Morover, builtins know their name, we can just call builtin.name() if we need it.
Possible consecuences: The method `add_additional_hash_builtin `used to store the new builtin under a different name ("hash_builtin" instead of "pedersen"), but this wont longer be needed after 0.11 changes
Other improvements: Simplified code catching NoBuiltinForLayout error
Performance: No difference


Oppen (approved):
Nice!

PR #929, opened 2023-03-29
Augment variation on proptest inputs

Oppen:
@martinacantaro noticed we were having `48` as the `Felt252` value in many proptests, but I merged too quickly.
The reasons were two: first, a bug where I was using `from_bytes_be` rather than `parse_bytes` to create them (so `"0"` was `48`) and the pattern being an alternation of `0` and any number not beginning with `0` led to about 50% of the time the generated input being `0`.
The fix here is to correct the parsing and to special case the `0` in separate tests and exclude it from the pattern.

### DIFF
--- a/felt/src/lib.rs
+++ b/felt/src/lib.rs
@@ -897,8 +897,8 @@ mod test {
 
         #[test]
         #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
-        fn to_be_bytes(ref x in FELT_PATTERN) {
-            let x = &Felt252::from_bytes_be(x.as_bytes());
+        fn to_be_bytes(ref x in "[1-9][0-9]*") {
+            let x = &Felt252::parse_bytes(x.as_bytes(), 10);
             let bytes = x.to_be_bytes();
             let y = &Felt252::from_bytes_be(&bytes);
             prop_assert!(x == y);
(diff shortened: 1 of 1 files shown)


felt/src/lib.rs
```diff
@@ -897,17 +897,17 @@ mod test {
 
         #[test]
         #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
-        fn to_be_bytes(ref x in FELT_PATTERN) {
-            let x = &Felt252::from_bytes_be(x.as_bytes());
+        fn to_be_bytes(ref x in "[1-9][0-9]*") {
```
fmoletta, line 900:
```suggestion
        fn to_be_bytes(ref x in FELT_NON_ZERO_PATTERN) {
```

felt/src/lib.rs
```diff
@@ -897,17 +897,17 @@ mod test {
...
             let bytes = x.to_be_bytes();
             let y = &Felt252::from_bytes_be(&bytes);
             prop_assert!(x == y);
         }
 
         #[test]
         #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
-        fn to_le_bytes(ref x in FELT_PATTERN) {
-            let x = &Felt252::from_bytes_be(x.as_bytes());
+        fn to_le_bytes(ref x in "[1-9][0-9]*") {
```
fmoletta, line 909:
```suggestion
        fn to_le_bytes(ref x in "FELT_NON_ZERO_PATTERN") {
```

fmoletta (approved):
Nice catch! left minor comments

felt/src/lib.rs
Oppen: Nice catch! I changed the strategy anyway tho.

juanbono (approved):

PR #942, opened 2023-04-03
perf!: avoid extra copies of immediate args

Oppen:
The `Instruction` structure no longer keeps a copy of whatever value follows the tentative instruction. If it's an immediate, other fields are enough to deduce that. If the argument is missing, memory access will make that evident.
There's no need for that copy.

Because a public field of a public structure is gone, it's a breaking change.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added


fmoletta (approved):
LGTM!

pefontana (approved):

pefontana:
nice one!

PR #947, opened 2023-04-03
Add Poseidon benchmark

fmoletta:
Add benchmark for poseidon (builtin)


Oppen (commented):
Looks neat! You should rebase against main.

bench/criterion_benchmark.rs
```diff
@@ -29,6 +29,7 @@ const BENCH_NAMES: &[&str] = &[
     "operations_with_data_structures_benchmarks",
     "uint256_integration_benchmark",
     "set_integration_benchmark",
+    "poseidon_integration_benchmark",
```
Oppen, line 32: You should add it to `iai_benchmark.rs`.

Oppen (changes requested):

bench/criterion_benchmark.rs
fmoletta: Nice catch! Added!

Oppen (approved):

juanbono (approved):

PR #955, opened 2023-04-10
Keep `Arc<Program>` in `CairoRunner`

Oppen:
Rather than owning, the `Program` is typically read only, so this allows caching it and avoids deep copying of a big object. This copy turned out to be a bottleneck when using the crate for executing a single program many times.
The change needs no new tests as it's mostly internal. It does change the signature of `CairoRunner::new()` to receive the `Program`.


Oppen:
We'll use a different approach.

PR #963, opened 2023-04-12
Update make deps

pefontana:
# Update make deps

## Description
* Update make deps
* Add make deps-macos

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,12 +94,21 @@ deps:
 	cargo install --version 1.14.0 hyperfine
 	cargo install --version 0.9.49 cargo-nextest
 	cargo install --version 0.5.9 cargo-llvm-cov
-	pyenv install pypy3.7-7.3.9
-	pyenv global pypy3.7-7.3.9
-	pip install -r requirements.txt
-	pyenv install 3.7.12
-	pyenv global 3.7.12
-	pip install -r requirements.txt
+	cargo install --version 0.11.0 wasm-pack
+	pyenv install 3.9.15
+	pyenv local 3.9.15	pip install -r requirements.txt
+
+deps-macos:
+	cargo install --version 1.1.0 cargo-criterion
+	cargo install --version 0.6.1 flamegraph
+	cargo install --version 1.14.0 hyperfine
+	cargo install --version 0.9.49 cargo-nextest
+	cargo install --version 0.5.9 cargo-llvm-cov
+	cargo install --version 0.11.0 wasm-pack
+	brew install gmp
+	pyenv install 3.9.15
+	pyenv local 3.9.15
+	CFLAGS=-I/opt/homebrew/opt/gmp/include LDFLAGS=-L/opt/homebrew/opt/gmp/lib pip install -r requirements.txt
 
 $(RELBIN):
 	cargo build --release
@@ -182,4 +191,3 @@ clean:
 	rm -f $(TEST_PROOF_DIR)/*.json
 	rm -f $(TEST_PROOF_DIR)/*.memory
 	rm -f $(TEST_PROOF_DIR)/*.trace
-
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 **/*.memory
 **/*.swp
 bench/results
+.python-version


Makefile
```diff
@@ -94,13 +94,29 @@ deps:
...
-	pyenv install pypy3.7-7.3.9
-	pyenv global pypy3.7-7.3.9
+	cargo install --version 0.11.0 wasm-pack
+	pyenv install pypy3.9-7.3.9
+	pyenv local pypy3.9-7.3.9
 	pip install -r requirements.txt
-	pyenv install 3.7.12
-	pyenv global 3.7.12
+	pyenv install 3.9.15
+	pyenv local 3.9.15
```
Oppen, line 102: It's better to use the `PYENV_VERSION` env var instead of changing local each time.

igaray:
Small comment, but the README in the Getting Started section has pyenv in the Optional section, saying "These dependencies are only necessary in order to run the original VM and compile Cairo programs.". It should clarify that pyenv is required to run the tests. 

Another nitpick is if some downloads the repo and runs `make`, it defaults to installing the deps and errors when pyenv is not found. The `deps` target should be separated into the base "running dependencies" which might be required by a plain user and "development/testing" dependencies which do in fact require python and pyenv.

Makefile
```diff
@@ -94,13 +94,29 @@ deps:
...
 	cargo install --version 0.5.9 cargo-llvm-cov
-	pyenv install pypy3.7-7.3.9
-	pyenv global pypy3.7-7.3.9
+	cargo install --version 0.11.0 wasm-pack
+	pyenv install pypy3.9-7.3.9
+	pyenv local pypy3.9-7.3.9
 	pip install -r requirements.txt
-	pyenv install 3.7.12
-	pyenv global 3.7.12
+	pyenv install 3.9.15
```
MegaRedHand, line 101: We can use the `-s` option for `pyenv install`, so it skips installation if the version is already installed. Right now it prompts the user for confirmation.

pefontana: Thanks Mega!
Didn't know that

Makefile
pefontana: I added the cairo-rs-env and cairo-rs-pypy-env
[0b18463](https://github.com/lambdaclass/cairo-rs/pull/963/commits/0b1846363793f5a449fc99e7be20f35901051488)

pefontana:
> Small comment, but the README in the Getting Started section has pyenv in the Optional section, saying "These dependencies are only necessary in order to run the original VM and compile Cairo programs.". It should clarify that pyenv is required to run the tests.
> 
> Another nitpick is if some downloads the repo and runs `make`, it defaults to installing the deps and errors when pyenv is not found. The `deps` target should be separated into the base "running dependencies" which might be required by a plain user and "development/testing" dependencies which do in fact require python and pyenv.

Thanks for the comment @igaray 

I update the README.md so users know that they need pyenv to run the test.
Regarding the second point, the only thing a plain user can do without pyenv is to compile the project. If the user wants to do anything more it will need the dependencies in `make deps`, so it thinks it doesn't make sense to add an empty `make deps` for plain users.

Maybe instead it will be better to add the `make deps` step in the README.md 

What do you think?

Makefile
```diff
@@ -94,12 +94,32 @@ deps:
...
+	pyenv install pypy3.9-7.3.9
+	PYENV_VERSION=pypy3.9-7.3.9 python -m venv cairo-rs-pypy-env
+	. cairo-rs-pypy-env/bin/activate ; \
+	pip install -r requirements.txt ; \
+	pyenv install 3.9.15
+	PYENV_VERSION=3.9.15 python -m venv cairo-rs-env
+	. cairo-rs-env/bin/activate ; \
+	pip install -r requirements.txt ; \
+
+deps-macos:
```
MegaRedHand, line 107: I think this can be simplified to:
```make
deps-macos:
    brew install gmp
    CFLAGS=-I/opt/homebrew/opt/gmp/include LDFLAGS=-L/opt/homebrew/opt/gmp/lib $(MAKE) deps
```

juanbono (approved):

Makefile
pefontana: I tried with this one and couldn't pass a flag to a make command, plus the pypy installation is different from the `make deps` one.
Aldo, I created a make cargo-deps to make the commands cleaner.

Oppen (approved):

PR #964, opened 2023-04-12
test: unit tests for cairo-vm-cli

Oppen:
This splits functionality into a `run` function that receives the iterator of arguments, and `main` that simply calls it with `std::env::args`.
Then, adds several test cases with `rstest`. It checks all combinations of successful runs and a few of invalid arguments and missing files. This should make the codecov path pipeline happier :)

NOTE: the dynamic layout is too slow, and it caused a panic (subsequently fixed) as well. It is commented out from the tests until optimized.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/cairo-vm-cli/src/main.rs
+++ b/cairo-vm-cli/src/main.rs
@@ -56,6 +56,8 @@ fn validate_layout(value: &str) -> Result<(), String> {
 
 #[derive(Debug, Error)]
 enum Error {
+    #[error("Invalid arguments")]
+    CLI(#[from] clap::Error),
     #[error("Failed to interact with the file system")]
     IO(#[from] std::io::Error),
     #[error("The cairo program execution failed")]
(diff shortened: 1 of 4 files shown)


cairo-vm-cli/src/main.rs
```diff
@@ -101,8 +103,15 @@ impl FileWriter {
...
 }
 
-fn main() -> Result<(), Error> {
-    let args = Args::parse();
+fn run(args: impl Iterator<Item = String>) -> Result<(), Error> {
+    let args = Args::try_parse_from(args);
+    let args = match args {
+        Ok(args) => args,
+        Err(error) => {
+            println!("{error}");
```
MegaRedHand, line 111: nit: I think this should print to stderr

Oppen:
```suggestion
            eprintln!("{error}");
```

pefontana (approved):

PR #983, opened 2023-04-13
feat: Implement hint on uint384_extension lib

fmoletta:
Depends on #960 
Implements the following hint: 
```
           def split(num: int, num_bits_shift: int, length: int):
               a = []
               for _ in range(length):
                   a.append( num & ((1 << num_bits_shift) - 1) )
                   num = num >> num_bits_shift
               return tuple(a)

           def pack(z, num_bits_shift: int) -> int:
               limbs = (z.d0, z.d1, z.d2)
               return sum(limb << (num_bits_shift * i) for i, limb in enumerate(limbs))

           def pack_extended(z, num_bits_shift: int) -> int:
               limbs = (z.d0, z.d1, z.d2, z.d3, z.d4, z.d5)
               return sum(limb << (num_bits_shift * i) for i, limb in enumerate(limbs))

           a = pack_extended(ids.a, num_bits_shift = 128)
           div = pack(ids.div, num_bits_shift = 128)

           quotient, remainder = divmod(a, div)

           quotient_split = split(quotient, num_bits_shift=128, length=6)

           ids.quotient.d0 = quotient_split[0]
           ids.quotient.d1 = quotient_split[1]
           ids.quotient.d2 = quotient_split[2]
           ids.quotient.d3 = quotient_split[3]
           ids.quotient.d4 = quotient_split[4]
           ids.quotient.d5 = quotient_split[5]

           remainder_split = split(remainder, num_bits_shift=128, length=3)
           ids.remainder.d0 = remainder_split[0]
           ids.remainder.d1 = remainder_split[1]
           ids.remainder.d2 = remainder_split[2]
```

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added
- [x] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.
  - [ ] CHANGELOG has been updated.

PR #985, opened 2023-04-13
feat: Implement hints on field_arithmetic lib

fmoletta:
Depends on #983 

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added


src/hint_processor/builtin_hint_processor/field_arithmetic.rs
```diff
@@ -0,0 +1,271 @@
...
+    let gx = generator * &x;
+    let success_gx = is_quad_residue(&gx, &p)?;
+
+    let root_gx = if success_gx {
+        sqrt_prime_power(&gx, &p).unwrap_or_default()
+    } else {
+        BigUint::zero()
+    };
+
+    if !&x.is_zero() && !(success_x as u8 + success_gx as u8).is_one() {
```
Oppen, line 92: I'm not sure this will behave exactly as the Python code. At this point, in the original code `x` could be `None`, which doesn't compare as equal to `0`. In the hint `x` is converted to `0` when originally `None` after the check.

Oppen: It would help if you could add an integration test for this to ensure that case fails or succeeds the same way (that is, you need `x` that isn't a quad residue (so it evaluates to `None` in Python) and `gx` that isn't either. In Python, I think, the check will fail (`x != 0` means we evaluate the `assert`, and because none of the values were quad residues the sum will be `0` and the `assert` fails), while in our code this `None` became `0`, so `!&x.is_zero()` evaluate to `false` and we don't reach the `assert`.

src/math_utils.rs
```diff
@@ -182,6 +185,176 @@ pub fn sqrt(n: &Felt252) -> Felt252 {
     }
 }
 
+// Adapted from sympy _sqrt_prime_power with k == 1
+pub fn sqrt_prime_power(a: &BigUint, p: &BigUint) -> Option<BigUint> {
+    let two = BigUint::from(2_u32);
+    let a = a.mod_floor(p);
```
Oppen, line 191: We should check `!p.is_zero()` before this.

src/math_utils.rs
```diff
@@ -182,6 +185,176 @@ pub fn sqrt(n: &Felt252) -> Felt252 {
...
+        return n.clone();
+    }
+    let s = trailing(prime - 1_u32);
+    let t = prime >> s;
+    let a = n.modpow(&t, prime);
+    #[cfg(not(feature = "std"))]
+    // Rng is not critical here so its safe to use a seeded value
+    let mut rng = SmallRng::seed_from_u64(11480028852697973135);
+    #[cfg(feature = "std")]
+    let mut rng = rand::thread_rng();
```
Oppen, line 234: I think it would be ok to just use the `SmallRng` for both. Just to simplify.

src/math_utils.rs
```diff
@@ -182,6 +185,176 @@ pub fn sqrt(n: &Felt252) -> Felt252 {
...
+    Some(sqrt_tonelli_shanks(&a, p))
+}
+
+fn sqrt_tonelli_shanks(n: &BigUint, prime: &BigUint) -> BigUint {
+    // Based on Tonelli-Shanks' algorithm for finding square roots
+    // and sympy's library implementation of said algorithm.
+    if n.is_zero() || n.is_one() {
+        return n.clone();
+    }
+    let s = trailing(prime - 1_u32);
```
Oppen, line 227:
```suggestion
    let s = (prime - 1_u32).trailing_zeros();
```
https://docs.rs/num-bigint/0.4.3/num_bigint/struct.BigUint.html#method.trailing_zeros

src/math_utils.rs
```diff
@@ -182,6 +185,176 @@ pub fn sqrt(n: &Felt252) -> Felt252 {
...
+    while (&n & BigUint::one()).is_zero() {
+        while (&n & ((BigUint::one() << p) - 1_u32)).is_zero() {
+            n >>= p;
+            t += p;
+            p *= 2;
+        }
+        p = num_integer::Integer::div_floor(&p, &2);
+    }
+    t
+}
```
Oppen, line 339: Let's just use the one in `BigUint`: https://docs.rs/num-bigint/0.4.3/num_bigint/struct.BigUint.html#method.trailing_zeros

src/math_utils.rs
Oppen, line 1: Could we add some simple IAI and Criterion benchmarks for the new functions?

Oppen (changes requested):
This is mostly ok, but I need some points to be discussed before formal approval.

src/math_utils.rs
```diff
@@ -638,6 +901,35 @@ mod tests {
...
+                 let x_sq = x * x;
+                 let sqrt = sqrt_prime_power(&x_sq, &CAIRO_PRIME).unwrap_or_default();
+
+                if &sqrt != x {
+                    assert_eq!(p - sqrt, *x);
+                } else {
+                    assert_eq!(&sqrt, x);
+                }
+            }
+
```
Oppen, line 932: nit: there seems to be an extra level of indentation compared to the other tests.

src/hint_processor/builtin_hint_processor/field_arithmetic.rs
Oppen: Ignore this, `root_x` is what could be `None`, not `x`.

src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
```diff
@@ -514,6 +516,9 @@ impl HintProcessor for BuiltinHintProcessor {
             hint_code::UNSIGNED_DIV_REM_UINT768_BY_UINT384 => {
                 unsigned_div_rem_uint768_by_uint384(vm, &hint_data.ids_data, &hint_data.ap_tracking)
             }
+            hint_code::GET_SQUARE_ROOT => {
+                get_square_root(vm, &hint_data.ids_data, &hint_data.ap_tracking)
```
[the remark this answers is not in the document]
fmoletta: I took the name from the lib's function name to make it easier to find if the integration test fails

src/math_utils.rs
Oppen: Alternatively, I think it would be safe, simpler and faster to go with this:
```suggestion
/* Computed from:
small_trailing = [0] * 256
for j in range(1,8):
    small_trailing[1<<j::1<<(j+1)] = [j] * (1<<(7-j))
*/
const SMALL_TRAILING: [u64; 256] = [
    0, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
    5, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
    6, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
    5, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
    7, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
    5, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
    6, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
    5, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 4, 0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0,
];
// Ported from sympy implementation
fn trailing(n: BigUint) -> u64 {
    let mut n_bytes_iter = n.to_bytes_le().iter();
    let zeros = n_bytes_iter.take_while(|x| x == 0).count().unwrap_or(0) as u64;
    let Some(non_zero) = n_bytes_iter.next() else {
        return zeros;
    };
    zeros + SMALL_TRAILING[non_zero];
}
```

src/math_utils.rs
fmoletta: Changed!

src/math_utils.rs
fmoletta: Yes, the proptest macro messes up cargo-fmt, Ill fix it

Oppen: That explains why it happened to Timothee back then.

Oppen (approved):

src/math_utils.rs
fmoletta: Added a benchmark for the `get_square_root` func

Oppen: I meant a microbenchmark for the individual functions.

Cargo.toml
```diff
@@ -80,6 +80,9 @@ felt = { package = "cairo-felt", path = "./felt", version = "0.3.0-rc1", default
 ] }
 bitvec = { version = "1", default-features = false, features = ["alloc"] }
 
+[dev-dependencies]
+proptest = "1.0.0"
```
Oppen, line 84: I'm not sure `wasm32` actually supported this. I don't know why the dep was there.

fmoletta: I saw that the felt crate had it as a dev-dependency, and used proptest unconditionally, while in cairo-vm we had the `#[cfg(not(target_arch = "wasm32"))]` before proptest, so I tried to make it work for cairo-vm like it does for felt. But wasm tests are now failling so Ill probably revert the change and see if I can fix it in a later PR

Oppen: Yes, that's why I'm asking. I remember seeing some `cfg(not(...))` around proptest at some point. But if it works all the better.

src/math_utils.rs
Oppen: It can be done separately if you consider that better, but I see four new functions in math_utils that are complex and common enough to want to keep track of their performance.
Note it's not a blocker.

Oppen (approved):

src/math_utils.rs
fmoletta: Oh, the ones inside the vm code, They are all used by the `sqrt_mod_prime_power` which is only used by the get_square_root hint so they are covered by the benchmark, but Its a good idea to add a benchmark for it. I think as it would be the first benchmark that is not for a cairo_program it is better to do it in a separate PR.

MegaRedHand (approved):
LGTM

src/hint_processor/builtin_hint_processor/field_arithmetic.rs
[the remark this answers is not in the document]
Oppen: Nice one, I missed it.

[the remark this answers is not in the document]
Oppen: @fmoletta do you feel like changing this? Otherwise you're free to merge 🎉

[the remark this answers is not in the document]
fmoletta: Nice!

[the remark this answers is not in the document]
fmoletta: We could also use and and get rid of the !

PR #999, opened 2023-04-17
perf!: refactor `Program` to reduce `clone` time

Oppen:
There was a report that `CairoRunner::new` took a significant time (~6%) of the runtime in a program that reuses the same `Program` structure to execute it many times for short runtimes. This is significant and couldn't be caught before by the regular workflow of `cairo-vm-cli` (it runs a program once, so it can only be caught if awfully slow).
This change should help any kind of sequencers, so I think it's worth it.

The fields that were extracted are either used only on initialization or on error paths. The rest are still being copied, as a previous approach of passing `Arc<Program>` instead of `Program` to `CairoRunner::new` led to a big performance hit.
The change should non-breaking because it's kept internal to `Program`: there's a new `pub(crate)` field that keeps an `Arc<SharedProgramData>`, so no public function changes. However, we did some oversharing of internal data in previous releases and now it's technically breaking: the fields we moved are no longer visible, and wouldn't be accessible with the same paths anyway.

Two other approaches we tried first:
- Store a reference: this forced an explicit lifetime in `CairoRunner` which broke in an unfixable way `cairo-rs-py`: `PyO3` can't deal with structs having their lifetime managed by Rust.
- Directly store an `Arc<Program>` in `CairoRunner`: as mentioned, this made execution much slower (10-20%).

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests ~~added~~ updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 #### Upcoming Changes
 
+* BREAKING CHANGE: refactor `Program` to optimize `Program::clone` [#999](https://github.com/lambdaclass/cairo-rs/pull/999)
+    * Breaking change: many fields that were (unnecessarily) public become hidden by the refactor.
+
 * BREAKING CHANGE: Add _builtin suffix to builtin names e.g.: output -> output_builtin [#1005](https://github.com/lambdaclass/cairo-rs/pull/1005)
 
 * Implement hint on uint384_extension lib [#983](https://github.com/lambdaclass/cairo-rs/pull/983)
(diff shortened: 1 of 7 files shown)


Oppen:
No slowdown above uncertainty was detected, so this solution is acceptable in principle.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 #### Upcoming Changes
 
+<<<<<<< HEAD
```
pefontana, line 5: jeje

pefontana (approved):

PR #1010, opened 2023-04-18
feat(hints): add NewHint#32

MegaRedHand:
## Description

This PR adds the missing hint _NewHint#32_, now known as `IS_ZERO_INT`.


src/hint_processor/builtin_hint_processor/secp/field_utils.rs
```diff
@@ -432,107 +432,114 @@ mod tests {
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn run_is_zero_nondet_ok_true() {
-        let hint_code = "memory[ap] = to_felt_or_relocatable(x == 0)";
-        let mut vm = vm_with_range_check!();
+        let hint_codes = vec![hint_code::IS_ZERO_NONDET, hint_code::IS_ZERO_INT];
+        for hint_code in hint_codes {
```
Oppen, line 436: advice(non-blockin): you can use `rstest` to define test cases externally. That way a test per case is generated and both run even if one fails, and it has the advantage of being explicit.

Oppen (approved):

src/hint_processor/builtin_hint_processor/secp/field_utils.rs
MegaRedHand: Right. How useful!

MegaRedHand: Added _rstest_ as a dev-dependency and used it here https://github.com/lambdaclass/cairo-rs/pull/1010/commits/dcb29c0dc14a537a12f24e813754c0147afa5c28. This also applies to previous PRs, so I'll be checking that!

Oppen:
I'm not sure what's wrong with Pedersen but it's the third time it erratically reports becoming much slower only to normalize later. I'll review it later, but could you trigger the benchmarks before merging just in case?

Oppen (approved):

fmoletta (approved):

PR #1027, opened 2023-04-20
fix : Fix `CairoRunner::get_memory_holes`

fmoletta:
This PR aims to fix the `get_memory_holes` method which was uncorreclty counting memory holes for builtin segments.
In the python VM, all addresses in the builtin's segment are marked as accessed before counting memory holes. In this PR, builtin segments are skipped when counting memory holes instead.

Changes:

-    Skip builtin segements when counting memory holes

- Check amount of memory holes for all tests in cairo_run_test

-  Remove duplicated tests in cairo_run_test

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Upcoming Changes
 
-* Fix `CairoRunner::get_memory_holes`
+* Fix `CairoRunner::get_memory_holes` [#1027](https://github.com/lambdaclass/cairo-rs/pull/1027):
 
     * Skip builtin segements when counting memory holes
     * Check amount of memory holes for all tests in cairo_run_test
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 #### Upcoming Changes
 
+* Fix `CairoRunner::get_memory_holes`
+
+    * Skip builtin segements when counting memory holes
+    * Check amount of memory holes for all tests in cairo_run_test
+    * Remove duplicated tests in cairo_run_test
+
 * Add methor `Program::data_len(&self) -> usize` to get the number of data cells in a given program [#1022](https://github.com/lambdaclass/cairo-rs/pull/1022)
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 #### Upcoming Changes
 
+* Fix `CairoRunner::get_memory_holes`
+
+    * Skip builtin segements when counting memory holes
+    * Check amount of memory holes for all tests in cairo_run_test
+    * Remove duplicated tests in cairo_run_test
+
```
Oppen: This is a breaking change AFAICT. `count_memory_holes` is marked as public and its signature changed.

fmoletta: Oh, I thought only `CairoRuner::get_memory_holes` was public, maybe we should consider making this one private, as the `MemorySegmentManager` is a private field of the VM. Users should only really  be using``CairoRunner::get_memory_holes`. What do you think?

Oppen: I'm all for privacy. It's not breaking if there's no way to call it tho, but it's better to be explicit. Is the structure itself private?

fmoletta: It would still be a breaking change (making something private), The structure itself is not private (`MemorySegmentManager`), but now you would need to get `vm.builtin_runners.len()` (another private field) to use it effectively, so there is not much reason to keep it in the public interface imo.

Oppen: Yes, it would. I'm not arguing for keeping it public, I'm just asking whether or not it really breaks to document properly. But I'm in favor of making both the structure and the method private.

fmoletta: Structure is still useful for the `VirtualMachineBuilder`, so Ill just make the method private

Oppen: I'm not sure what you mean but I don't want to waste your time and we can go back to that later. Let's leave it on making the method private and counting it as a breaking change just in case, do you agree?

MegaRedHand (approved):

PR #1049, opened 2023-04-25
feat(hints): add NewHint#51

MegaRedHand:
## Description

This PR adds the missing hint _NewHint#51_, now known as `UNSIGNED_DIV_REM_UINT768_BY_UINT384_STRIPPED`.
NOTE: this hint is a copy of _NewHint#6_ but with less spaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -240,6 +240,46 @@
         ids.res.high = res_split[1]
     ```
 
+* Implement hint on vrf.json lib [#1049](https://github.com/lambdaclass/cairo-rs/pull/1049)
+
+    `BuiltinHintProcessor` now supports the following hint:
+    
+    ```python
+        def split(num: int, num_bits_shift: int, length: int):
+            a = []
+            for _ in range(length):
+                a.append( num & ((1 << num_bits_shift) - 1) )
+                num = num >> num_bits_shift
+            return tuple(a)
+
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -379,13 +419,13 @@
...
                 a.append( num & ((1 << num_bits_shift) - 1) )
-                num = num >> num_bits_shift
+                num = num >> num_bits_shift 
             return tuple(a)
 
         def pack(z, num_bits_shift: int) -> int:
             limbs = (z.d0, z.d1, z.d2)
             return sum(limb << (num_bits_shift * i) for i, limb in enumerate(limbs))
-
+
```
Oppen, line 428:
```suggestion

```

Oppen (approved):

CHANGELOG.md
MegaRedHand: The hint implemented in this entry includes the whitespace so I think it's clearer if we also add it on the changelog. Although, now that I think about it a bit more, maybe we should add a comment to clarify why there are two identical hints. Something like: _"Note: this hint is similar to the one in #983, but with some trailing whitespace removed"_
What do you think?

Oppen: Sorry, I missed that this was hint code 🤦

juanbono (approved):

Issue #1007, opened 2023-04-18
Pattern used by proptests doesn't produce numbers bigger than 110 bits


fmoletta:
**Describe the bug**
The pattern `"([1-9][0-9]*)"` doesn't produce numbers bigger than 110 bits, as we are handling big felt values in the vm, we must test values inside the full felt range.
This pattern is used in the felt crate (both lib.rs and bigint_felt.rs) and in math_utils.rs

Issue #1076, opened 2023-04-26
Review proptest failures


pefontana:
While trying to address issue #1007 it was found that the functions `Felt252::sqrt()` is buggy for certain ranges. `Felt::252` signedness related functions (`Felt252::abs()`, `Felt252::is_(positive|negative)()` and `Felt252::signum()`) have unclear semantics, so much that the proptests contradict the implementation, even tho both are "correct" for some interpretation (similarly to #1074).

We need to take a look and find fixes. Do note none seem __currently__ used by `cairo_vm`, so we're safe for now in terms of production exploding because of this.

Issue #1074, opened 2023-04-26
Implement proper to_bigint and rename the current one to something appropriate


pefontana:
During development of the `cairo-felt`, we chose to name the routine converting to the `[-P/2..P/2]` signed representation of field elements `Felt252::to_bigint(&self)`.
At the same time, relatively often we need to get the equivalent to `felt.to_biguint().to_bigint()` to get the exact same representation __typed__ as a signed `BigInt`. This choice of name leads to developers often using `.to_bigint()` expecting the latter behavior, which leads to bugs. Because in the range `[0..P/2]` the value is the same, it's also often not caught by tests, and relies on developer folklore during the code reviews (luckily, most times it __was__ caught, but no guarantees there).

The task is making `.to_bigint()` behave as everyone would expect and to rename the old one to something that makes sense (possibly `.to_signed_felt()` or something similar), and carefully audit uses to see whether they need updating to the new name or the original code was actually buggy.

It could be done in two steps if that makes it easier, as currently __correct__ code does `.to_biguint().to_bigint()` when it just wants the regular type conversions, so the developer could do the renaming first and the conversion routine later.

PR #1094, opened 2023-04-28
test: add `proptest::Strategy` for generating arbitrary `Felt252`s

MegaRedHand:
Fixes: #1007

## Description

This PR adds an implementation of a custom `proptest::Strategy` for generating `Felt252`s. This fixes the issue of generating strings and parsing them on each test and also enables us to explore the whole field range without having to check the number is less than the Cairo prime.

Also bumps the _proptest_ dependency version.

NOTE: this implementation appeared to have some problems in the shrinking process (first half of the digits doesn't go down) when manually testing it.

NOTE: some tests are failing because of #1076 and #1074. These are currently ignored, with the reason documented in-place.

## Checklist
- [X] Linked to Github Issue
- [ ] Unit tests added


felt/src/arbitrary.rs
```diff
@@ -0,0 +1,39 @@
...
+        .prop_map(|(high, low)| {
+            let biguint = (BigUint::from(high) << 128) + low;
+            FeltBigInt::from(biguint)
+        })
+}
+
+/// Returns a [`Strategy`] that generates any valid Felt252
+pub fn any_felt252() -> impl Strategy<Value = Felt252> {
+    any_felt_big_int::<FIELD_HIGH, FIELD_LOW>().prop_map(|value| Felt252 { value })
+}
```
Oppen, line 34: I think these come for free as `any::<T>` as long as the types implement the `Arbitrary` trait.

felt/src/arbitrary.rs
```diff
@@ -0,0 +1,39 @@
...
+        .prop_map(|x| x)
+        // choose second 128-bit limb capped by first one
+        .prop_flat_map(|high| {
+            let low = if high == PH {
+                Strategy::sboxed((0..PH).prop_map(|x| x))
+            } else {
+                Strategy::sboxed(any::<u128>())
+            };
+            (Just(high), low)
+        })
```
Oppen, line 23: I'm not sure I follow this logic. Wouldn't it make more sense to enable `BigUint`'s implementation of `Arbitrary` and giving it a bound of `(PH << 128) | PL`?

felt/src/lib.rs
```diff
@@ -875,20 +878,20 @@ assert_felt_impl!(Felt252);
...
-        // Property-based test that ensures, for 100 felt values that are randomly generated each time tests are run, that a new felt doesn't fall outside the range [0, p].
-        // In this and some of the following tests, The value of {x} can be either [0] or a very large number, in order to try to overflow the value of {p} and thus ensure the modular arithmetic is working correctly.
-        fn new_in_range(ref x in any::<[u8; 40]>()) {
-            let x = &Felt252::from_bytes_be(x);
+        // Property-based test that ensures, for 100 felt values that are randomly generated
+        // each time tests are run, that a new felt doesn't fall outside the range [0, p].
+        // In this and some of the following tests, The value of {x} can be either [0] or a
+        // very large number, in order to try to overflow the value of {p} and thus ensure the
+        // modular arithmetic is working correctly.
+        fn new_in_range(ref x in any_felt252()) {
```
Oppen, line 894: Careful here. We want specifically to test inputs that are out of range in this case, to confirm the constructor adjusts them. So, here we should use an arbitrary `BigUint` or `BigInt` (so we can also work with signed numbers) rather than a "correct" `Felt252`.

felt/src/lib.rs
```diff
@@ -931,20 +934,22 @@ mod test {
...
-        fn from_bytes_be_in_range(ref x in FELT_PATTERN) {
-            let x = &Felt252::from_bytes_be(x.as_bytes());
-            let max_felt = &Felt252::max_value();
+        // Property-based test that ensures, for 100 felt values that are randomly
+        // generated each time tests are run, that a felt created using Felt252::from_bytes_be doesn't
+        // fall outside the range [0, p].
+        // In this and some of the following tests, The value of {x} can be either [0] or a very large number,
+        // in order to try to overflow the value of {p} and thus ensure the modular arithmetic is working correctly.
+        fn from_bytes_be_in_range(x in any_felt252()) {
+            let max_felt = Felt252::max_value();
```
Oppen, line 943: And the same applies here. As a general rule, if we're testing construction from arbitrary input, the __input__ is what should be arbitrary, not the __value__.

felt/src/arbitrary.rs
MegaRedHand: The thing is, we need `proptest::Arbitrary`, and `BigUint` only implements _quickcheck_'s and _arbitrary_'s.

felt/src/lib.rs
MegaRedHand: Thanks for the input. I completely overlooked these types of tests. I'll search for these cases and fix them.

felt/src/arbitrary.rs
MegaRedHand: Yeah, it's better to impl the trait. I started doing it and quickly got stuck because in the trait you have to be explicit with the return type of its method (i.e. no `impl Strategy` magic), so it requires a bit more thought. Now that I have this implementation in place, I think that issue could be solved just by using a boxed strategy. I'll check this later.

felt/src/arbitrary.rs
Oppen: I thought it used [the arbitrary crate's version](https://github.com/rust-fuzz/arbitrary), which `num_bigint` [also implements](https://github.com/rust-num/num-bigint/blob/master/src/biguint/arbitrary.rs#L21-L34). But looking at the source code it seems it uses its own as you said.
Re: the logic. I think the reason I don't get it is that inside the `if` there's a typo. It should be `(0..PL)` rather than `(0..PH)`.

MegaRedHand: Fixed

felt/src/arbitrary.rs
MegaRedHand: Done. Although I haven't found an easy way to generate a nonzero felt with the trait. As far as I know there are two ways:
- use `prop_filter`, like the function `nonzero_felt252` does
- receive some parameters that change the generation strategy in the trait, and use `any_with::<Felt252>(args)` instead of plain `any::<Felt252>()`

felt/src/lib.rs
```diff
@@ -886,37 +889,36 @@ assert_felt_impl!(Felt252);
...
     use super::*;
+    use crate::arbitrary::nonzero_felt252;
     use core::cmp;
     use proptest::prelude::*;
 
-    const FELT_PATTERN: &str = "(0|[1-9][0-9]*)";
-    const FELT_NON_ZERO_PATTERN: &str = "[1-9][0-9]*";
-
     proptest! {
+        #![proptest_config(ProptestConfig::with_cases(100000))]
```
fmoletta, line 897:
```suggestion
```

fmoletta (approved):
LGTM :)

juanbono (approved):

PR #1098, opened 2023-05-02
Run `CasmContractClass`

fmoletta:
# TITLE

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


.github/workflows/rust.yml
```diff
@@ -11,6 +11,33 @@ env:
...
+    # actually newer than the sources, otherwise it will try to rebuild them
+    - name: Restore timestamps
+      uses: chetan/git-restore-mtime-action@v1
+
+    - name: Install cairo 1 compiler
+      run: make build-cairo-1-compiler
+
+    - name: Build contracts
+      run: make -j cairo_1_test_contracts
+
```
Oppen, line 40: Can this be done extending the matrix in the other job?

Cargo.toml
```diff
@@ -79,6 +79,12 @@ felt = { package = "cairo-felt", path = "./felt", version = "0.3.0-rc1", default
     "alloc",
 ] }
 bitvec = { version = "1", default-features = false, features = ["alloc"] }
+cairo-lang-starknet = { git = "https://github.com/starkware-libs/cairo.git", rev = "a08bedbf26b711e81cec8b7660698a3788e67d46"}
+cairo-lang-casm = { git = "https://github.com/starkware-libs/cairo.git", rev = "a08bedbf26b711e81cec8b7660698a3788e67d46" }
+
+# TODO: check these dependencies for wasm compatibility
+ark-ff = "0.4.0-alpha.7"
+ark-std = "0.3.0"
```
Oppen, line 87: We can start by using the `dep = { version = "x.y.z", default-features = false }` notation.

Makefile
```diff
@@ -110,6 +132,7 @@ deps:
 
 deps-macos:
 	$(MAKE) cargo-deps
+	$(MAKE) build-cairo-1-compiler
```
Oppen, line 135: I think both of these should be dependencies for the rule instead.

src/hint_processor/cairo_1_hint_processor/dict_manager.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+/// Helper object for the management of dict_squash hints.
+#[derive(Default, Debug)]
+pub struct DictSquashExecScope {
+    /// A map from key to the list of indices accessing it, each list in reverse order.
+    pub access_indices: HashMap<Felt252, Vec<Felt252>>,
+    /// Descending list of keys.
+    pub keys: Vec<Felt252>,
```
Oppen, line 97: This are probably better left private.

src/tests/mod.rs
```diff
@@ -79,3 +86,111 @@ pub(self) fn run_program(
...
     }
 }
+
+// Runs a contract entrypoint with given arguments and checks its return values
+// Doesn't use a syscall_handler
+pub(self) fn run_cairo_1_entrypoint(
+    program_content: &[u8],
+    entrypoint_offset: usize,
+    args: &Vec<MaybeRelocatable>,
+    expected_retdata: &Vec<Felt252>,
```
Oppen, line 96: Use slices if possible.

Oppen (commented):
I think we should extract this into a separate crate. I think I should also insist on trying to integrate early, hiding the feature behind a compile time flag instead.

.github/workflows/rust.yml
fmoletta: The file extension and the compiler is different so I felt it was more clear to do it in a separate job (as there would be added subtasks anyway)

src/hint_processor/cairo_1_hint_processor/dict_manager.rs
fmoletta: pub(crate) is my best offer before we refactor the hint processor code

.github/workflows/rust.yml
Oppen: That can be parameterized too. See for example the `run-cairo-release` job.

Oppen:
> This was a separate crate before, but that meant we couldn't import the Cairo1HintProcessor for testing in cairo-rs

The tests in the crate can import the VM tho.

fmoletta:
> > This was a separate crate before, but that meant we couldn't import the Cairo1HintProcessor for testing in cairo-rs
> 
> The tests in the crate can import the VM tho.

But then we'd have tests that run contracts in a crate that is just meant to be a hint processor

pefontana (approved):

pefontana:
Nice one @fmoletta !!!
I too think that it would be better to have this in a separate crate, but as long as it's easy to move it later to a cairo-lang crate then it's OK for me

.github/codecov.yml
```diff
@@ -8,4 +8,5 @@ ignore:
   - src/vm/errors
   - src/types/errors
   - hint_accountant
+  - src/hint_processor/cairo-1-hint-processor # TODO: Remove this line
```
pefontana, line 11: when we should remove this?

fmoletta: When we add unit tests/integration tests for all hints

juanbono (approved):
LGTM

Oppen (approved):

PR #1151, opened 2023-05-10
feat: Add method `CairoRunner::initialize_function_runner_cairo1` 

fmoletta:
This PR aims to solve the issue of `CasmContractClass` having the builtins stored by entrypoint instead of a universal builtin field.
In order to fix this, the method `CairoRunner::initialize_function_runner_cairo1` was added which allows setting the program builtins
## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,22 @@
 
 #### Upcoming Changes
 
-* Add more hints to `Cairo1HintProcessor` [#1143](https://github.com/lambdaclass/cairo-rs/pull/1098)
+* feat: Add method `CairoRunner::run_from_entrypoint_with_builtins` [#1151](https://github.com/lambdaclass/cairo-rs/pull/1151)
+
+  * Add method `run_from_entrypoint_with_builtins(
+        &mut self,
+        entrypoint: usize,
+        args: &[&CairoArg],
+        verify_secure: bool,
+        program_segment_size: Option<usize>,
+        vm: &mut VirtualMachine,
+        hint_processor: &mut dyn HintProcessor,
+        program_builtins: Option<&[BuiltinName]>,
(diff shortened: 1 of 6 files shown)


src/vm/runners/cairo_runner.rs
```diff
@@ -969,12 +966,30 @@ impl CairoRunner {
         Ok(())
     }
 
+    /// Intitializes the runner in order to run cairo 1 contract entrypoints
+    /// Initializes builtins & segments
+    /// All builtins are initialized regardless of use following program ordering then starknet ordering
```
Oppen, line 971: Is the ordering really relevant or just how it's implemented? I.e., would the ordering change the behavior in a meaningful way?
If it isn't, I wouldn't mention it. If it is, I think it may merit some more details about why.

Oppen (approved):
LGTM, I only have a non-blocking question about a comment.

src/vm/runners/cairo_runner.rs
fmoletta: Its not relevant, Ill remove it

Oppen:
To save you time: the error in the tests for wasm boils down to the global prelude of Rust not being there. The new helpers and macro need to use `crate::stdlib::String` instead of just `String`.

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 #### Upcoming Changes
 
+* feat: Add method `CairoRunner::initialize_function_runner_cairo1` [#1151](https://github.com/lambdaclass/cairo-rs/pull/1151)
```
pefontana, line 5: typo
initialize_function_runner_cairo_1

fmoletta: Nice catch!

PR #1175, opened 2023-05-22
Implement run resources funcs

mmsc2:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


src/utils.rs
```diff
@@ -47,6 +47,33 @@ pub fn from_relocatable_to_indexes(relocatable: Relocatable) -> (usize, usize) {
         (relocatable.segment_index as usize, relocatable.offset)
     }
 }
+// ================
+//   RunResources
+// ================
+
+/// Maintains the resources of a cairo run. Can be used across multiple runners.
+#[allow(dead_code)]
```
fmoletta, line 55: Remove these

src/utils.rs
```diff
@@ -47,6 +47,33 @@ pub fn from_relocatable_to_indexes(relocatable: Relocatable) -> (usize, usize) {
...
+pub struct RunResources {
+    n_steps: Option<usize>,
```
fmoletta, line 57: `RunResources` is already an optional argument, do we need an option inside it too?

src/vm/runners/cairo_runner.rs
```diff
@@ -516,20 +517,26 @@ impl CairoRunner {
     pub fn run_until_pc(
         &mut self,
         address: Relocatable,
+        run_resources: &mut Option<RunResources>,
```
fmoletta, line 520: Maybe this could be `Option<&mut RunResources>` instead?
As to avoid having to call it with `&mut None`

src/vm/runners/cairo_runner.rs
```diff
@@ -883,6 +890,7 @@ impl CairoRunner {
         &mut self,
         entrypoint: usize,
         args: &[&CairoArg],
+        run_resources: &mut Option<RunResources>,
```
fmoletta, line 893: Why is it not mutable here?

mmsc2: what is not mutable? run_resources?

fmoletta: Sorry, misread

src/vm/runners/cairo_runner.rs
mmsc2: i was unsure which one would be a better option but the suggestion is fair enough, besides it matches the behavior it has in the function.

src/utils.rs
fmoletta: Using 0 as default would lead to unwanted behaviour. What I mean is that if the `RunResources` struct is itself optional and its inner value is never updated when it is set as  None, then all use cases in which the function would receive a run_resources = RunResources(None) can jus receive run_resources = None. In that case we don't need a default value

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* feat: Add structure RunResources [#1173](https://github.com/lambdaclass/cairo-rs/pull/1173)
```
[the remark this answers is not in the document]
mmsc2: fixed

src/vm/runners/cairo_runner.rs
Oppen: It will be ugly but I think `mut Option<&mut RunResources>` is likely to work.

src/vm/runners/cairo_runner.rs
```diff
@@ -516,20 +517,32 @@ impl CairoRunner {
...
         vm.execute_before_first_step(self, &hint_data_dictionary)?;
-        while vm.run_context.pc != address {
+
+        while vm.run_context.pc != address
+            && !if let Some(r) = run_resources.as_ref() {
+                r.consumed()
+            } else {
+                false
+            }
+        {
```
Oppen, line 536: This looks confusing. Can you find a way to make it more straightforward?

Oppen (approved):
LGTM but see comment.

src/vm/runners/cairo_runner.rs
mmsc2: i think i can pull out a solution by calling a function that engulfs that behavior.

pefontana (changes requested):
I think that this PR https://github.com/lambdaclass/cairo-rs/pull/1181 will add this feature without making changes to the `CairoRunner.run_from_entrypoint` API
For now, let's leave this PR on stand by

fmoletta (approved):

juanbono (approved):

pefontana (approved):

PR #1185, opened 2023-05-29
Catch possible subtraction error in `QUAD_BIT` & `DI_BIT` hints

fmoletta:
Now returns an error if ids.m is zero

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/src/vm/errors/hint_errors.rs
+++ b/src/vm/errors/hint_errors.rs
@@ -179,4 +179,6 @@ pub enum HintError {
     RecoverYPointNotOnCurve(Felt252),
     #[error("Invalid value for {0}. Got: {1}. Expected: {2}")]
     InvalidValue(&'static str, Felt252, Felt252),
+    #[error("Attempt to subtract with overflow: ids.m - 1")]
+    NPairBitsMZero,
 }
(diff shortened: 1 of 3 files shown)


juanbono (approved):

src/vm/errors/hint_errors.rs
```diff
@@ -179,4 +179,6 @@ pub enum HintError {
     RecoverYPointNotOnCurve(Felt252),
     #[error("Invalid value for {0}. Got: {1}. Expected: {2}")]
     InvalidValue(&'static str, Felt252, Felt252),
+    #[error("Attempt to subtract with overflow: ids.m - 1")]
+    NPairBitsMZero,
```
juanbono, line 183: Maybe we should use a more descriptive name to this new error

fmoletta: I chose this name as it is more useful when debugging, as it has the name of the function that fails and the failure case, while leaving the error string useful for users, a more descriptive name could be `NPairBitsMMinusOneSubtractioOverflow` or `MMinusOneSubtractionOverflow` but it doesn't sound that much useful

pefontana (approved):

PR #1214, opened 2023-06-07
perf: reference management optimizations

Oppen:
- Store the reference list rather than the manager structure
- Process the reference list only on `Program` creation rather than in each call to `run_for_steps`
- Use a `Vec<_>` rather than a `HashMap<_, _>` since the instance is dense
- Keep this data in the `SharedProgramData` structure instead of copying it each time
- Improve the accuracy of the benchmarks

## Benchmarks

```
build runner            time:   [773.76 ns 774.50 ns 775.47 ns]                          
                        change: [-93.906% -93.895% -93.885%] (p = 0.00 < 0.05)
initialize              time:   [20.632 µs 20.649 µs 20.672 µs]                        
                        change: [-16.707% -15.351% -14.383%] (p = 0.00 < 0.05)
parse program           time:   [9.2019 ms 9.2176 ms 9.2341 ms]                          
                        change: [+0.1218% +0.2751% +0.4491%] (p = 0.00 < 0.05)
```


unbalancedparentheses:
image

MegaRedHand (approved):

pefontana (approved):

PR #1215, opened 2023-06-08
add ensure-no-std crate to workspace + add ensure_no_std crate

tdelabro:
# no_std check

## Description
add a way to check that the crates are not breaking their no-std compatiblity

Some times ago, new features were added to cairo-vm that broke it's no-std compatibility. 
This PR add the crate `ensure_no_std` to the repo. It is empty a binary, to be compiled with dependency on the crates we want to test for no-std (cairo-vm, and cairo-felt). We should build it in the CI, if the build fail it is that one of the dependency has break it's no-std compatibility.

It has to be build with nightly and to a the `wasm32-unknow-unknow` target. Which mean it cannot be part of the workspace. But for some reason, `members = ["."]` combined to the `exclude = ["ensure_no_std"]` caused trouble to the rust compiler. 

So I moved the VM in it's own directory, `vm` and split the root Cargo.toml between its `workspace` part, which stay at the same place, and it's `cairo-vm` part which has been moved into the `vm` directory.

This allow for a much more clear config, clearly stating what is at the workspace level and what only concerns the `vm` crate.

It is not perfect. It would be great that a subsequent PR make every crates in the repo depends on the workspace rather than each one re-specifying it's own deps. But it is still better than the previous state.

The `ensure_no_std` only check for `cairo-felt` for now because `cairo-vm` no-std is broken. It is not used in the CI yet, it will when the `vm` will be fixed again


tdelabro:
@juanbono @Oppen Would you like to give it a review?
The only thing blocking now is this problem with path.

The `.json` file keeps a relative path to itself, starting from where it has been compiled. If you execute the program in the same directory, everything works nicely, but if you do it from another directory, it won't be able to find the `.json` file.
I don't know how to fix this exactly, because `run_program` takes the program as a `&[u8]` without ever knowing where the file was generated initially.

tdelabro:
And I don't know what is going on with this one: https://github.com/lambdaclass/cairo-rs/actions/runs/5220573543/jobs/9423831041?pr=1215

Cargo.toml
```diff
@@ -83,35 +56,12 @@ felt = { package = "cairo-felt", path = "./felt", version = "0.5.1", default-fea
 bitvec = { version = "1", default-features = false, features = ["alloc"] }
 
 # Dependencies for cairo-1-hints feature
-cairo-lang-starknet = {version = "1.1.0", optional = true}
-cairo-lang-casm = {version = "1.1.0", optional = true }
+cairo-lang-starknet = { version = "1.1.0"}
+cairo-lang-casm = { version = "1.1.0"}
```
Oppen, line 60: Why are these no longer optional?

Cargo.lock
Oppen, line 1: Some deps' versions seem to have gone downgraded. Why is that?

Oppen (approved):
On the whole it looks good. I have just two questions.

Oppen:
There's one more issue. It seems the IAI benchmark no longer finds the data from the previous run. That may happen if the `target` dir moves to the subcrates rather than the project's root, which I think is the case when there's no crate in the root as it is the case now.

Cargo.toml
tdelabro: The workspace does not accept optional dependencies.
It is just a register of versions of crates for the crates `Cargo.toml` to use. 
The crate's responsible for saying if they are optional or not.

Cargo.lock
tdelabro: I truly have no idea ¯\_(ツ)_/¯

MegaRedHand (approved):

tdelabro:
@Oppen nope, target is still being build at the root level

vm/src/vm/errors/vm_exception.rs
```diff
@@ -256,6 +256,14 @@ impl Location {
...
         let mut string = self.to_string(message);
         let input_file_path = std::path::Path::new(&self.input_file.filename);
+        #[cfg(test)]
+        let input_file_path = {
+            use std::path::PathBuf;
+            let current_dir = std::env::current_dir().unwrap();
+            let mut parent_dir: PathBuf = current_dir.parent().unwrap().into();
+            parent_dir.push(input_file_path);
+            parent_dir
+        };
```
tdelabro, line 266: This fixes the path problem.
As tests are run from the crate directory itself, we can just get the current_dir, take its parent, and then append the `input_file_path`.

It gives us an absolute path that works.
That doesn't solve the case where programs are not compiled from the same dir as the one the vm is executed. But it will be a problem for later

Oppen: That's cool. Could you add that information as a little comment in place?

Oppen:
> @Oppen nope, target is still being build at the root level

The build itself is doing that, but `iai-callgrind` has peculiar behavior with regards to that. I just confirmed in the server the benchmark results are stored within `vm`:

```
[pelito@archlinux-latest-64-minimal cairo-rs]$ ls vm/target/iai/iai_benchmark/
callgrind.build_runner.out  callgrind.load_program_data.out  callgrind.parse_program.out
```

We need to update the `iai` workflows so the caches look like this:
```
    - name: Save cache
      if: ${{ steps.cache-iai-results.outputs.cache-hit != 'true' }}
      uses: actions/cache/save@v3
      with:
        path: |
          */target
        key: ${{ runner.os }}-iai-benchmark-cache-${{ github.event.pull_request.base.sha }}
 ```

Note the path changes so it looks __inside crate specific target dirs__.

File: ensure_no_std/rust-toolchain.toml [hunk 3]
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "nightly"
```
pefontana, line 2: Is nightly necessary?

Cargo.lock
Oppen: We can run `cargo update` later then lol

pefontana (approved):
Amazing work @tdelabro !!!! 
🎉

pefontana:
@tdelabro Can you please add a line in the CHANGELOG.md explaining the added feature and we will be OK to go!

PR #1218, opened 2023-06-09
fix: substraction between relocatables should be signed

Oppen:
It is valid as per the reference implementation to substract two relocatables `a - b` where `b.offset > a.offset`. The current code returns an error due to negative result in that case.
The fix consists in not using the trait as implemented for `Relocatable` (where it represents a positive distance between pointers) but to promote the offsets to `i128` and only substract them for `MaybeRelocatable`, as those are the language-level types.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/src/tests/cairo_run_test.rs
+++ b/src/tests/cairo_run_test.rs
@@ -970,9 +970,6 @@ fn cairo_run_reduce() {
 
 #[test]
 fn cairo_run_if_reloc_equal() {
-    const PROGRAM_DATA: &[u8] = include_bytes!(concat!(
-        env!("CARGO_MANIFEST_DIR"),
-        "/cairo_programs/if_reloc_equal.json",
-    ));
-    run_program_simple(PROGRAM_DATA);
+    let program_data = include_bytes!("../../cairo_programs/if_reloc_equal.json");
+    run_program_simple_with_memory_holes(program_data, 4);
 }
--- a/src/tests/cairo_run_test.rs
+++ b/src/tests/cairo_run_test.rs
@@ -967,3 +967,12 @@ fn cairo_run_reduce() {
     let program_data = include_bytes!("../../cairo_programs/reduce.json");
     run_program_simple(program_data.as_slice());
 }
+
+#[test]
+fn cairo_run_if_reloc_equal() {
+    const PROGRAM_DATA: &[u8] = include_bytes!(concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/cairo_programs/if_reloc_equal.json",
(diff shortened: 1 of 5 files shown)


src/tests/cairo_run_test.rs
```diff
@@ -967,3 +967,12 @@ fn cairo_run_reduce() {
...
     run_program_simple(program_data.as_slice());
 }
+
+#[test]
+fn cairo_run_if_reloc_equal() {
+    const PROGRAM_DATA: &[u8] = include_bytes!(concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/cairo_programs/if_reloc_equal.json",
+    ));
+    run_program_simple(PROGRAM_DATA);
```
MegaRedHand, line 977:
```suggestion
    run_program_simple_with_memory_holes(PROGRAM_DATA, 4);
```

src/vm/vm_core.rs
```diff
@@ -449,6 +452,7 @@ impl VirtualMachine {
             *instruction = Some(self.decode_current_instruction()?);
         }
         let instruction = instruction.as_ref().unwrap();
+        dbg!(instruction);
```
MegaRedHand, line 455:
```suggestion
```

MegaRedHand (approved):
LGTM. We can address TODOs and FIXMEs in following PRs.

pefontana (approved):

src/types/relocatable.rs
```diff
@@ -256,7 +256,9 @@ impl MaybeRelocatable {
                 MaybeRelocatable::RelocatableValue(rel_b),
             ) => {
                 if rel_a.segment_index == rel_b.segment_index {
-                    return Ok(MaybeRelocatable::from(Felt252::new((*rel_a - *rel_b)?)));
+                    return Ok(MaybeRelocatable::from(Felt252::from(
+                        rel_a.offset as i128 - rel_b.offset as i128,
```
pefontana, line 260: I think it should be better to leave this implementation as it is and move this modification to the Relocatable sub Relocatable operation. So we don't have two different implementations of the Relocatable sub.
Do you think it is ok?

pefontana (changes requested):

pefontana (approved):

src/types/relocatable.rs
pefontana: I think I was wrong on this one
the Relocatable - Relocatable returns an usize

pefontana (approved):

juanbono (approved):

PR #1227, opened 2023-06-13
Cairo fuzzer

Juan-M-V:
# Fuzzing

## Description

Create a simple fuzzer to test the VM


Oppen (approved):

fuzzer/src/fuzz_json.rs
```diff
@@ -0,0 +1,165 @@
...
+use std::io::{self, Write};
+use std::path::PathBuf;
+use thiserror::Error;
+
+#[cfg(feature = "with_mimalloc")]
+use mimalloc::MiMalloc;
+
+#[cfg(feature = "with_mimalloc")]
+#[global_allocator]
+static ALLOC: MiMalloc = MiMalloc;
```
MegaRedHand, line 19: I think this isn't needed, as the VM already does this.

MegaRedHand (approved):
🚀 🚀

fuzzer/src/fuzz_json.rs
Oppen: That would (or should) be the CLI. But in any case, I'm not sure it's necessary for the fuzzer. The speed gains may be useful to try more paths, but other than that I don't think it should matter.

PR #1231, opened 2023-06-13
Add fuzzer job

daphneherlambda:
# Add fuzzer job 

## Description

This PR adds a workflow to run the fuzzer everyday at midnight

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


.github/workflows/fuzzer.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential
+
+    - name: Run fuzzer
+        run: |
+          cd fuzzer
+          cargo hfuzz run fuzzer
```
Juan-M-V: Idk if this is necessary but this is the command to limit how long the fuzzer runs. It could be chained with corpus minimization and such
--run_time VALUE
	Number of seconds this fuzzing session will last (default: 0 [no limit])

daphneherlambda: yes!!! it will be updated later, is a work in progress yet

PR #1233, opened 2023-06-14
Add `disable_trace_padding` to `CairoRunConfig`

fmoletta:
This option will allow the user to disable the trace padding added by default when running in proof-mode

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added


pefontana (approved):

vm/src/cairo_run.rs
```diff
@@ -64,7 +66,12 @@ pub fn cairo_run(
     cairo_runner
         .run_until_pc(end, &mut run_resources, &mut vm, hint_executor)
         .map_err(|err| VmException::from_vm_error(&cairo_runner, &vm, err))?;
-    cairo_runner.end_run(false, false, &mut vm, hint_executor)?;
+    cairo_runner.end_run(
+        cairo_run_config.disable_trace_padding,
```
Oppen, line 70: Can't this be copied to `CairoRunner` so we don't add parameters?

fmoletta: I think adding an extra field to the CairoRunner is too much of a breaking change for something that has a limited use case.

fmoletta: This change only adds a field to the CairoRunConfig

Oppen (approved):

juanbono (approved):

PR #1232, opened 2023-06-13
Add `ensure_no_std` build to CI

MegaRedHand:
# TITLE

## Description

This PR builds the recently added `ensure_no_std` crate in the CI, to ensure that no-std works.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.
  - [ ] CHANGELOG has been updated.

PR #1238, opened 2023-06-15
use our own is_prime impl

tdelabro:
# custom is_prime

## Description

`num-prime` cannot be used in a `no_std` env.
This PR replace the calls to `num_prime::is_prime` with our custom impl

It incidentally restore `no_std` support for `cairo-vm`


MegaRedHand (approved):

vm/src/math_utils/is_prime.rs
```diff
@@ -0,0 +1,930 @@
...
+            "25195908475657893494027183240048398571429282126204032027777137836043662020707595556264018525880784406918290641249515082189298559149176184502808489120072844992687392807287776735971418347270261896375014971824691165077613379859095700097330459748808428401797429100642458691817195118746121515172654632282216869987549182422433637259085141865462043576798423387184774447920739934236584823824281198163815010674810451660377306056201619676256133844143603833904414952634432190114657544454178424020924616515723350778707749817125772467962926386356373289912154831438167899885040445364023527381951378636564391212010397122822120720357",
+        ];
+
+        for n in rsa_semiprimes {
+            let n = BigUint::from_str(n).unwrap();
+            assert!(is_prime(&n));
```
MegaRedHand, line 927:
```suggestion
assert!(!is_prime(&n));
```

tdelabro:
By the way, do you have a way to check if this new way of checking for prime numbers does not hurt performances?

tdelabro:
@MegaRedHand 
`base field_arithmetic_get_square_benchmark` | 153.4 ± 5.7 | 148.9 | 168.5 | 1.00 |\\n|` 

`± 5.7` is it bad?

MegaRedHand:
It ran ~14% slower than _main_ (you can see that [here](https://github.com/lambdaclass/cairo-rs/actions/runs/5282756274/jobs/9558102630?pr=1238#step:7:90)), but we already patched it by adding the original implementation as an _std_ alternative in #1232. Also opened an issue to look for a better alternative: #1243.

Issue #1235, opened 2023-06-14
Update the custom hint example


MegaRedHand:
[This example](https://github.com/lambdaclass/cairo-rs/tree/main/custom_hint_example) currently isn't included in the workspace, so it's not run in the CI. Also, its deps [need updating](https://github.com/lambdaclass/cairo-rs/security/dependabot/2).

PR #1258, opened 2023-06-19
Update the custom hint example

d-roak:
# Update the custom hint example

## Description
- Added custom_hint_example to workspace
- Removed local Cargo.lock
- Fixed the example

## Checklist
- [x] Linked to Github Issue (#1235)
- [ ] Unit tests added


CHANGELOG.md
```diff
@@ -32,6 +32,8 @@
 
 * replace the use of `num-prime::is_prime` by a custom implementation, therefore restoring `no_std` compatibility
 
+* fix: updated the `custom_hint_example` and added it to the workspace [#1258](https://github.com/lambdaclass/cairo-rs/pull/1258)
```
MegaRedHand, line 35: This should go in a new section (0.6.0 was already released)

Cargo.toml
```diff
@@ -5,6 +5,7 @@ members = [
     "vm",
     "hint_accountant",
     "./deps/parse-hyperlinks",
+    "./custom_hint_example"
```
MegaRedHand, line 8: I was thinking it could be added as an "example" instead of as a crate. That way it shouldn't be considered by codecov.

d-roak: For that, I need to change the folder structure to put the example inside an `examples` folder (https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples), is it fine?

MegaRedHand: I think that would be for the better

d-roak: I am not allowed to add examples in root `Cargo.toml`, because it is a workspace definition.
I added the example definition, in the vm package. I could have also kept the same folder structure for the example, but changed it anyway to the folder `examples`, because it is a rust standard and it seems cleaner.
Plz check if it is the intended solution

Oppen: I give my 👍 to that solution. WDYT @MegaRedHand?

MegaRedHand: I think the same. This way it's easier to find them in the repo.

examples/custom_hint/src/main.rs
```diff
@@ -1,30 +1,36 @@
...
     ap_tracking: &ApTracking,
-    _constants: &HashMap<String, BigInt>,
+    _constants: &HashMap<String, Felt252>,
 ) -> Result<(), HintError> {
     let a = get_integer_from_var_name("a", vm, ids_data, ap_tracking)?;
     println!("{}", a);
     Ok(())
 }
 
+#[cfg(feature = "std")]
```
MegaRedHand, line 33: This `#[cfg]` may be a bit confusing. I think we can avoid it by adding a [`required-features`](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field) field to the `Cargo.toml` target:
```
[[example]]
...
required-features = ["std"]
```

MegaRedHand (approved):

MegaRedHand:
Closes: #1235

examples/custom_hint/src/main.rs
d-roak: Applied change. Although we have that tag in many places because of the `no_std` (so it shouldn't be confusing), I agree that is way cleaner to have it in the `Cargo.toml`, since the example is not supposed to run without `std` ever

MegaRedHand (approved):

examples/custom_hint/src/main.rs
```diff
@@ -1,14 +1,18 @@
-use cairo_vm::cairo_run::cairo_run;
+#![allow(unused_imports)]
```
fmoletta, line 1: Is this necessary?

fmoletta (approved):

examples/custom_hint/src/main.rs
d-roak: It was on [35afff0](https://github.com/lambdaclass/cairo-rs/pull/1258/commits/35afff024c5519e15161832edcab0de63d48a63b) (not currently)

pefontana (approved):

PR #1266, opened 2023-06-20
build(vm): remove unused implicit features

tdelabro:
## Description

As stated in the [rust doc](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies)
`In some cases, you may not want to expose a feature that has the same name as the optional dependency. For example, perhaps the optional dependency is an internal detail, or you want to group multiple optional dependencies together, or you just want to use a better name. If you specify the optional dependency with the dep: prefix anywhere in the [features] table, that disables the implicit feature.`

This makes sure no implicit features are created for the optional crates that are imported only to be used in another feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* build: remove unused implicit features from cairo-vm
+
 #### [0.6.1] - 2023-6-23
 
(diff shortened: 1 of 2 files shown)


fmoletta (approved):

tdelabro:
@juanbono @fmoletta I solved the conflicts. Ready to be merged

PR #1267, opened 2023-06-20
CI(vm): remove the wasm32 testing target

tdelabro:
# remove the wasm32 testing targe

## Description

I added the whole wasm-bingen, wasm32 testing target at the beginning of the no_std integration.
But with time it became more and more obvious to me that it is not needed. Everything that compiles in no_std always also compiles in wasm32.
On the other hand, it makes the code messier and adds a thrid round of tests in the CI which takes time and resources.

 I think we can safely get rid of it.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -172,7 +172,7 @@ jobs:
       fail-fast: false
       matrix:
         special_features: ["", "lambdaworks-felt"]
-        target: [ test, test-no_std, test-wasm ]
+        target: [ test, test-no_std ]
         # TODO: features
     name: Run tests
     runs-on: ubuntu-22.04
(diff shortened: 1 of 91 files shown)


felt/src/lib.rs
Oppen, line 1: The editor seems to have messed up with the formatting in this file. There are many over-indented lines.

Oppen (dismissed):

felt/src/lib.rs
tdelabro: there is no more change to this file

tdelabro:
@Oppen I just rebased :)

Oppen:
@juanbono could you take a look? We also need to remove the check for `test-wasm`.

tdelabro:
> We also need to remove the check for `test-wasm`.

Did I missed removing it?

Oppen:
> Did I missed removing it?

No, I mean the "check" in Actions parlance. It's a repository level config, I need someone with admin rights (@juanbono) to do it. Otherwise it blocks the merge.

tdelabro:
Still think this one is relevant

PR #1268, opened 2023-06-20
Add `--air_public_input` to runner

### LINKED ISSUE #1201: Public Input as an output of Cairo Runner
Starkware Cairo Runner has an option `--air_public_input` to output the public input that's used in the prover. This feature is missing in Cairo RS. This forces the Lambdaworks Prover to do some workarounds to deduce the inputs by themselves, but it's hacky and inneficient. 

We would like to have this flag in Cairo RS

xqft:
# Add `--air_public_input` to runner

## Description

Closes #1201 by implementing the `--air_public_input` flag to the Cairo runner based on cairo-lang's.

This PR also:
- Fixes runner CLI errors
- Fixes builtins segments stop addresses as suggested by @fmoletta 

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added


vm/src/vm/vm_core.rs
```diff
@@ -1024,6 +1026,26 @@ impl VirtualMachine {
...
+        self.builtin_runners
+            .iter()
+            .map(|builtin| (builtin.name(), builtin.get_memory_segment_addresses()))
+            .collect()
+    }
+
+    pub fn get_trace(&self) -> &Option<Vec<TraceEntry>> {
+        &self.trace
+    }
 }
```
fmoletta, line 1049: Not needed, use `VirtualMachine::get_relocated_trace` instead

cairo-vm-cli/src/main.rs
```diff
@@ -160,6 +172,30 @@ fn run(args: impl Iterator<Item = String>) -> Result<(), Error> {
...
+        };
+
+        let public_input = PublicInput::new(
+            cairo_runner.relocated_memory.clone(),
+            &args.layout,
+            dyn_layout,
+            vm.get_public_memory_addresses()
+                .map_err(VirtualMachineError::Memory)?,
+            vm.get_memory_segment_addresses(),
+            vm.get_trace()
```
fmoletta, line 188:
```suggestion
            vm.get_relocated_trace()
```

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,78 @@
...
+            layout_params: dyn_layout_params,
+            rc_min,
+            rc_max,
+            n_steps: trace.len(),
+            memory_segments: {
+                let mut memory_segment_addresses = memory_segment_addresses.clone();
+                memory_segment_addresses
+                    .insert("program", (trace[0].pc, Some(trace[trace.len() - 1].pc)));
+                memory_segment_addresses
+                    .insert("execution", (trace[0].ap, Some(trace[trace.len() - 1].ap)));
```
fmoletta, line 67: Use `get` insead of indexing directly into the trace vec, as this can panic

fmoletta (commented):
We should also add a method `CairoRunner::get_air_public_input() -> PublicInput` so that library users can get it without needing to use the cli/read from a file

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,78 @@
...
+        memory_segment_addresses: HashMap<&'static str, (usize, Option<usize>)>,
+        trace: &Vec<TraceEntry>,
+        rc_limits: Option<(isize, isize)>,
+    ) -> Self {
+        let public_memory = public_memory_addresses
+            .into_iter()
+            .map(|(address, page)| PublicMemoryEntry {
+                address,
+                page: *page,
+                value: memory[address].clone(),
```
fmoletta, line 46: Use `get` instead of indexing directly

cairo-vm-cli/src/main.rs
```diff
@@ -160,6 +172,30 @@ fn run(args: impl Iterator<Item = String>) -> Result<(), Error> {
...
     }
 
+    if let Some(file_path) = args.air_public_input {
+        let dyn_layout = match args.layout.as_str() {
+            "dynamic" => Some(cairo_runner.get_layout()),
+            _ => None,
+        };
+
+        let public_input = PublicInput::new(
+            cairo_runner.relocated_memory.clone(),
```
fmoletta, line 182: Could we take a reference instead of cloning the vec?

vm/src/vm/vm_memory/memory_segments.rs
```diff
@@ -213,6 +213,25 @@ impl MemorySegmentManager {
...
+    pub fn get_public_memory_addresses(
+        &self,
+        segment_offsets: &Vec<usize>,
+    ) -> Vec<(usize, &usize)> {
+        let len = self.num_segments().min(self.public_memory_offsets.len());
+        let mut addresses = Vec::with_capacity(len);
+
+        for segment_index in 0..len {
+            let offsets = &self.public_memory_offsets[&segment_index];
+            let segment_start = segment_offsets[segment_index];
```
fmoletta: As `self.relocate_segments()` is called outside this func, we can't be sure that its length is higher than len so indexing directly can panic

vm/src/vm/vm_core.rs
```diff
@@ -1028,19 +1028,40 @@ impl VirtualMachine {
...
+        let relocation_table = self
+            .relocation_table
+            .as_ref()
+            .ok_or(MemoryError::UnrelocatedMemory)?;
+
+        let relocate = |segment: (usize, Option<usize>)| -> (usize, Option<usize>) {
+            let (index, stop_ptr_offset) = segment;
+            (
+                relocation_table[index],
+                stop_ptr_offset.map(|s| relocation_table[index] + s),
```
fmoletta, line 1051: Use `get` instead of indexing directly

cairo-vm-cli/src/main.rs
```diff
@@ -112,12 +115,21 @@ fn run(args: impl Iterator<Item = String>) -> Result<(), Error> {
...
     };
-    let trace_enabled = args.trace_file.is_some();
+
+    if args.air_public_input.is_some() && !args.proof_mode {
+        let error = Args::command().error(
+            clap::ErrorKind::ArgumentConflict,
+            "--air_public_input can only be used in proof_mode.",
+        );
+        return Err(Error::Cli(error));
+    }
```
fmoletta, line 125: This error is not being shown when running the cli, must be due to: #1270

vm/src/vm/vm_core.rs
xqft: done!

cairo-vm-cli/src/main.rs
xqft: done!

vm/src/air_public_input.rs
xqft: done!

cairo-vm-cli/src/main.rs
xqft: sure! done

vm/src/air_public_input.rs
xqft: done!

vm/src/vm/vm_core.rs
xqft: done!

cairo-vm-cli/src/main.rs
xqft: Done!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * fix: change error returned when subtracting two `MaybeRelocatable`s to better reflect the cause [#1271](https://github.com/lambdaclass/cairo-rs/pull/1271)
 
 * fix: CLI error message when using --help [#1270](https://github.com/lambdaclass/cairo-rs/pull/1270)
+* feat: implement the `--air_public_input` flag to the runner for outputting public inputs into a file [#1268](https://github.com/lambdaclass/cairo-rs/pull/1268)
```
MegaRedHand, line 16: This should go in the "Upcoming Changes" section

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[error("The trace slice provided is empty")]
+    EmptyTrace,
+    #[error("The provided memory doesn't contain public address {0}")]
+    MemoryNotFound(usize),
+    #[error("Failed to interact with the file system")]
+    IO(#[from] std::io::Error),
+    #[error("Failed to (de)serialize data")]
+    Serde(#[from] serde_json::Error),
+    #[error("VM related error")]
+    VirtualMachine(#[from] VirtualMachineError),
```
fmoletta, line 102:
```suggestion
    #[error(transparent)]
    VirtualMachine(#[from] VirtualMachineError),
```
Lets just use the VirtualMachineError message

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[error("Trace related error")]
+    Trace(#[from] TraceError),
```
fmoletta, line 104:
```suggestion
    #[error(transparent)]
    Trace(#[from] TraceError),
```
Same here

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -775,6 +776,8 @@ impl CairoRunner {
             .segments
             .relocate_segments()
             .expect("compute_effective_sizes called but relocate_memory still returned error");
+        vm.relocation_table = Some(relocation_table.clone());
+
         if relocate_mem {
             if let Err(memory_error) = self.relocate_memory(vm, &relocation_table) {
                 return Err(TraceError::MemoryError(memory_error));
```
fmoletta, line 783: We can reorder this code to avoid cloning here 
```rust
let relocation_table = vm
            .segments
            .relocate_segments()
            .expect("compute_effective_sizes called but relocate_memory still returned error");

        if relocate_mem {
            if let Err(memory_error) = self.relocate_memory(vm, &relocation_table) {
                return Err(TraceError::MemoryError(memory_error));
            }
        }
        vm.relocate_trace(&relocation_table)?;
        vm.relocation_table = Some(relocation_table);
        Ok(())
```

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,105 @@
...
+        errors::{trace_errors::TraceError, vm_errors::VirtualMachineError},
+        trace::trace_entry::TraceEntry,
+    },
+};
+
+#[derive(Serialize, Debug)]
+pub struct PublicMemoryEntry {
+    address: usize,
+    page: usize,
+    value: Option<Felt252>,
```
fmoletta, line 19: When serializing to json, the inner felt representation is being used:
```json
{
      "address": 1,
      "page": 0,
      "value": {
        "value": {
          "val": [
            2147450879,
            67600385
          ]
        }
      }
    },
```
vs cairo-lang implementation using hexadecimal string representation:
```json
{
            "page": 0,
            "address": 1,
            "value": "0x40780017fff7fff"
        },
```
We coud add a `#[serde(serialize_with = "path")]` attribute to the `value` field in order to serialize it as a hex string

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#[derive(Serialize, Debug)]
+pub struct PublicInput<'a> {
+    layout: &'a str,
+    layout_params: Option<&'a CairoLayout>,
+    rc_min: Option<isize>,
+    rc_max: Option<isize>,
+    n_steps: usize,
+    memory_segments: HashMap<&'a str, (usize, Option<usize>)>,
```
fmoletta, line 29: We could add a #[serde(serialize_with = "path")] attribute to the `memory_segments` field to turn this:
```json
"memory_segments": {
    "range_check_builtin": [
      311638,
      314768
    ],
    "keccak_builtin": [
      622934,
      622942
    ],
    "poseidon_builtin": [
      639318,
      639327
    ],
```
Into something like this:
```json
"memory_segments": {
        "program": {
            "stop_ptr": 5,
            "begin_addr": 1
        },
        "execution": {
            "stop_ptr": 115030,
            "begin_addr": 1689
        },
        "output": {
            "stop_ptr": 115030,
            "begin_addr": 115030
        },
        "pedersen": {
            "stop_ptr": 115030,
            "begin_addr": 115030
        },
```
Without having to store extra strings

vm/src/vm/runners/cairo_runner.rs
xqft: Dome!

CHANGELOG.md
xqft: Done!

vm/src/air_public_input.rs
fmoletta: On the same tone, we should also add a comment on the `memory_segments` field indicating which tuple value is the `begin_addr` and which is the `stop_ptr` to make it more clear to library users

vm/src/vm/vm_core.rs
```diff
@@ -1024,6 +1026,45 @@ impl VirtualMachine {
...
+            .ok_or(MemoryError::UnrelocatedMemory)?;
+
+        let relocate = |segment: (usize, Option<usize>)| -> Result<(usize, Option<usize>), VirtualMachineError> {
+            let (index, stop_ptr_offset) = segment;
+            let base = relocation_table
+                .get(index)
+                .ok_or(VirtualMachineError::RelocationNotFound(index))?;
+            Ok((
+                *base,
+                stop_ptr_offset.map(|s| base + s),
```
fmoletta, line 1054: We could return a `Runner::Error(NoStopPointer)` here if stop_ptr_offset is None and avoid handling an Option in the following code. It doesn't make much sense to return an air_public_input with these values missing

vm/src/air_public_input.rs
xqft: Yeah I think we should go the way of cairo-lang and make a `MemoryAddress` struct (or something like that I don't remember the exact name) and replace all tuples with that, I'm not sure how many times tuples are used and if this would go out of the scope of the PR

vm/src/air_public_input.rs
xqft: done!

vm/src/air_public_input.rs
Oppen: You mean `Relocatable`?

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+}
+
+#[derive(Serialize, Debug)]
+pub struct PublicInput<'a> {
+    layout: &'a str,
+    layout_params: Option<&'a CairoLayout>,
+    rc_min: Option<isize>,
+    rc_max: Option<isize>,
```
Oppen, line 44: Does it make sense for these to be `Option`? In what scenario would they be `None`?

vm/src/air_public_input.rs
```diff
@@ -0,0 +1,122 @@
...
+    memory_segments: HashMap<&'a str, (usize, Option<usize>)>,
+    public_memory: Vec<PublicMemoryEntry>,
+}
+
+impl<'a> PublicInput<'a> {
+    pub fn new(
+        memory: &[Option<Felt252>],
+        layout: &'a str,
+        dyn_layout_params: Option<&'a CairoLayout>,
+        public_memory_addresses: &[(usize, &usize)],
```
Oppen, line 55: I find this type strange. Why is the second element of the tuple a reference?

vm/src/vm/runners/builtin_runner/bitwise.rs
```diff
@@ -166,8 +166,7 @@ impl BitwiseBuiltinRunner {
             self.stop_ptr = Some(stop_ptr);
             Ok(stop_pointer_addr)
         } else {
-            let stop_ptr = self.base;
-            self.stop_ptr = Some(stop_ptr);
+            self.stop_ptr = Some(0);
```
Oppen, line 169: Why does this change?

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -1102,6 +1107,32 @@ impl CairoRunner {
     pub fn get_program(&self) -> &Program {
         &self.program
     }
+
+    /// Return CairoRunner.layout
+    pub fn get_layout(&self) -> &CairoLayout {
+        &self.layout
+    }
```
Oppen, line 1114: Does this need to be public? Also, does this need to return the whole (internal) `CairoLayout` structure or is the name enough?

vm/src/vm/vm_memory/memory_segments.rs
```diff
@@ -213,6 +213,25 @@ impl MemorySegmentManager {
...
+    /// Returns a list of addresses of memory cells that constitute the public memory.
+    /// segment_offsets is the result of self.relocate_segments()
+    pub fn get_public_memory_addresses(&self, segment_offsets: &[usize]) -> Vec<(usize, &usize)> {
+        let len = self.num_segments().min(self.public_memory_offsets.len());
+        let mut addresses = Vec::with_capacity(len);
+
+        for (offsets, segment_start) in self
+            .public_memory_offsets
+            .values()
+            .zip(segment_offsets.iter())
```
Oppen, line 225: You can used `cloned()` on this iterator and just save yourself a lot of dealing with references:
```suggestion
            .zip(segment_offsets.iter().cloned())
```

Oppen (changes requested):
I'd like to discuss in depth why so many private structs now derive `Serialize` before green lighting this. That's quite an increase in API surface that is likely to haunt us tomorrow.

Any chance we have a quick huddle with you, @juanbono and @fmoletta tomorrow?

vm/src/air_public_input.rs
xqft: I was talking about [this](https://github.com/starkware-libs/cairo-lang/blob/3a32740d0a743af6f4bfe227d1b5ecab0fc2884b/src/starkware/cairo/lang/vm/utils.py#L32C1-L40)

vm/src/vm/vm_core.rs
xqft: Done!

vm/src/vm/runners/builtin_runner/bitwise.rs
xqft: It's a bugfix, I can link you to a discussion about it

vm/src/vm/runners/cairo_runner.rs
xqft: The whole structure, as it's used for writing the layout's parameters when it's a dynamic one.

vm/src/vm/vm_memory/memory_segments.rs
xqft: I'm not sure why is this a problem, though I could clone `page_id` to solve the weird `(usize, &usize)` return type that you mentioned in another review

vm/src/air_public_input.rs
xqft: To avoid a `.clone()` call, but if you prefer I can make that call and change this (as I mentioned in another review)

vm/src/vm/runners/cairo_runner.rs
Oppen: What about it being public? Could that be avoided?

vm/src/vm/vm_memory/memory_segments.rs
Oppen: References leak into your structures and make life harder in the long term (you end up having to add lifetime parameters in several places). We use them because copies are typically expensive, but when we're talking about `usize` it's actually cheaper to just copy the value.

vm/src/vm/runners/cairo_runner.rs
xqft: Nope it doesn't, thanks!

vm/src/air_public_input.rs
xqft: I was also cloning it for creating the `PublicInput` struct, I think I was going to remove the reference but forgot about it. Thank you! done.

vm/src/air_public_input.rs
xqft: Done!

vm/src/vm/vm_memory/memory_segments.rs
xqft: That's great to know! thanks, but would it be applicable in this case? I'm using references only for the sum operation (`segment_start + offset`), whose result will be moved into the tuple that I'm pushing to the vec. Then at the end I'm returning a vector of tuples of values, not references. Am I missing something here?

vm/src/air_public_input.rs
xqft: Done applying the solution above, but didn't replace all tuple instances with the new struct

Oppen (approved):

vm/src/vm/vm_memory/memory_segments.rs
xqft: No problem! thanks for the good info nevertheless

Oppen:
Codecov complaining a lot about coverage. Could you add some tests?

cairo-vm-cli/src/main.rs
```diff
@@ -160,16 +167,22 @@ fn run(args: impl Iterator<Item = String>) -> Result<(), Error> {
...
     match run(std::env::args()) {
-        Ok(()) => Ok(()),
-        Err(Error::Cli(_)) => {
-            Ok(()) // Exit with code 0 to avoid printing CLI error message
-        }
-        Err(error) => Err(error),
+        Err(Error::Cli(err)) => err.exit(),
+        other => other,
     }
 }
```
Oppen, line 187: I'm not sure I understand these changes or how they relate to the issue at hand. Could you explain?

xqft: Sure!
I'm not sure why, but as the comment says, the previous error handling made it so CLI errors were avoided, so if a malformed command was entered, the program wouldn't do anything, exit and don't even display an error message. Calling `err.exit()` instead fixes this.

The program exits with an error code (not a panic) when arguments are bad, so `fn test_main()` was failing because when calling `main()` directly arguments will be invalid. I'm not sure if there's a way for the test to handle the exiting case because the process is being killed. This is why `main()`'s code is replaced with `return Ok(());` when running that test, so it can never fail. The reason of existence of that test is just to fool codecov, `run()` is tested already.

pefontana (approved):

PR #1279, opened 2023-06-26
build(vm): remove thiserror form deps

tdelabro:
# Remove dep to thiserror

## Description

Rather than alternating between thiserror and thiserror-no-std depending on the feature "std", we can use thiserror-no-std/std.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* build: remove dependecy to `thiserror` (use `thiserror-no-std/std` instead)
+
 * chore: use LambdaWorks' implementation of bit operations for `Felt252` [#1291](https://github.com/lambdaclass/cairo-rs/pull/1291)
 
 #### [0.8.0] - 2023-6-26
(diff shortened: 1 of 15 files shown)


fmoletta (approved):
Thanks!

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * build: remove unused implicit features from cairo-vm
 
+* build: remove dependecy to `thiserror` (use `thiserror-no-std/std` instead)
```
MegaRedHand, line 7: I think this is on an old version of the changelog. We should merge main to this PR and move this again to the "Upcoming Changes" section.

MegaRedHand (approved):
🚀 🚀

vm/Cargo.toml
```diff
@@ -19,6 +19,7 @@ std = [
     "parse-hyperlinks/std",
     "felt/std",
     "dep:num-prime",
+    "thiserror-no-std/std",
```
Oppen, line 22: All of this makes me wonder, do we actually use anything that requires the `std` feature on `thiserror`? Did you try building without it?

Oppen (approved):
Looks great, thanks @tdelabro!

IMO it's ready to merge, but please move the changelog entry up as @MegaRedHand pointed out. If you have time, could you also check my question in the other comment? Not a blocker, it's mostly curiosity.

vm/Cargo.toml
tdelabro: In the future rust will move the error in `core`

CHANGELOG.md
tdelabro: I just fixed it :)

PR #1286, opened 2023-06-27
make `ProgramJson` serializable

tarrencev:
Resolves https://github.com/lambdaclass/cairo-rs/issues/1183

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -324,7 +324,7 @@ dependencies = [
  "cairo-lang-sierra-generator",
  "cairo-lang-syntax",
  "cairo-lang-utils",
- "clap 4.3.8",
+ "clap 4.3.6",
  "log",
  "salsa",
  "smol_str",
@@ -594,7 +594,7 @@ dependencies = [
  "cairo-lang-sierra-ap-change",
  "cairo-lang-sierra-gas",
  "cairo-lang-utils",
- "clap 4.3.8",
+ "clap 4.3.6",
  "indoc",
  "itertools",
  "log",
@@ -627,7 +627,7 @@ dependencies = [
  "cairo-lang-sierra-to-casm",
  "cairo-lang-syntax",
  "cairo-lang-utils",
- "clap 4.3.8",
+ "clap 4.3.6",
  "convert_case",
  "genco",
  "indoc",
@@ -812,9 +812,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.3.8"
+version = "4.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9394150f5b4273a1763355bd1c2ec54cc5a2593f790587bcd6b2c947cfa9211"
+checksum = "6320c6d1c98b6981da7bb2dcecbd0be9dc98d42165fa8326b21000f7dbfde6d0"
 dependencies = [
  "clap_builder",
  "clap_derive 4.3.2",
@@ -823,9 +823,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.3.8"
+version = "4.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a78fbdd3cc2914ddf37ba444114bc7765bbdcb55ec9cbe6fa054f0137400717"
+checksum = "2e53afce1efce6ed1f633cf0e57612fe51db54a1ee4fd8f8503d078fe02d69ae"
 dependencies = [
  "anstream",
  "anstyle",
@@ -856,7 +856,7 @@ dependencies = [
  "heck 0.4.1",
  "proc-macro2",
  "quote",
- "syn 2.0.22",
+ "syn 2.0.18",
 ]
 
 [[package]]
@@ -1249,7 +1249,7 @@ checksum = "89ca545a94061b6365f2c7355b4b32bd20df3ff95f02da9329b34ccc3bd6ee72"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.22",
+ "syn 2.0.18",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


Cargo.lock
Oppen, line 1: I'm not sure this change justifies changing the `Cargo.lock`. Why was is this here?

vm/src/serde/deserialize_program.rs
```diff
@@ -349,6 +352,22 @@ pub fn deserialize_felt_hex<'de, D: Deserializer<'de>>(d: D) -> Result<Felt252,
     d.deserialize_str(Felt252Visitor)
 }
 
+pub fn serialize_program_data<S: Serializer>(
+    v: &[MaybeRelocatable],
+    serializer: S,
+) -> Result<S::Ok, S::Error> {
+    let v = v
+        .iter()
+        .map(|val| match val.clone() {
```
Oppen, line 361: This `clone` doesn't seem necessary.

vm/src/serde/deserialize_program.rs
```diff
@@ -349,6 +352,22 @@ pub fn deserialize_felt_hex<'de, D: Deserializer<'de>>(d: D) -> Result<Felt252,
...
+            MaybeRelocatable::Int(value) => format!("0x{:x}", value.to_biguint()),
```
Oppen, line 362: We could leave this as is for now, but this hints to a need to implement `Display` for `Felt252`. @fmoletta WDYT?

Oppen (changes requested):
I'm not 100% convinced this is the right solution. That said, I'm willing to merge something like this mostly because it's harmless for the most part. `ProgramJson` is out of the hot path so it's less likely to change and break the serialization format.

That said, we need tests for this, at the very least unit tests but preferably integration tests too to see how the feature is used, etc.

vm/src/serde/deserialize_program.rs
fmoletta: There is a `Display` implementation for `Felt252`

MegaRedHand: I guess it's done this way because the `Display`/`Debug` impls produce a decimal representation instead of a hexadecimal one. This use case is equivalent to `value.to_str_radix(16)`

Oppen: Doesn't the `Display` trait receive a formatter? Or is that a different trait? Typically you'll get one and then the format string can be processed to pick the right representation.

kariy:
> That said, we need tests for this, at the very least unit tests but preferably integration tests too to see how the feature is used, etc.

Not sure how to write a meaningful test for this, we only need it to convert `Program` into bytes, because we exclusively use it to marshal between contract types coming from different crates... 

This is our use case for it:
https://github.com/dojoengine/dojo/blob/86f448c38cb21d7feb52d89cf48aea3b73648739/crates/katana/rpc/src/utils/contract.rs#L26

Oppen:
> Not sure how to write a meaningful test for this

I'd be happy enough with something with a logic like the following:
1. Write a JSON string that looks like what you expect.
2. Deserialize it into `ProgramJson`.
3. Serialize the result.
4. Deserialize again into `ProgramJson`.
5. Check both instances are equal.

The logic seems convoluted, but it'll save you from considering all the irrelevant differences that might come from formatting or field ordering if comparing the raw strings, while ensuring future changes remain compatible with that text representation.
Do note you can't rely on a hash of this data here to match between two different serializations, and that should probably be documented somewhere. That said, the team still needs to define a good policy regarding docs so I'm not gonna be pressing that yet.

> we only need it to convert Program into bytes, because we exclusively use it to marshal between contract types coming from different crates...

All API surface needs to consider a bit more than the special cases that point out a need, as it always come with some maintenance baggage. I don't intend to be a PITA, I'm just trying to make sure we don't create problems for downstream users and collaborators (including you)  in the future. Serialization format is part of that.

vm/src/serde/deserialize_program.rs
Oppen: Since @MegaRedHand created the issue at lambdaworks I'll mark this one resolved.

vm/src/serde/deserialize_program.rs
```diff
@@ -461,38 +461,41 @@ pub fn parse_program_json(
...
-    ProgramJson {
-        prime: program.prime().to_string(),
-        builtins: program.builtins.clone(),
-        data: program.shared_program_data.data.clone(),
-        identifiers: program.shared_program_data.identifiers.clone(),
-        hints: program.shared_program_data.hints.clone(),
-        attributes: program.shared_program_data.error_message_attributes.clone(),
-        debug_info: program
+impl From<Program> for ProgramJson {
+    fn from(program: Program) -> Self {
```
Oppen, line 465: Let's take a reference here. Since we're already cloning the contents there's no need to drop the old instance.

Oppen (commented):
Hey, sorry I didn't notice the new changes!
It looks ready to merge except for a last suggestion.

vm/src/serde/deserialize_program.rs
```diff
@@ -349,6 +352,22 @@ pub fn deserialize_felt_hex<'de, D: Deserializer<'de>>(d: D) -> Result<Felt252,
...
+pub fn serialize_program_data<S: Serializer>(
+    v: &[MaybeRelocatable],
+    serializer: S,
+) -> Result<S::Ok, S::Error> {
+    let v = v
+        .iter()
+        .map(|val| match val {
+            MaybeRelocatable::Int(value) => format!("0x{:x}", value.to_biguint()),
+            MaybeRelocatable::RelocatableValue(_) => {
+                panic!("Got unexpected relocatable value in program data")
```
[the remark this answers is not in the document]
Oppen: Thinking it a bit more, it's an internal error if this doesn't hold. Eventually I'd like to refactor it anyway to use the proper types, so I don't think this is a problem right now.

pefontana:
This implementation is done here #1458 
Thanks @tarrencev, for the contribution, we used your commits!

PR #1290, opened 2023-06-27
Add missing methods to lambdaworks implementation

MegaRedHand:
## Description

This PR adds the missing methods on the LambdaWorks `Felt252` implementation and marks them as `#[deprecated]` in both implementations.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


Oppen (approved):

juanbono (approved):
I'm ok with just returning an &[u8; 32] instead of vec. The main issue is that the name was modified and as a consequence it broke the interface.

So I think that another option can be to just change the name to the impl that returns a &[u8;32].

MegaRedHand:
Did some changes. The diff would be:
* Added `From<&BigUint> for Felt252` (this was already in the old felt, but wasn't migrated to the LW felt)
* Added `to_bytes_be` to the LW felt
* Both `to_signed_bytes_le` and `modpow` were marked as deprecated in the old felt (they won't be migrated to the new one)

PR #1298, opened 2023-06-29
feat: improve dependency installation experience

MegaRedHand:
## Description

This PR:
- updates the rust version in the list of dependencies (compilation in 1.66.1 is broken)
- adds the `install.sh` script, that detects the OS/distro and calls the corresponding installation script and `make deps`
- adds instructions on the README for running the script
- modifies the `test_install` workflow to use the script and to compile the whole project, to catch any missing dependencies
- disables the `test_install` workflow, so it runs only on pushes to main branch

## Checklist
- [X] Linked to Github Issue
- [ ] Unit tests added
- [X] Integration tests added.
- [X] This change requires new documentation.
  - [X] Documentation has been added/updated.


Oppen (approved):

Oppen:
A few questions:
1. Can we use the `setup-python` action instead? It has an option to cache pip deps.
2. Why can't we use `${HOME}` rather than explicitly spelling out the home path?

MegaRedHand:
1. I was trying to leave the Python installation to the script.
2. The action treats the `${HOME}` syntax as a literal part of a relative path.

It seems the _install-python_ action doesn't support Debian containers anyways 😞

Oppen (approved):

install-scripts/install-macos.sh
```diff
@@ -26,4 +26,3 @@ source ~/cairo_venv/bin/activate
 
 # Install cairo & its dependencies
 pip3 install -r requirements.txt
-
```
pefontana, line 29: I think we should install `pyenv install 3.9.15`

pefontana: Also, this isn't duplicated? Since then `install_macos` runs `make deps-macos`
But we can leave it duplicate because `make deps-macos` won't fail.

What we need to check is to add the `CFLAGS=-I/opt/homebrew/opt/gmp/include LDFLAGS=-L/opt/homebrew/opt/gmp/lib` before the pip install

pefontana (changes requested):

install-scripts/install-macos.sh
MegaRedHand: Good catch! Removed the whole Python install 👍

pefontana (approved):

PR #1308, opened 2023-07-03
feat: lambdaworks-felt flag for cli

Oppen:
Allow using lambdaworks for the CLI

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


pefontana (changes requested):
Just to double-check.
The Default compilation of `vm-cli` will still be with de old Felt252 no?
I think it is ok to add the feature but all are focus and the CI checks should be done with the old Felt252, since it is what will be released to production

Oppen:
> Just to double-check.
> The Default compilation of `vm-cli` will still be with de old Felt252 no?
> I think it is ok to add the feature but all are focus and the CI checks should be done with the old Felt252, since it is what will be released to production

The default remains the old `Felt252` with `num-bigint` backend, yes.

pefontana (approved):

Issue #1304, opened 2023-06-29
Check the Make sure all builds work GitHub Action workflow


pefontana:
Check the Make sure all builds work GitHub Action workflow
Now is taking significantly less time than before.
check that we are building the project in Wasm

PR #1312, opened 2023-07-03
fix: check all features on each crate

MegaRedHand:
Closes: #1304

## Description

`cargo check-all-features` seems to compile crates while using their default features (noticed after running it with `--verbose` and no compiled dependencies). This results in no compile failures reported when running it even if one combination doesn't really work (for example, _cairo-felt_ without any features).

## Checklist
- [X] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -121,6 +121,9 @@ jobs:
 
     - name: Set up cargo cache
       uses: Swatinem/rust-cache@v2
+      with:
+        cache-on-failure: true
+
     - name: Install cargo-all-features
       uses: taiki-e/install-action@v2
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -157,8 +160,26 @@ jobs:
...
+        cargo check-all-features
+
+    - name: Check all features (CLI)
+      run: |
+        cd cairo-vm-cli
+        cargo check-all-features
+    
+    - name: Check all features (workspace)
+      run: |
+        cargo check-all-features --workspace --all-targets
```
Oppen, line 182: Maybe a strategy matrix would work better here? I think you can use the `-p` flag.

Oppen (approved):

.github/workflows/rust.yml
MegaRedHand: I thought about that, but using a matrix wouldn't work here as the four steps share a lot of compiled artifacts. And for the `-p` flag, it works the same way as `--workspace` or no flag (it compiles the whole workspace).

pefontana (approved):

PR #1315, opened 2023-07-04
feat: add WASM example

MegaRedHand:
## Description

This PR moves the [cairo-vm-wasm](https://github.com/lambdaclass/cairo-rs-wasm/tree/main) example to this repo and updates it.

We can later improve it, maybe adding an input and passing it to the Cairo program.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


examples/wasm-demo/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "wasm-demo"
+description = "A demo using cairo-vm in a WASM environment"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+readme.workspace = true
+keywords.workspace = true
```
Oppen, line 9: It may be a good idea to set `publish = false`.

examples/wasm-demo/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "wasm-demo"
+description = "A demo using cairo-vm in a WASM environment"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+readme.workspace = true
```
Oppen, line 8: I see we have a README made specifically for the demo. Are we sure we don't want to point to that one?

examples/wasm-demo/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+keywords.workspace = true
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = ["console_error_panic_hook"]
+
+[dependencies]
+serde_json = "1.0.87"
```
Oppen, line 18: Could we use a workspace dep for this?

examples/wasm-demo/Cargo.toml
MegaRedHand: Yeah 😅

Oppen (approved):

examples/wasm-demo/Cargo.toml
MegaRedHand: You're right! I didn't think about that.

examples/wasm-demo/Cargo.toml
MegaRedHand: Sure!

fmoletta (approved):

Oppen:
@juanbono I see we're stuck waiting for a "Run tests" job. That doesn't exist, can we remove it? We want "QA / Run tests" instead in all its variants (I see some missing).

MegaRedHand:
We need to remove the _"Run tests (test)"_ required check and add:

- Run tests (test#1)
- Run tests (test#2)
- Run tests (lambdaworks-felt, test#1)
- Run tests (lambdaworks-felt, test#2)
- Run tests (lambdaworks-felt, test-no_std)
- Run tests (lambdaworks-felt, test-wasm)

MegaRedHand:
I'm leaving this here for any future reappearances of this problem (`Unhandled exception. System.IO.IOException: No space left on device`):

Ideally, we would enable some option of `llvm-cov` that doesn't produce intermediate artifacts (`.profraw` files). Alas, I didn't find any way to do this, so I assume it isn't possible.

We can work around this problem by:
- freeing disk space on the GitHub runners via [maximize-build-disk-space action](https://github.com/marketplace/actions/maximize-build-disk-space) (this is a hack),
- dividing the tests into more partitions and running them in parallel,
- dividing the tests into more partitions but running them sequentially, while cleaning `*.profraw` files in-between runs,

.github/workflows/rust.yml
```diff
@@ -184,17 +187,16 @@ jobs:
     - name: Check no-std
       run: |
         cd ensure-no_std
-        cargo build --no-default-features
-        cargo build
+        cargo check --no-default-features
```
pefontana, line 190: I think it is better to run cargo build instead of a check.

pefontana (approved):

PR #1318, opened 2023-07-07
Unpin sha minor versions

MauroToscano:
# Unpin sha minor versions

## Description

Unpin sha's minor version so users of the library can use their own, and if they use another library with a pinned version, the conflict can be resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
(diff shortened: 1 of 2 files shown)


pefontana (approved):

MegaRedHand (approved):

MauroToscano:
This is not needed, cargo can solve the issues with an update, as long as no dependency is asking for a version older than this one. Usually they are set with an =, which enfores at least that version, allowing for minor updates.

PR #1332, opened 2023-07-13
perf: change expensive usage of `ok_or` for `ok_or_else`

### LINKED ISSUE #1195: Audit usage of `.ok_or()`
There are places where we use `.ok_or()` to handle error cases where the error variant is expensive to create (of the top of my mind, many `.to_string()` calls). We need to replace those uses with `.ok_or_else()` instead, so that setup is no longer unconditional but executed only after an error has been found instead.

**Additional context**
Noticed while observing a __big__ performance hit when boxing some bigger error variants, which should instead lead to a boost for the happy path.

MegaRedHand:
Closes #1195

## Description

In some cases, we were using `ok_or` with error variants that needed heap allocations. This PR changes those to `ok_or_else`.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 #### Upcoming Changes
 
-* feat: add `arbitrary` feature to enable arbitrary derive in `Program` and `CairoRunConfig`
+* perf: changed `ok_or` usage for `ok_or_else` in expensive cases [#1332](https://github.com/lambdaclass/cairo-vm/pull/1332)
+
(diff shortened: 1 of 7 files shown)


Oppen (approved):
Beautiful!

Oppen:
It seems like there's been a regression for the set integration benchmark. I'll check the changes again to see if it makes sense, but maybe we'll need to trigger the workflow again.

MegaRedHand:
@Oppen Re-run it and it seems to be a false positive

fmoletta (approved):
Nice catch!

PR #1334, opened 2023-07-14
docs: document deserialization modules and update lib module's doc

MegaRedHand:
## Description

This PR updates the lib module's documentation, and adds module level documentation for the `Program` deserialization related modules. It also makes the `deserialization_utils` module private and deletes an unused error struct.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/serde/deserialize_utils.rs
+++ b/vm/src/serde/deserialize_utils.rs
@@ -1,7 +1,7 @@
 //! # Deserialization utils
 //!
 //! This module contains some helper functions used in [`Program`](crate::types::program::Program) deserialization.
-//! Namely, [`maybe_add_padding`] and [`take_until_unbalanced`].
+//! Namely, [`maybe_add_padding`] and [`parse_value`].
 //!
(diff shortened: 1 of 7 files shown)


vm/src/serde/deserialize_utils.rs
```diff
@@ -1,3 +1,10 @@
+//! # Deserialization utils
+//!
+//! This module contains some helper functions used in [`Program`](crate::types::program::Program) deserialization.
+//! Namely, [`maybe_add_padding`] and [`take_until_unbalanced`].
```
fmoletta, line 4: These are specific utils, and not something end users should be importing (as we might deprecate/change them according to our needs). Could we make them (and if we can the whole module) private?

MegaRedHand: Sure! Doing that I also found one of the errors isn't even being used 😅

pefontana (approved):

Oppen (approved):

PR #1337, opened 2023-07-17
chore(examples): remove `wee_alloc` from wasm-demo and ensure-no_std

MegaRedHand:
## Description

The `wee_alloc` crate is unmaintained. Even if we use it only for examples, it triggers security vulnerability warnings.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/ensure-no_std/Cargo.toml
+++ b/ensure-no_std/Cargo.toml
@@ -9,7 +9,7 @@ cairo-felt = { path = "../felt", default-features = false, features = [
 ] }
 cairo-vm = { path = "../vm", default-features = false }
 
-wee_alloc = "0.4.5"
+esp-alloc = "0.3.0"
 
 [features]
 default = ["lambdaworks-felt"]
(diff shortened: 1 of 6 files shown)


pefontana (approved):

ensure-no_std/Cargo.toml
```diff
@@ -9,7 +9,7 @@ cairo-felt = { path = "../felt", default-features = false, features = [
 ] }
 cairo-vm = { path = "../vm", default-features = false }
 
-wee_alloc = "0.4.5"
+esp-alloc = "0.3.0"
```
Oppen, line 12: Is this correct? The description on crates.io seems to suggest this is ESP only.

MegaRedHand: It seems to compile fine, so I would assume it is. As we use this for compilation testing, I just looked for a _no-std_ allocator. I also found [this one](https://crates.io/crates/alloc-no-stdlib), which seems more general.

Oppen: Hmmm OK, if it's just compilation testing and it compiles, it works for me.

Oppen (approved):

PR #1344, opened 2023-07-25
Stabilize fuzz program & json

Juan-M-V:
# Implement arbitrary by hand for some structs

## Description

This PR aims to increase the edge coverage of `fuzz_json` and `fuzz_program` by implementing arbitrary by hand in some of the internal structs like `Location`. Also words were added to the `dict.json` to include some common tokens present in `.json` programs, this last change will be temporary as a better strategy is developed

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


fuzzer/src/fuzz_json.rs
```diff
@@ -1,165 +1,23 @@
...
-    Ok(())
-}
 
 fn main() {
     loop {
-        fuzz!(|args: Args| {
-            let _ = run(args);
+        fuzz!(|data: (CairoRunConfig, &[u8])| {
+            let (cairo_run_config, program_json) = data;
+            let _ = cairo_run(
```
Juan-M-V, line 11: Using `cairo_run` instead of `cairo_run_fuzzed_program` due to the unlikeliness of the fuzzer creating an infinite loop and wanting to get as close to normal VM execution as possible

pefontana (approved):

vm/src/serde/deserialize_program.rs
```diff
@@ -152,7 +153,7 @@ pub struct Attribute {
     pub flow_tracking_data: Option<FlowTrackingData>,
 }
 
-#[cfg_attr(all(feature = "arbitrary", feature = "std"), derive(Arbitrary))]
+//#[cfg_attr(all(feature = "arbitrary", feature = "std"), derive(Arbitrary))]
```
Oppen, line 156: Maybe remove this comment.

vm/src/serde/deserialize_program.rs
```diff
@@ -163,6 +163,44 @@ pub struct Location {
...
+            },
+            parent_location: None,
+            start_line: 0,
+            start_col: 0,
+        });
+
+        while let Some(mut location) = locations.pop() {
+            location.parent_location = Some((Box::new(iter_location), String::arbitrary(u)?));
+            iter_location = location;
+        }
```
Oppen, line 197: This essentially converts the location vector into a linked list, right? Is this the intended behavior?

vm/src/types/program.rs
```diff
@@ -65,6 +64,30 @@ pub(crate) struct SharedProgramData {
     pub(crate) reference_manager: Vec<HintReference>,
 }
 
+#[cfg(all(feature = "arbitrary", feature = "std"))]
+impl<'a> Arbitrary<'a> for SharedProgramData {
+    /// Create an arbitary [`SharedProgramData`] using `flatten_hints` to generate `hints` and
+    /// `hints_ranges`
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let data = Vec::<MaybeRelocatable>::arbitrary(u)?;
```
Oppen, line 72: We should probably find a way to bias towards:
1. `Felt252`, as `Relocatable` should not be allowed (and indeed we should change it in the struct);
2. Values in the range `[0..2^63)`, so it's more likely to create instructions than random data that will be rejected.
Maybe this can be done with an alternate function taking weights for those?

Oppen (commented):
I left one question and a suggestion. Overall looks good, waiting on the answer for approval.

vm/src/serde/deserialize_program.rs
Juan-M-V: Yes, it is. Though I want to improve the `Location` generation, I mainly wanted arbitrary to stop crashing, as it currently does with recursive structures.

vm/src/types/program.rs
Juan-M-V: how about: https://github.com/lambdaclass/cairo-vm/pull/1344/commits/732f9f908e427e92a737ce0b721fdf81c73e1916
```rust
let mut data = Vec::new(); //arbitrary(u)?;
for _ in 0..usize::arbitrary(u)? {
    let instruction = u64::arbitrary(u)?;
    data.push(MaybeRelocatable::from(Felt252::from(instruction)));
    // Check if the Imm falg is on and add an immediate value if it is
    if instruction & 0x0004000000000000 != 0 {
        data.push(MaybeRelocatable::from(Felt252::arbitrary(u)?));
    }
}
```

Juan-M-V: I'm not too sure how to avoid a magic number `if instruction & 0x0004000000000000 != 0 {` while also reducing overhead for generating teh data vector

vm/src/types/program.rs
```diff
@@ -65,6 +64,39 @@ pub(crate) struct SharedProgramData {
...
+#[cfg(all(feature = "arbitrary", feature = "std"))]
+impl<'a> Arbitrary<'a> for SharedProgramData {
+    /// Create an arbitary [`SharedProgramData`] using `flatten_hints` to generate `hints` and
+    /// `hints_ranges`
+    fn arbitrary(u: &mut Unstructured<'a>) -> arbitrary::Result<Self> {
+        let mut data = Vec::new(); //arbitrary(u)?;
+        for _ in 0..usize::arbitrary(u)? {
+            let instruction = u64::arbitrary(u)?;
+            data.push(MaybeRelocatable::from(Felt252::from(instruction)));
+            // Check if the Imm falg is on and add an immediate value if it is
```
Oppen, line 76:
```suggestion
            // Check if the Imm flag is on and add an immediate value if it is
```

vm/src/types/program.rs
Oppen: I think that's a good solution.

Oppen: With a caveat: an interesting case to test is where the last instruction expects an immediate but there's none. So maybe something like:
```rust
let (mut i, len) = (0, u64::arbitrary(u)?);
while i < len {
    i += 1;
    // generate instruction
    if i < len && instruction & 0x0004000000000000 != 0 {
         // generate Felt
         i += 1;
    }
}
```

I think that should cover all interesting cases?

Juan-M-V: Similar code: https://github.com/lambdaclass/cairo-vm/pull/1344/commits/240726a40667f66730a08deb1a2b8582f1459c6e
I trust `for i in ...` more than `while`, less likely to forget the `i += 1`

Oppen (approved):

PR #1355, opened 2023-08-02
feat: make "arbitrary" feature enable `proptest` features

MegaRedHand:
## Description

This PR makes public the `arbitrary` module, with the `proptest::arbitrary::Arbitrary` implementation for `Felt252` along with some helper functions, via the *arbitrary* feature.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Upcoming Changes
 
-feat: add "proptest" feature, that makes public `proptest` related implementations [#1355](https://github.com/lambdaclass/cairo-vm/pull/1355)
+feat: make *arbitrary* feature also enable a `proptest::arbitrary::Arbitrary` implementation for `Felt252` [#1355](https://github.com/lambdaclass/cairo-vm/pull/1355)
 
 #### [0.8.5] - 2023-7-31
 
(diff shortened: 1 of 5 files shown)


pefontana (approved):

Oppen:
Shouldn't the `arbitrary` feature take care of that already?

MegaRedHand:
The `arbitrary` feature enables the `arbitrary` dependency, while this one enables the `proptest` dependency. Those are two different things. Albeit they both have an `Arbitrary` trait, they are different (it's really confusing).
Do you think that combining the two features would be ok?

Oppen:
Oh. Is this about the `proptest::Arbitrary` and the `quickcheck::Arbitrary`? I would suggest instead those features are either collapsed into one as you suggest or be something like `pt_arbitrary` and `qc_arbitrary` (maybe not abbreviated, however you prefer).

MegaRedHand:
I ended up merging the two flags and added a comment clarifying the difference between [**Arbitrary**](https://docs.rs/arbitrary/latest/arbitrary/trait.Arbitrary.html), and [_Arbitrary_](https://docs.rs/proptest/latest/proptest/arbitrary/trait.Arbitrary.html).

Oppen (approved):

PR #1356, opened 2023-08-02
Better json generation

Juan-M-V:
# Create `.json` programs that resemble real programs

## Description

This is an improvement upon `fuzz_json`, which currently creates json files by spewing completely random bytes. Now the fuzzer implements an internal `ProgramJson` struct with custom `serialize` and `arbitrary` implementations to guide the fuzzing. Already, these changes double the coverage reached by the fuzzer

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


.github/workflows/fuzzer.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     - name: Initializing fuzzer with previous inputs 
       run: |
         cd fuzzer
-        HFUZZ_RUN_ARGS="--dict=json.dict --run_time 10800 --timeout 60 -T" cargo hfuzz run fuzz_json
+        HFUZZ_RUN_ARGS="--dict=json.dict --run_time 10800 --timeout 60" cargo hfuzz run fuzz_json
```
Juan-M-V, line 44: Removed the `-T` flag so that timeouts aren't flagged as errors. This change is needed because
1. Sometimes running honggfuzz is just slower than the release or debug version of the program
2. There is currently no way to tell if the program hangs because of a correct infinite loop

fuzzer/src/fuzz_json.rs
```diff
@@ -1,23 +1,355 @@
...
+    "ecdsa",
+    "keccak",
+    "bitwise",
+    "ec_op",
+    "poseidon",
+    "segment_arena",
+];
+
+const HEX_SYMBOLS: [&str; 16] = [
+    "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f",
```
Oppen, line 29: I wonder if it would be interesting to mix with the uppercase version.

Oppen (approved):

fuzzer/Cargo.toml
```diff
@@ -17,6 +17,8 @@ honggfuzz = "0.5.55"
 libfuzzer-sys = "0.4"
 cairo-vm = { path = "../vm", features = ["arbitrary"] }
 cairo-felt = { path = "../felt", features = ["arbitrary"] }
+serde = { version = "1.0.180", features = ["derive"] }
+serde_json = "1.0.104"
```
pefontana, line 21: can we use the `serde_json = { workspace = true } `version?
same for the serde import

vm/src/cairo_run.rs
```diff
@@ -19,6 +19,7 @@ use arbitrary::{self, Arbitrary, Unstructured};
 
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 pub struct CairoRunConfig<'a> {
+    #[cfg_attr(feature = "arbitrary", arbitrary(value = "main"))]
```
pefontana, line 22: sorry @Juan-M-V 
I don't get why we are adding this

pefontana (changes requested):
Nice one @Juan-M-V !
I left you some comments

vm/src/cairo_run.rs
Juan-M-V: I wanted to just test the "main" entrypoint when generating `CairoRunConfig` with arbitrary. Otherwise most runs fail because they will look for a random string as their entrypoint

fuzzer/Cargo.toml
Juan-M-V: I have a commit ready to go with the updated `Cargo.toml`s to include the fuzzer directory into the list of workspaces. Let me know if I should push it

pefontana (approved):

fuzzer/Cargo.toml
pefontana: Thanks @Juan-M-V !
I think that for now, it is better to keep it aside from the workspace
I would like to know what you think about that?

PR #1360, opened 2023-08-03
Fix `VirtualMachineError::JumpRelNotInt` Display message

fmoletta:
Message says that Res can't be integer instead of Res can't be relocatable
Note: coverage label added as this change doesn't merit a changelog entry


pefontana (approved):

pefontana:
nice catch fede!

Oppen (approved):

PR #1370, opened 2023-08-11
Basic example to run Cairo 1 programs

entropidelic:
This PR shows the minimal steps required for running a Cairo 1 program, as it can be found in the StarkWare's [cairo](https://github.com/starkware-libs/cairo) repo.

It is done with a simple fibonacci Cairo 1 program. The sierra code can be read directly from the `fibonacci.sierra` file or compiled inside the example. 

To test it, just run `cargo run` inside the `cairo1-run` repo.


cairo1-run/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "cairo1-run"
+version = "0.1.0"
+edition = "2021"
+
+[workspace]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cairo-vm = "0.8.2"
```
pefontana, line 11: can we use the workspace version?

fmoletta (commented):
Is having the entire corelib necessary?
If it is, can we install it using makefile targets instead of having it in the repo? This will also mean less changes when the corelib is updated

cairo1-run/cairo1-programs/factorial.cairo
```diff
@@ -0,0 +1,14 @@
...
+    let result = factorial(n);
+    result
+}
+
+fn factorial(n: felt252) -> felt252 {
+    match n {
+        0 => 1,
+        _ => n * factorial(n - 1),
+    }
+}
```
fmoletta, line 14: We can move this programs to the top-level `cairo_programs` and make a `cairo_1_programs` folder inside it to keep all programs in the same directory

cairo1-run/cairo1-programs/fibonacci.cairo
```diff
@@ -0,0 +1,14 @@
...
+    let result = fib(1, 1, n);
+    result
+}
+
+fn fib(a: felt252, b: felt252, n: felt252) -> felt252 {
+    match n {
+        0 => a,
+        _ => fib(b, a + b, n - 1),
+    }
+}
```
fmoletta, line 14: Same here

fmoletta (commented):
Please add an example on how to use this new cli to the readme (similar to the cairo-vm-cli example)

cairo-vm-cli/Cargo.toml
```diff
@@ -8,7 +8,7 @@ readme.workspace = true
 keywords.workspace = true
 
 [dependencies]
-bincode = { version = "2.0.0-rc.3", tag = "v2.0.0-rc.3", git = "https://github.com/bincode-org/bincode.git" }
+bincode = { version = "2.0.0-rc.3" }
```
Oppen, line 11: Let's use `bincode.workspace = true` instead. That way we ensure no incompatibilities in the future.

cairo-vm-cli/src/main.rs
```diff
@@ -75,7 +75,7 @@ enum Error {
     PublicInput(#[from] PublicInputError),
 }
 
-struct FileWriter {
+pub struct FileWriter {
```
Oppen, line 78: This was ineffective IIRC, so I'd rather leave it out instead.

cairo1-run/Cargo.toml
Oppen: Yes please 🙏

cairo1-run/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+cairo-lang-compiler = { git = "https://github.com/starkware-libs/cairo" }
+cairo-lang-sierra = { git = "https://github.com/starkware-libs/cairo" }
+cairo-lang-runner = { git = "https://github.com/starkware-libs/cairo" }
+cairo-lang-utils = { git = "https://github.com/starkware-libs/cairo" }
+itertools = "0.11.0"
+clap = { version = "4.3.10", features = ["derive"] }
+thiserror = { version = "1.0.40" }
+bincode.workspace = true
+assert_matches = "1.5.0"
+rstest = "0.17.0"
```
Oppen, line 20: And use the workspace version as much as possible in general.
We should also separate dev-dependencies from regular dependencies. End users don't need them.

cairo1-run/Makefile
```diff
@@ -0,0 +1,2 @@
+make test:
+	cargo test
```
Oppen, line 2: We missed the newline. I wonder if we can configure the repo globally to insert them.

cairo-vm-cli/Cargo.toml
mmsc2: done

pefontana:
The feature of running cairo 1 Programs with proof mode, will be implement directly in the prover repo
https://github.com/lambdaclass/lambdaworks_stark_platinum

cairo1-run/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+
+fn run(args: impl Iterator<Item = String>) -> Result<Vec<MaybeRelocatable>, Error> {
+    let args = Args::try_parse_from(args)?;
+
+    let compiler_config = CompilerConfig {
+        replace_ids: true,
+        ..CompilerConfig::default()
+    };
+    let sierra_program =
+        (*compile_cairo_project_at_path(&args.filename, compiler_config).unwrap()).clone();
```
pefontana, line 144: Can we handle all these unwraps()?

fmoletta: sure

Cargo.toml
```diff
@@ -69,6 +70,12 @@ bitvec = { version = "1", default-features = false, features = ["alloc"] }
 cairo-lang-starknet = { version = "2.1.0-rc2", default-features = false }
 cairo-lang-casm = { version = "2.1.0-rc2", default-features = false }
 
+cairo-lang-compiler = { version = "2.1.0-rc2", default-features = false }
```
juanbono, line 73: here you pin the cairo compiler crates to 2.1.0-rc2 but then you compile the contracts with cairo 2.2.0, this can lead to errors

cairo1-run/Makefile
```diff
@@ -0,0 +1,33 @@
...
+	cargo run --release $< --memory_file $@ 
+
+CAIRO_1_PROGRAMS=$(wildcard ../cairo_programs/cairo-1-programs/*.cairo)
+TRACES:=$(patsubst $(CAIRO_1_FOLDER)/%.cairo, $(CAIRO_1_FOLDER)/%.trace, $(CAIRO_1_PROGRAMS))
+MEMORY:=$(patsubst $(CAIRO_1_FOLDER)/%.cairo, $(CAIRO_1_FOLDER)/%.memory, $(CAIRO_1_PROGRAMS))
+
+deps:
+	git clone https://github.com/starkware-libs/cairo.git \
+	&& cd cairo \
+	&& git checkout v2.2.0 \
```
juanbono, line 18: here is where the 2.2.0 version of the compiler is downloaded

juanbono (approved):

Cargo.toml
fmoletta: This is actually cargo working in our favour, as depsite the version being pinned to "2.1.0-rc2", the "2.2.0" version is the one downloaded. But ill change it to avoid confusions

pefontana (approved):

Issue #1364, opened 2023-08-08
`VirtualMachine.compute_dst_deductions` and `VirtualMachine.deduce_dst` share the same behaviour


fmoletta:
We should remove one of them

PR #1382, opened 2023-08-17
Remove duplicated code is dst deduction

fmoletta:
Combines `compute_dst_deductions` & `deduce_dst` into `deduce_dst`
Solves #1364 
No changelog update needed as this is an internal refactor

### DIFF
--- a/vm/src/vm/vm_core.rs
+++ b/vm/src/vm/vm_core.rs
@@ -318,9 +318,9 @@ impl VirtualMachine {
         instruction: &Instruction,
         res: &Option<MaybeRelocatable>,
     ) -> Result<MaybeRelocatable, VirtualMachineError> {
-        let dst = match instruction.opcode {
-            Opcode::AssertEq if res.is_some() => res.clone().unwrap(),
-            Opcode::Call => MaybeRelocatable::from(self.run_context.get_fp()),
+        let dst = match (instruction.opcode, res) {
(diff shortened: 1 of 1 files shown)


pefontana (approved):

vm/src/vm/vm_core.rs
```diff
@@ -316,13 +316,14 @@ impl VirtualMachine {
...
-            Opcode::Call => Some(self.get_fp().into()),
-            _ => None,
-        }
+        res: &Option<MaybeRelocatable>,
+    ) -> Result<MaybeRelocatable, VirtualMachineError> {
+        let dst = match instruction.opcode {
+            Opcode::AssertEq if res.is_some() => res.clone().unwrap(),
+            Opcode::Call => MaybeRelocatable::from(self.run_context.get_fp()),
+            _ => return Err(VirtualMachineError::NoDst),
+        };
```
Oppen, line 325: Maybe we could avoid the `unwrap` if we match both arguments at the same time:
```suggestion
        let dst = match (instruction.opcode, res) {
            (Opcode::AssertEq, Some(res)) => res.clone(),
            (Opcode::Call, _) => MaybeRelocatable::from(self.run_context.get_fp()),
            _ => return Err(VirtualMachineError::NoDst),
        };
```

Oppen (approved):

PR #1390, opened 2023-08-18
Fix `cairo1` & `cairo2` builds in Makefile

### LINKED ISSUE #1389: Makefile target `build-cairo-2-compiler-macos` depends on a directory that is never built
The target `build-cairo-2-compiler-macos` checks if the directory `cairo2-macos` exists before running its rule, but its rule creates the `cairo2` directory. This leads to makefile errors when running the target when the `cairo2 ` directory exists, as the `cairo2-macos` directory is never created but its rule fails as the `cairo2` directory already exists.

fmoletta:
In order to install versions 1 & 2 of the cairo compiler for both macos and not macos systems, a rule was used to check weather the directory existed and build it. As two different behaviours where expected when building for macos and not for macos, two rules were added, one for the `cairoX` directory and one for `cairoX-macos` directory, but both of them built the `cairoX` directory. This lead to makefile erros when building for macos, as the rule would check for the `cairoX-macos` directory which didn't exist and then fail when the `cairoX` directory did exist. 
This PR uses conditionals to check weather the directory `cairoX` exists, solving this issue.
Closes #1389 
Also contains fix from #1388

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,24 +152,22 @@ $(CAIRO_2_CONTRACTS_TEST_DIR)/%.casm: $(CAIRO_2_CONTRACTS_TEST_DIR)/%.sierra
 # Setup Cairo 2 Compiler
 # ======================
 
-cairo-repo-2-dir = cairo2
-cairo-repo-2-dir-macos = cairo2-macos
+CAIRO_2_REPO_DIR = cairo2
+CAIRO_2_VERSION = 2.1.0-rc1
 
-build-cairo-2-compiler-macos: | $(cairo-repo-2-dir-macos)
+build-cairo-2-compiler-macos:
+	@if [ ! -d "$(CAIRO_2_REPO_DIR)" ]; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -115,22 +115,22 @@ $(CAIRO_1_CONTRACTS_TEST_DIR)/%.casm: $(CAIRO_1_CONTRACTS_TEST_DIR)/%.sierra
...
-	&& tar -xzvf cairo-1.1.1.tar \
-	&& mv cairo/ cairo1/
-
-build-cairo-1-compiler: | $(cairo-repo-1-dir)
-
-$(cairo-repo-1-dir):
-	curl -L -o cairo-1.1.1.tar https://github.com/starkware-libs/cairo/releases/download/v1.1.1/release-x86_64-unknown-linux-musl.tar.gz \
-	&& tar -xzvf cairo-1.1.1.tar \
-	&& mv cairo/ cairo1/
+CAIRO_2_REPO_DIR = cairo1
```
Oppen, line 118: I think there's a typo here:
```suggestion
CAIRO_1_REPO_DIR = cairo1
```

Oppen (approved):

pefontana (approved):

PR #1393, opened 2023-08-23
Don't build wasm-demo in `build` target +  ci job to run the wasm demo

### LINKED ISSUE #1392: Project fails to build due to `wasm-demo` crate
**Describe the bug**
The wasm-demo crate imports a compiled json file using `include_bytes` macro, causing its build to fail if the file is not present. This causes the `build` target to fail.

**To Reproduce**
1- Clone the repo
2- Run `make deps`/`make deps-macos` (Optional)
3- Run `make build`
4- See error

```bash
cargo build --release
   Compiling wasm-demo v0.8.6 (/Users/federica/cairo-vm/examples/wasm-demo)
error: couldn't read examples/wasm-demo/src/./array_sum.json: No such file or directory (os error 2)
  --> examples/wasm-demo/src/lib.rs:30:33
   |
30 |     const PROGRAM_JSON: &[u8] = include_bytes!("./array_sum.json");
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)
```

**Expected behavior**
Running `make build` should build the repo without failing

commit 2d214f760e1c966954ed64ce2db736051977f74b
(After v0.8.6)

**Additional context**
Compiling cairo files shouldn't be a prerequisite to build the project

fmoletta:
Aims to fix #1392 .

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


Makefile
```diff
@@ -310,3 +304,10 @@ fuzzer-deps:
 run-cairo-compiled-fuzzer:
 	cd fuzzer
 	cargo +nightly fuzz run --fuzz-dir . cairo_compiled_programs_fuzzer
+
+wasm-demo:
+	cd examples/wasm-demo; \
+	cairo-compile src/array_sum.cairo --no_debug_info --output src/array_sum.json && \
+	wasm-pack build --target=web && \
+	npm install live-server && \
+	npx live-server
```
Oppen, line 313: These are new deps AFAICT. We should add them both to our deps targets/scripts and to the docs.

fmoletta: These dependencies are only needed for the wasm-demo, which is not an integral part of the project, so I wouldn't force its dependencies for any user of the project (the main purpose of this PR is to separate it's build from the project's build). There is a separate README with instructions on how to build & run the demo, and its dependencies, so we could remove this target altogether

juanbono (approved):

Cargo.toml
```diff
@@ -6,6 +6,11 @@ members = [
     "hint_accountant",
     "examples/wasm-demo",
 ]
+default-members = [
```
pefontana, line 9: Just to check @fmoletta 
Why are we adding this default-members?

pefontana (approved):
LGTM!!

Cargo.toml
fmoletta: The problem we want to fix is that the build fails due to the wasm-demo needing the compiled cairo files in order to build.
Someone that wants to build the cairo-vm crate shouldn't have to compile a cairo file in order to build a demo for the vm crate to build.
Therefore, by setting the default members we can build the core crates (vm, cli & felt) with cargo/make build without building the wasm demo and running into compilation errors

PR #1399, opened 2023-08-24
refactor: Use a generic backend for  BigIntN & UintN structs used for hints

### LINKED ISSUE #1048: Reduce duplicate versions of math hints
Some functions have multiple versions, each one only using a different `UintNNN` wrapper to interact with the VM memory. An example are the `u384_get_square_root` and `u256_get_square_root` functions.

This could be solved by extracting the common interface into a trait, and using generics to write these hints only one time.

An alternative would be to make a generic `UintNNN` struct, but that way we wouldn't be able to add custom "member X doesn't exist" for each variant.

Another alternative would be to add all of these variants into an enum.

fmoletta:
Multiple structs to represent big integers are used in hints, which have the same behaviour and a similar structure. This PR aims to make a generic backend for all of these types, removing duplicated code.
Closes #1048 
This is an internal refactor, with no changes to the code's public api

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/.github/workflows/hint_accountant.yml
+++ b/.github/workflows/hint_accountant.yml
@@ -17,6 +17,8 @@ jobs:
       uses: Swatinem/rust-cache@v2
     - name: Checkout
       uses: actions/checkout@v3
+    - name: Check Build
+      run: cargo build -p hint_accountant
     - name: Run the hint accounting script
       run: cargo r -p hint_accountant | tee comment.md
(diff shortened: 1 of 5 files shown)


pefontana (approved):

pefontana:
Nice refactor!
Love it

juanbono (approved):

PR #1402, opened 2023-08-24
CI: Add fresh_run daily workflow

fmoletta:
Adds a workflow that runs the deps, build, and test makefile targets without using any cached builds/files. The aim of this workflow is to simulate what a new user does once they download the repo.
As this workflow doesn't use any cached data it will take some time to run, so it should be run daily and not in each PR

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


Juan-M-V (approved):

pefontana (approved):
Nice one @fmoletta !

PR #1423, opened 2023-08-28
Use cairo-lang whitelists for the hint accountant + update cairo-lang version

fmoletta:
Pull whitelists from cairo-lang latest instead of using own lists

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


hint_accountant/src/main.rs
```diff
@@ -46,6 +31,21 @@ struct Whitelist {
 }
 
 fn run() {
+    // We use the files in the cairo-lang repo, cloned from the latest version
+    let whitelist_paths = fs::read_dir("../cairo-lang/src/starkware/starknet/security/whitelists")
+        .expect("Failed to read whitelist directory from cairo-lang, did you forget to clone it?");
```
pefontana, line 36: lets add this tip in case we want to run it locally

```suggestion
        .expect("Failed to read whitelist directory from cairo-lang, did you forget to clone it? \ 
        run https://github.com/starkware-libs/cairo-lang.git");
```

Not sure if the suggestion will work :P

pefontana (changes requested):

hint_accountant/src/main.rs
fmoletta: I wouldn't add the command there as we don't know where we are running from and the relative path may be wrong (also cloning cairo-lang is pretty self-explainatory). We can add a makefile target that clones the repo and runs the hint accountant if we want to do the comparison locally.

pefontana: Nice, let add the makefile command

Makefile
```diff
@@ -308,10 +308,19 @@ clean:
...
 run-cairo-compiled-fuzzer:
 	cd fuzzer
 	cargo +nightly fuzz run --fuzz-dir . cairo_compiled_programs_fuzzer
+
+CAIRO_LANG_REPO_DIR: cairo-lang
+
+$(CAIRO_LANG_REPO_DIR):
+	git clone --depth=1 https://github.com/starkware-libs/cairo-lang
+
+hint-accountant: $(CAIRO_LANG_REPO_DIR)
```
pefontana, line 325: This command is returning me 
```
thread 'main' panicked at 'Failed to read whitelist directory from cairo-lang, did you forget to clone it?
: Os { code: 2, kind: NotFound, message: "No such file or directory" }', hint_accountant/src/main.rs:37:57
```

Also, can we add the`cairo-lang/` folder to the .gitignore?

fmoletta: Fixed and added

pefontana (approved):

juanbono (approved):

PR #1442, opened 2023-09-11
perf: share constants between `Program` instances

Oppen:
This reduces `Program` cloning for building the `CairoRunner` by 86%
according to our Criterion benchmarks.
This is a significant cost for running many short-lived programs such
as most Starknet transactions.
No negative impact in runtime was observed.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* perf: share `constants` field across `Program` instances [#1442](https://github.com/lambdaclass/cairo-vm/pull/1442)
+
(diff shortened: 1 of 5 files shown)


Oppen:
Interestingly this shows a slowdown. I guess I'll have to run the benchmarks again in the server and my workstation just in case. Converting to draft in the meantime.

gabrielbosio:
#2223 would make this PR unnecessary. If the former does not get merged, updating this PR with main would take an amount of effort bigger than starting implementing these changes in another PR.

PR #1469, opened 2023-10-30
chore: bump starknet-crypto to v0.6.1

xJonathanLEI:
# Update `starknet-crypto`

## Description

The new version contains a bug fix on ECDSA signature verification. Previously, the `s` range was incorrectly enforced to be lower than EC order instead of element bound. While this did not allow invalid signatures to be used, it was theoretically possible (1 in 2^48 signatures) to have a signature that passes `verify` yet cannot be proven (due to being out of element bound).

This wouldn't be an issue though, as long as other parts of the VM already enforce the range, but I'm not sure if it's the case. But in any case it wouldn't be a bad thing to upgrade anyways.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* chore: bump starknet-crypto to v0.6.1 [#1469](https://github.com/lambdaclass/cairo-vm/pull/1469)
+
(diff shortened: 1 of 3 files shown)


xJonathanLEI:
Okay so I need to update changelog.

Oppen (approved):

PR #1488, opened 2023-11-19
Proof mode, second draft

MauroToscano:
# Proof mode, second draft

## Description

Add proof mode for cairo1 inside the VM. 

TODO:
- [ ] Solve issue with gas builtin pointer when it's between other built ins
- [ ] Add SegmentArena support for dictionaries


MauroToscano:
Final version done in #1517

PR #1491, opened 2023-11-23
[WIP] Add `HintProcessor::execute_hint_extensive` + use a HashMap with Relocatable index for hint ranges

fmoletta:
This PR proposes a solution to the hint-relates problems faced when running the starknet os:

-  Hints can be associated to pc's that are not in the segment 0
- Hints can load more hints during their execution

The proposed solution for each problem is:

- Use a HashMap indexed by a Relocatable pc value for hint ranges
- Extend the `HintProcessor` api with an `execute_hint_extensive` method which can return a map of extra hints to be loaded after the hint is executed. This new method comes with a default implementation that just calls `execute_hint` so that this change may not affect current implementors of the trait


vm/src/hint_processor/hint_processor_definition.rs
```diff
@@ -52,8 +49,27 @@ pub trait HintProcessorLogic {
...
+        constants: &HashMap<String, Felt252>,
+    ) -> Result<Option<HintExtension>, HintError> {
+        self.execute_hint(vm, exec_scopes, hint_data, constants)?;
+        Ok(None)
+    }
 }
 
+// A map of hints that can be used to extend the current map of hints for the vm run
+// The map matches the pc at which the hints should be executed to a vec of compiled hints (Outputed by HintProcessor::CompileHint)
+pub type HintExtension = Box<HashMap<Relocatable, Vec<Box<dyn Any>>>>;
```
Oppen, line 71: I think we can avoid the outer `Box`. The `HashMap` already stores its contents on the heap and will be moved.
```suggestion
pub type HintExtension = HashMap<Relocatable, Vec<Box<dyn Any>>>;
```

vm/src/vm/vm_core.rs
```diff
@@ -443,15 +445,36 @@ impl VirtualMachine {
...
+                let res = hint_processor
+                    .execute_hint_extensive(
+                        self,
+                        exec_scopes,
+                        hint_datas.get(idx).ok_or(VirtualMachineError::Unexpected)?,
+                        constants,
+                    )
+                    .map_err(|err| VirtualMachineError::Hint(Box::new((idx - s, err))))?;
+                // Check if the hint executed adds new hints, and update the hint_ranges & hint_datas accordingly
+                if let Some(hint_extension) = res {
```
Oppen, line 469: Maybe we can return an empty hashmap by default instead of an `Option<_>` so the code is slightly simpler.

fmoletta: sure

Oppen (approved):

pefontana (approved):

PR #1492, opened 2023-11-23
Partially Revert `Optimize trace relocation #906`

fmoletta:
When running the starknet os, the pc can take values outside of the program segment, which conflicts with our trace optimization, that takes advantage of the trace pc always being on segment 0.
Partially reverts changes made by #906

### DIFF
--- a/vm/src/air_public_input.rs
+++ b/vm/src/air_public_input.rs
@@ -10,7 +10,7 @@ use crate::{
     types::layout::CairoLayout,
     vm::{
         errors::{trace_errors::TraceError, vm_errors::VirtualMachineError},
-        trace::trace_entry::TraceEntry,
+        trace::trace_entry::RelocatedTraceEntry,
     },
 };
 
@@ -73,7 +73,7 @@ impl<'a> PublicInput<'a> {
         dyn_layout_params: Option<&'a CairoLayout>,
         public_memory_addresses: &[(usize, usize)],
         memory_segment_addresses: HashMap<&'static str, (usize, usize)>,
-        trace: &[TraceEntry],
+        trace: &[RelocatedTraceEntry],
         rc_limits: (isize, isize),
     ) -> Result<Self, PublicInputError> {
         let memory_entry =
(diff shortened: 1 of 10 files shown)


vm/src/air_public_input.rs
```diff
@@ -73,7 +73,7 @@ impl<'a> PublicInput<'a> {
         dyn_layout_params: Option<&'a CairoLayout>,
         public_memory_addresses: &[(usize, usize)],
         memory_segment_addresses: HashMap<&'static str, (usize, usize)>,
-        trace: &[TraceEntry],
+        trace: &[RelocatedTraceEntry],
```
Oppen, line 76: Breaking.

vm/src/cairo_run.rs
```diff
@@ -158,7 +158,7 @@ pub struct EncodeTraceError(usize, bincode::error::EncodeError);
 /// Bincode encodes to little endian by default and each trace entry is composed of
 /// 3 usize values that are padded to always reach 64 bit size.
 pub fn write_encoded_trace(
-    relocated_trace: &[crate::vm::trace::trace_entry::TraceEntry],
+    relocated_trace: &[crate::vm::trace::trace_entry::RelocatedTraceEntry],
```
Oppen, line 161: Breaking.

Oppen (approved):
LGTM but we need the following additions to the changelog:
– List the breaking changes;
– Optional, but IMO useful, mention the actual feature we're working to achieve, which is running code from arbitrary segments.

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -756,12 +761,39 @@ impl CairoRunner {
         Ok(())
     }
 
+    ///Relocates the VM's trace, turning relocatable registers to numbered ones
+    fn relocate_trace(
```
pefontana, line 765: Should we make this functions public?
Since the deleted method `pub fn relocate_trace(&mut self, relocation_table: &[usize]) -> Result<(), TraceError>` was public?

fmoletta: we can

pefontana (approved):

PR #1494, opened 2023-11-24
Add a test that runs a cairo 0 contract from a cairo program

fmoletta:
Depends & includes changes from #1492 #1493 #1491 
Acts as an example on how to run a cairo 0 contract from a program


cairo_programs/noretrocompat/starknet_os_deprecated_cc.cairo
Oppen, line 1: Shouldn't this be retrocompatible if the original VM supports it?

vm/src/tests/run_deprecated_contract_class_simplified.rs
```diff
@@ -0,0 +1,419 @@
...
+        vm_core::VirtualMachine,
+    },
+};
+
+struct SimplifiedOsHintProcessor {
+    builtin_hint_processor: BuiltinHintProcessor,
+    run_resources: RunResources,
+}
+
+impl ResourceTracker for SimplifiedOsHintProcessor {
```
pefontana, line 48: I write this comment just to remember,
We should document this trait since developers are going to implement it

pefontana (approved):
Amaizing work @fmoletta !!!

PR #1496, opened 2023-11-26
feat: add cairo_run_program function that takes a Program as an arg

ElijahVlasov:
# Add ```cairo_run_program``` function to ```cairo_runner```.

## Description

It seems convenient to have an API function that runs a cairo program from a ```Program``` struct, not a serialized JSON representation.  For instance, in my use-case the program comes from a compressed binary representation, not from a json, so don't want to copy the content of ```cairo_run``` function in my project for that.

If this is useful for someone barring me please consider this tiny PR :)


fmoletta (approved):

Oppen:
Also, clippy is complaining about taking a reference to `program`, which is now already a reference.

ElijahVlasov:
> Also, clippy is complaining about taking a reference to `program`, which is now already a reference.

Should be fine now

PR #1503, opened 2023-12-04
perf: Add `extensive_hints` feature to prevent performance regression for the common use-case

fmoletta:
Gates behaviour added by #1491 under `extensive_hints` feature to prevent performance regressions on most use cases


Oppen (approved):

vm/src/vm/vm_core.rs
```diff
@@ -444,33 +445,43 @@ impl VirtualMachine {
...
             }
         }
+        #[cfg(not(feature = "extensive_hints"))]
+        for (hint_index, hint_data) in hint_datas.iter().enumerate() {
+            hint_processor
+                .execute_hint(self, exec_scopes, hint_data, constants)
+                .map_err(|err| VirtualMachineError::Hint(Box::new((hint_index, err))))?
+        }
         Ok(())
     }
```
Oppen, line 486: I think if we just split to two different implementations it'll be easier to read. Like so:
```rust
    pub fn step_hint(
        &mut self,
        hint_processor: &mut dyn HintProcessor,
        exec_scopes: &mut ExecutionScopes,
        hint_datas: &mut Vec<Box<dyn Any>>,
        hint_ranges: &mut HashMap<Relocatable, HintRange>,
        hint_datas: &mut Vec<Box<dyn Any>>,
        hint_ranges: &mut HashMap<Relocatable, HintRange>,
        constants: &HashMap<String, Felt252>,
        // ...
    }

    #[cfg(not(feature = "extensive_hints"))]
    pub fn step_hint(
        &mut self,
        hint_processor: &mut dyn HintProcessor,
        exec_scopes: &mut ExecutionScopes,
        hint_datas: &mut Vec<Box<dyn Any>>,
        hint_ranges: &mut HashMap<Relocatable, HintRange>,
        hint_datas: &[Box<dyn Any>],
        constants: &HashMap<String, Felt252>,
        // ...
    }
```

Oppen:
I suggest we first try to see if we can alleviate the hit with a faster hash. It may not be as fast as the array approach, but it will be significantly easier to maintain than having two separate cases and could come quite close.
Popular alternatives:
https://crates.io/crates/ahash
https://crates.io/crates/ritehash
https://crates.io/crates/fnv_rs

We can also make the input to the hash much smaller by simply converting the `Relocatable` to its compact form before hashing, for example:
```rust
impl Hash for Relocatable {
    fn hash<H: Hasher>(&self, state: &mut H) {
        (self.segment_index.unsigned_abs() as u64 << 48 | self.offset as u64).hash(state);
    }
}
```
This halves the input bytes to the hash for nearly zero effort, which is typically about linear in CPU time. I used `unsigned_abs` for convenience, but of course this would cause collisions for temporary segments sharing the same offset. I'm not sure we store temporary segments on any `HashMap`, but if we do we could use the MSB to store the sign if we need.

vm/Cargo.toml
```diff
@@ -28,6 +28,9 @@ cairo-1-hints = [
 ]
 arbitrary = ["dep:arbitrary", "felt/arbitrary", "felt/std", "std"]
 lambdaworks-felt = ["felt/lambdaworks-felt"]
+# Allows extending the set of hints for the current vm run from within a hint.
+# For a usage example checkout vm/src/tests/run_deprecated_contract_class_simplified.rs
+extensive_hints = []
```
pefontana, line 33: Can we also add the feature to the cairo-vm-cli?
In `cairo-vm-cli/Cargo.toml`
`extensive_hints = ["cairo-vm/extensive_hints"]`

Oppen:
Just tried the suggestions in my previous comment on the draft PR #1505. The speedup is quite similar with ahash. I can clean that one up so we don't add the extra feature flag, it should _Just Work (TM)_ for both cases.

pefontana (approved):

pefontana (approved):

PR #1517, opened 2023-12-18
Cairo1 proof mode

MauroToscano:
# Cairo1 proof mode

## Description

This PR adds an initial working version of Cairo1 proof mode

- Adds a proof mode header, working very similarly to the Cairo0 proof mode header to the code
- Makes some small changes to SW Cairo header, so the pointers to the segment are properly aligned now that there are two extra instructions
- Adds a RunnerMode to select from ExecutionMode, ProofModeCanonical, and ProofModeCairo1
- Adds a test for dictionaries in cairo1, since they work a bit different in the headers and wasn't being tested
- Reduces the number of iterations of the sample program to 100 so Cairo1 tests run faster

In the future information should be added so [fp - 2] = fp can be enforced


pefontana (approved):

fmoletta (commented):
We should add a command line argument to set the run to proof_mode instead of just running proof_mode by default

pefontana:
> should add a command line argument to set the run to proof_mode instead of just running proof_mode by default

Yes @fmoletta, that is the plan
We talked with Mauro, first we can merge this to run always on proof mode, and then we can make a refactor so the user can decide if he wants to execute in proof mode or not

Oppen (approved):

PR #1532, opened 2024-01-02
Use `ProjectivePoint` from types-rs in ec_op builtin impl

fmoletta:
Use types-rs `ProjectivePoint` implementation (aka lambdaworks implementation) for ec_op builtin operations
Pending:

- [x] Merge [lambdaworks PR](https://github.com/lambdaclass/lambdaworks/pull/737)
- [x] Create [PR in types-rs](https://github.com/starknet-io/types-rs/pull/32)
- [x] Merge PR in types-rs
- [x] Update `starknet-core-types` version

### DIFF
--- a/vm/src/vm/runners/builtin_runner/ec_op.rs
+++ b/vm/src/vm/runners/builtin_runner/ec_op.rs
@@ -4,13 +4,13 @@ use crate::types::instance_definitions::ec_op_instance_def::{
     EcOpInstanceDef, CELLS_PER_EC_OP, INPUT_CELLS_PER_EC_OP,
 };
 use crate::types::relocatable::{MaybeRelocatable, Relocatable};
-use crate::utils::{felt_to_bigint, CAIRO_PRIME};
+use crate::utils::felt_to_bigint;
 use crate::vm::errors::memory_errors::MemoryError;
 use crate::vm::errors::runner_errors::RunnerError;
 use crate::vm::vm_memory::memory::Memory;
 use crate::vm::vm_memory::memory_segments::MemorySegmentManager;
 use crate::Felt252;
-use num_bigint::{BigInt, ToBigInt};
+use num_bigint::BigInt;
 use num_integer::{div_ceil, Integer};
 use num_traits::{One, Zero};
 use starknet_types_core::curve::ProjectivePoint;
@@ -62,8 +62,6 @@ impl EcOpBuiltinRunner {
         partial_sum: (Felt252, Felt252),
         doubled_point: (Felt252, Felt252),
         m: &Felt252,
-        alpha: &BigInt,
-        prime: &BigInt,
         height: u32,
     ) -> Result<(Felt252, Felt252), RunnerError> {
         let mut slope = felt_to_bigint(*m);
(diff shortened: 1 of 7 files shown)


vm/src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -62,31 +60,30 @@ impl EcOpBuiltinRunner {
...
         height: u32,
-    ) -> Result<(BigInt, BigInt), RunnerError> {
-        let mut slope = felt_to_bigint(*m);
-        let mut partial_sum_b = (felt_to_bigint(partial_sum.0), felt_to_bigint(partial_sum.1));
-        let mut doubled_point_b = (
-            felt_to_bigint(doubled_point.0),
-            felt_to_bigint(doubled_point.1),
-        );
+    ) -> Result<(Felt252, Felt252), RunnerError> {
+        let mut slope = m.to_bigint();
```
Oppen, line 65: Is the `slope` allowed to be negative? We're only considering the bits (not sign) in the body. So, I think that's either buggy or assumed positive, and in that case we could use the unsigned version.

vm/src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -62,31 +60,30 @@ impl EcOpBuiltinRunner {
...
         for _ in 0..height {
-            if (doubled_point_b.0.clone() - partial_sum_b.0.clone()).is_zero() {
-                #[allow(deprecated)]
+            if partial_sum_b.x() * doubled_point_b.z() == partial_sum_b.z() * doubled_point_b.x() {
                 return Err(RunnerError::EcOpSameXCoordinate(
                     Self::format_ec_op_error(partial_sum_b, slope, doubled_point_b)
                         .into_boxed_str(),
                 ));
             };
             if !(slope.clone() & &BigInt::one()).is_zero() {
```
Oppen, line 77: We could check the bit instead without creating any new temporary `BigInt`s (pseudo-code):
```rust
for i in 0..height.min(slope.bits()) {
    // check equality (does this need to happen several times?)
    if slope.bit(i) {
         partial_sum += doubled_point;
    }
    doubled_point = double_point.double(); // Is there an in-place version?
}
```

Oppen (approved):

vm/src/vm/runners/builtin_runner/ec_op.rs
```diff
@@ -62,23 +60,22 @@ impl EcOpBuiltinRunner {
...
         height: u32,
     ) -> Result<(Felt252, Felt252), RunnerError> {
-        let mut slope = m.to_bigint();
+        let slope = m.to_biguint();
         let mut partial_sum_b = ProjectivePoint::from_affine(partial_sum.0, partial_sum.1)
             .map_err(|_| RunnerError::PointNotOnCurve(Box::new(partial_sum)))?;
         let mut doubled_point_b = ProjectivePoint::from_affine(doubled_point.0, doubled_point.1)
             .map_err(|_| RunnerError::PointNotOnCurve(Box::new(doubled_point)))?;
-        for _ in 0..height {
+        for i in 0..(Into::<u64>::into(height)).min(slope.bits()) {
```
Oppen, line 68: This would be more brief and equally valid:
```suggestion
        for i in 0..(height as u64).min(slope.bits()) {
```

Oppen (approved):

Oppen (approved):

pefontana (approved):

PR #1539, opened 2024-01-04
Add `air_public_input` flag to cairo1-run

fmoletta:
Air Public Input can now be obtained from cairo 1 proof_mode runs


cairo1-run/src/main.rs
```diff
@@ -352,8 +359,67 @@ fn run(args: impl Iterator<Item = String>) -> Result<Vec<MaybeRelocatable>, Erro
...
+        // Set stop pointer for each builtin
+        for builtin in vm.builtin_runners.iter_mut() {
+            builtin.final_stack(
+                &vm.segments,
+                builtin_name_to_stack_pointer
+                    .get(builtin.name())
+                    .cloned()
+                    .unwrap_or_default(),
+            )?;
+        }
```
Oppen, line 410: Can't `finalize_segments` do this so we avoid exposing `VirtualMachine::builtin_runners` and `VirtualMachine::segments`?

fmoletta: In cairo 0, the builtin's stop pointers can be found contiguously at the end of the execution segment. Therefore we can take the current value of ap as stack_pointer and then iterate over the builtin runners calling `final_stack` (which will fetch the builtin's stop_ptr from memory[stack_ptr - 1]) on each of them, while advancing the stack_ptr by one each iteration to set all builtin's stop_ptr to the correct value. 
In cairo 1 this doesn't happen, the builtin's stop_ptr are mixed up with the other return values so we can't iterate contiguously over a memory range to fetch the stop pointers. The logic above this code calculates an individual stack_ptr for each builtin, therefore we need to call each builtin's `final_stack` method independently, without being able to rely on `read_return_values` or `get_builtins_final_stack` like we do with cairo 0 execution.

Oppen: Maybe a new method in the runner can do it?

fmoletta: It is kind of an specific behaviour to have as part of the public api, but I think we could add it to the vm as a `docs_hidden` method

cairo1-run/src/main.rs
```diff
@@ -290,20 +303,14 @@ fn run(args: impl Iterator<Item = String>) -> Result<Vec<MaybeRelocatable>, Erro
...
-    let mut vm = VirtualMachine::new(args.trace_file.is_some());
+    let mut vm = VirtualMachine::new(args.trace_file.is_some() || args.air_public_input.is_some());
     let end = runner.initialize(&mut vm)?;
 
     additional_initialization(&mut vm, data_len)?;
 
     // Run it until the end/ infinite loop in proof_mode
     runner.run_until_pc(end, &mut vm, &mut hint_processor)?;
-
-    if args.proof_mode {
```
juanbono, line 301: why did you remove this?

juanbono (dismissed):

cairo1-run/src/main.rs
fmoletta: The previous code called `runner.run_until_next_power_of_2(&mut vm, &mut hint_processor)` if proof_mode was activated. This same function is called in the `end_run` method if proof_mode is activated

juanbono (approved):

Oppen (approved):

PR #1551, opened 2024-01-09
Add `args` flag to `cairo1-run`

fmoletta:
Allows passing arguments to cairo 1 programs
As there isn't a set convention for arguments passed to cairo_run (and the current design doesn't allow using starknet's convention for arrays) the following convention was taken:
- Values should be passed separated by whitespaces (akin to starknet commands)
- Array values should be wrapped between brackets (New)
Example:
```
cargo run ../cairo_programs/cairo-1-programs/with_input/array_input_sum.cairo --layout all_cairo --args '2 [1 2 3 4] 0 [9 8]'
```


cairo1-run/src/main.rs
```diff
@@ -562,13 +615,35 @@ fn create_entry_code(
...
+        let FuncArg::Array(values) = arg else { continue };
+        array_args_data.push(ap_offset);
+        casm_extend! {ctx,
+            %{ memory[ap + 0] = segments.add() %}
+            ap += 1;
+        }
+        for (i, v) in values.iter().enumerate() {
+            let arr_at = (i + 1) as i16;
+            casm_extend! {ctx,
+                [ap + 0] = (v.to_bigint());
```
Oppen, line 637: Shouldn't this also increment `ap` after writing `v`?

fmoletta: This is copied from the original code [here](https://github.com/starkware-libs/cairo/blob/e0411d11efd530ba5b456a7c801416e55e680c4b/crates/cairo-lang-runner/src/lib.rs#L419)

Oppen (approved):

juanbono (approved):

Oppen (dismissed):

Oppen (approved):

juanbono (approved):

PR #1554, opened 2024-01-10
Remove cargo deps from make deps command

pefontana:
# Remove cargo deps from make deps command

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ cargo-deps:
 cairo1-run-deps:
 	cd cairo1-run; make deps
 
-deps: cargo-deps build-cairo-1-compiler build-cairo-2-compiler cairo1-run-deps
+deps: build-cairo-1-compiler build-cairo-2-compiler cairo1-run-deps
 	pyenv install -s pypy3.9-7.3.9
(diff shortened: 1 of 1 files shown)


Oppen:
Did you try updating the dependency? Current is 0.6.4 according to crates.io.

fmoletta (commented):
I suggest commenting the targets rather than removing them and adding a comment indicating that they were temporarily removed due to version issues so we don't forget to restore them once we update the rust version in the future

fmoletta:
> Did you try updating the dependency? Current is 0.6.4 according to crates.io.

The problem is that one 

> Did you try updating the dependency? Current is 0.6.4 according to crates.io.

Updating the dependency doesn't solve it (as the problem is that later versions use rust library features that got stabilized in higher rust versions.
Downgrading the version to `0.5.1` does work

Oppen (dismissed):

pefontana:
> I suggest commenting the targets rather than removing them and adding a comment indicating that they were temporarily removed due to version issues so we don't forget to restore them once we update the rust version in the future

Sure!
Done

fmoletta (approved):

Oppen (approved):

PR #1555, opened 2024-01-10
Update proof_programs symlinks

fmoletta:
In order to compile and run cairo programs in proof_mode, a proof_mode directory was created inside the cairo_programs directory, with symbolic links to cairo files in the parent directory. Various test files were added later to the cairo_programs directory without adding a symlink for them in the proof_programs directory. This PR adds those missing symlinks.


Oppen (approved):
LGTM. A further improvement we could make is maybe just creating them with a `make` target so it happens automatically via the dependency resolution.

juanbono (approved):

PR #1562, opened 2024-01-12
Remove unecessary conversion functions between `Felt` & `BigUint`/`BigInt`

### LINKED ISSUE #1495: Refactor biguint_to_felt and bigint_to_felt functions
We should delete the functions `felt_to_biguint`, `felt_to_bigint`, `biguint_to_felt` and `bigint_to_felt`
https://github.com/lambdaclass/cairo-vm/blob/55eaaa7be3778b84871e24d1f87e4ae6f62184dd/vm/src/utils.rs#L67
And use the BigUint and BigInts conversion to felt, provided by the `starknet-types-core` crate

fmoletta:
Closes #1495


vm/src/hint_processor/builtin_hint_processor/ec_utils.rs
```diff
@@ -144,10 +144,10 @@ pub fn recover_y_hint(
     let p_x = get_integer_from_var_name("x", vm, ids_data, ap_tracking)?.into_owned();
     let p_addr = get_relocatable_from_var_name("p", vm, ids_data, ap_tracking)?;
     vm.insert_value(p_addr, p_x)?;
-    let p_y = biguint_to_felt(
-        &recover_y(&felt_to_biguint(p_x))
+    let p_y = Felt252::from(
+        &recover_y(&p_x.to_biguint())
```
Oppen, line 148: I see some redundant conversion here. Maybe `recover_y` could simply return the `Felt252` instead.

Oppen (commented):
Nice cleanup. There's some slowdown that may be worth investigating.

vm/src/hint_processor/builtin_hint_processor/ec_utils.rs
fmoletta: `recover_y` is also used by other functions that use BigUint and don't need to convert  from Felt to use it, so it is better to leave the conversion outside and let the function just handle BigUint

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 #### Upcoming Changes
 
+* feat: Remove unecessary conversion functions between `Felt` & `BigUint`/`BigInt` [#1562](https://github.com/lambdaclass/cairo-vm/pull/1562)
+  * Remove the following functions:
+    * felt_from_biguint
+    * felt_from_bigint
+    * felt_to_biguint
+    * felt_to_bigint
```
Oppen, line 10: These functions were public, so we need to report the change as a breaking one.

Oppen (dismissed):

pefontana (approved):

pefontana (dismissed):

juanbono (approved):

pefontana (approved):

PR #1578, opened 2024-01-17
Add `cairo_pie_output` flag to `cairo_vm_cli`

fmoletta:
This also involves:
- Fixing serialization of CairoPie to be fully compatible with the python version
- Implementing CairoPie::write_zip_file
- Adding comparison of cairo pie outputs between python and rust vm
Bonus:
- Moved handling of required and exclusive arguments to struct definition using clap derives


Oppen:
Suggestion: maybe just write to a folder instead and use a tiny script for zipping?

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -199,6 +231,41 @@ mod serde_impl {
...
+            let mem_cap = values.len() * ADDR_BYTE_LEN + values.len() * FIELD_BYTE_LEN;
+            let mut res = Vec::with_capacity(mem_cap);
+
+            for ((segment, offset), value) in values.iter() {
+                let mem_addr = ADDR_BASE + *segment as u64 * OFFSET_BASE + *offset as u64;
+                res.extend_from_slice(mem_addr.to_le_bytes().as_ref());
+                match value {
+                    // Serializes RelocatableValue(little endian):
+                    // 1bit |   SEGMENT_BITS |   OFFSET_BITS
+                    // 1    |     segment    |   offset
```
Oppen, line 248: This is 64 bits, so `u64` seems like a better choice than `BigUint`.

fmoletta: Its 8 bits for the address (which is built as an u64 in line 243) but in this case we are handling a memory value that is a relocatable, which takes 32 bytes

Oppen (approved):

vm/src/vm/runners/cairo_pie.rs
Oppen: `Relocatable` is 8 bytes AFAIR.

Oppen (approved):

pefontana (approved):

PR #1580, opened 2024-01-18
Output builtin features for bootloader support

odesenfans:
This commit introduces the following features/changes:

* Paging: pages can now be added to the output builtin. These pages are reflected in the public memory of the VM when exporting the public input.
* The state of the output builtin can now be modified using the new `set_state` method.
* The output builtin can now handle attributes. These are used to generate the fact topologies of the bootloader.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/vm/src/vm/runners/builtin_runner/output.rs
+++ b/vm/src/vm/runners/builtin_runner/output.rs
@@ -31,16 +31,6 @@ impl OutputBuiltinRunner {
         }
     }
 
-    pub fn from_segment(segment: &Relocatable, included: bool) -> Self {
-        Self {
-            base: segment.segment_index as usize,
-            pages: HashMap::default(),
-            attributes: HashMap::default(),
-            stop_ptr: None,
-            included,
-        }
-    }
-
     pub fn new_state(&mut self, base: usize, included: bool) {
         self.base = base;
         self.pages = HashMap::default();
@@ -530,19 +520,6 @@ mod tests {
 
         let memory = memory![((0, 0), 0), ((0, 1), 1), ((0, 2), 2), ((0, 3), 3)];
         assert!(builtin.air_private_input(&memory).is_empty());
-    }
-
-    #[test]
-    fn from_segment() {
-        let segment_base = Relocatable::from((10, 0));
-        let included = true;
-
-        let builtin = OutputBuiltinRunner::from_segment(&segment_base, included);
-        assert_eq!(builtin.base, segment_base.segment_index as usize);
-        assert_eq!(builtin.stop_ptr, None);
-        assert_eq!(builtin.pages, HashMap::default());
-        assert_eq!(builtin.attributes, HashMap::default());
-        assert_eq!(builtin.included, included);
     }
 
     #[test]
--- a/vm/src/vm/runners/builtin_runner/output.rs
+++ b/vm/src/vm/runners/builtin_runner/output.rs
@@ -556,6 +556,27 @@ mod tests {
     }
 
     #[test]
+    fn new_state() {
+        let mut builtin = OutputBuiltinRunner {
+            base: 10,
+            pages: HashMap::from([(1, PublicMemoryPage { start: 0, size: 3 })]),
+            attributes: HashMap::from([("gps_fact_topology".to_string(), vec![0, 2, 0])]),
+            stop_ptr: Some(10),
+            included: true,
+        };
+
+        let new_base = 11;
+        let new_included = false;
+        builtin.new_state(new_base, new_included);
+
+        assert_eq!(builtin.base, new_base);
+        assert!(builtin.pages.is_empty());
+        assert!(builtin.attributes.is_empty());
+        assert_eq!(builtin.stop_ptr, None);
+        assert_eq!(builtin.included, new_included);
+    }
+
+    #[test]
     fn add_page() {
         let mut builtin = OutputBuiltinRunner::new(true);
         assert_eq!(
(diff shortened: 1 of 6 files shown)


vm/src/vm/runners/builtin_runner/output.rs
```diff
@@ -20,11 +24,31 @@ impl OutputBuiltinRunner {
...
 
+    pub fn from_segment(segment: &Relocatable, included: bool) -> Self {
+        Self {
+            base: segment.segment_index as usize,
+            pages: HashMap::default(),
+            attributes: HashMap::default(),
+            stop_ptr: None,
+            included,
+        }
+    }
```
fmoletta, line 42: Why is this method needed?

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -41,6 +41,8 @@ pub type Pages = HashMap<usize, PublicMemoryPage>;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct OutputBuiltinAdditionalData {
+    #[serde(skip)]
+    pub base: usize,
```
fmoletta, line 45: Why is this field needed?

odesenfans: The problem is the following: the bootloader starts by saving the VM output builtin and replaces it by a dedicated one to run the "simple" bootloader program and capture its output.

To resolve that problem, I thought about adding this field here so that we can get the full state of the output builtin in one go. We don't want to serialize it, hence the `skip`. But I'm open to other proposals, I could create a dedicated struct for example.

vm/src/vm/runners/builtin_runner/output.rs
```diff
@@ -11,7 +13,9 @@ use super::OUTPUT_BUILTIN_NAME;
 
 #[derive(Debug, Clone)]
 pub struct OutputBuiltinRunner {
-    base: usize,
+    pub(crate) base: usize,
```
odesenfans, line 16: To check: I'm pretty sure I can revert this line.

vm/src/vm/runners/builtin_runner/output.rs
odesenfans: As mentioned in my other reply, the first thing the bootloader does is create a new output builtin to capture the output of the simple bootloader program. To do this, it allocates a new segment and uses it as the base. It's basically a `new` method with the `base` provided as argument.

fmoletta: Shouldn't that be done by calling:
```
let mut output = OutputBuiltinRunner::new(true);
output.initialize_segments();
```
I don't see the need for this specific method

vm/src/vm/runners/cairo_pie.rs
fmoletta: Why does the new output builtin need to know the base of the old one?

odesenfans: We just store it so that we can restore it later on.

vm/src/vm/runners/builtin_runner/output.rs
odesenfans: The way the bootloader works, this method makes more sense (the hints always create a segment, then the builtin from it), so this felt natural. But it can be done the way you suggest.

odesenfans: Dropped the method.

vm/src/vm/runners/cairo_pie.rs
fmoletta: The output builtin's segment index can already be found on the `CairoPie` itself on the  CairoPie.metdata.builtin_segments map. I don't see the need to add it in the `BuiltinAdditionalData` too

odesenfans: When implementing the bootloader hint there is no `CairoPie` object to store this info. The flow goes like this:

1. VM starts, output builtin is configured
2. Bootloader starts, one of the first steps is to run the "simple" bootloader program and capture its output
3. To achieve this, the first bootloader hint stores the original output builtin configured at step 1 and replaces it with a new one to capture the output of the simple bootloader.
4. After having executed the simple bootloader, the bootloader restores the original output builtin to write its own output.

The simplest way to store the output builtin state was to store this information in the additional data (as is done in the Python VM). I can as easily create a new struct if you prefer it that way.

fmoletta: I think there is a mixup in what the output builtin's state is (base, attributes & pages) and what the additional data is (attributes & pages). The python vm has its own `get_state` method that has the behaviour you are adding to `get_additional_data`

odesenfans: I can definitely do that, but then the methods to retrieve/restore the output builtin state become more specific. I can either:

1. Replace the whole builtin object
2. Add a method that takes a base + additional data to set the builtin state
3. Add a new struct (ex: `OutputBuiltinState`)  and add a getter + setter for this particular struct.

What do you prefer?

fmoletta: 2 & 3 are both good solutions, 2 sound simpler so I would go for that one

odesenfans: I went with solution 3 and introduced an `OutputBuiltinState` struct.

fmoletta (approved):

juanbono (approved):

pefontana:
@odesenfans can you solve merge conflicts so we can merge it

fmoletta (approved):

odesenfans:
FYI the CI/CD fails when publishing the benchmarks to Github, no idea why.

juanbono (approved):

juanbono (changes requested):
@odesenfans looks like some tests are failing. Can you take a look?

odesenfans:
@juanbono I think the failures are related to https://github.com/lambdaclass/cairo-vm/pull/1639, I rebased on top of the latest master.

pefontana (approved):

juanbono (approved):

PR #1585, opened 2024-01-18
bugfixes: Fix memory hole count inconsistencies

fmoletta:
BugFixes:
* In the method `MemorySegmentManager::get_memory_holes`, builtin segments are skipped istead of copying the behaviour of the python vm, which marks all builtin addresses as accessed before counting memory holes. This works fine for all builtins, but not for the output_builtin, as the implementation of get_memory_accesses for it is not the same as the rest of the builtins. Now the output builtin's memory is searched for memory holes like the non-builtin segments
* In the method Memory::RelocateMemory (the one that relocates temporary memory) the new relocated memory cells were left as not accessed regardless of their prior non-relocated state. Now new relocated cells are marked as accessed if their pre-relocation version was marked as accessed
Other:
* Cairo PIE comparison script no longer skips files affected by the above bugs
* Updated memory_hole values for integration tests


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 #### Upcoming Changes
 
+* bugfixes: Fix memory hole count inconsistencies #[1585] (https://github.com/lambdaclass/cairo-vm/pull/1585)
+  * Output builtin memory segment is no longer skipped when counting memory holes
+  * Temporary memory cells now keep their accessed status when relocated
+  * Signature change: `get_memory_holes(&self, builtin_count: usize) -> Result<usize, MemoryError>` ->  `get_memory_holes(&self, builtin_count: usize,  has_output_builtin: bool) -> Result<usize, MemoryError>`
```
Oppen, line 8: This is a breaking change, let's mention it. Otherwise, LGTM.

juanbono (approved):

Oppen (dismissed):

pefontana (approved):

edg-l (approved):

juanbono (approved):

PR #1597, opened 2024-02-01
[Cairo1] Append return values to the output segment when running in proof_mode

fmoletta:
This involves:

- Forcing the output builtin to be included when running cairo 1 programs in proof_mode (And manually setting it's stop_ptr and segment_size)
- Adding extra instructions in the proof_mode_header (after calling the first program instruction and before initiating the infinite loop) to copy the return values into the output builtin segment

Testing:

- This new behaviour was tested manually by printing out the memory when running the available test programs to ensure that it behaves correctly with one or more return values and all flags enabled.

- There is currently no way to create a test for this behaviour without some code refactors that escape the scope of the PR.

Also in this PR:

- Restored a test that was accidentally removed when solving merge conflicts in a previous PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 #### Upcoming Changes
 
+* feat: Append return values to the output segment when running cairo1-run in proof_mode [#1597](https://github.com/lambdaclass/cairo-vm/pull/1597)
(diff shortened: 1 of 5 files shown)


pefontana (approved):

juanbono (approved):

juanbono:
We should work on adding the required changes to test this

PR #1599, opened 2024-02-05
Feature: make test_utils macros public

### LINKED ISSUE #1595: Make it easier to test hints implemented out of the cairo-vm repository
Hello,

We're currently developing some hints for a project. These hints live in a different Git repository. We developed hints for the bootloader in a fork of `cairo-vm` so I hadn't noticed yet, but it is surprisingly way more difficult to write unit tests out of the repository. Example:

If we take a hint

```
%{
    a = ids.a
%}
```

We can easily write a unit test for this in the `cairo-vm` repository like this:

```
let mut vm = vm!();
vm.segments = segments![((1, 0), 42)];
vm.run_context.fp = 1;
let mut exec_scopes = ExecutionScopes::new();
let ids_data = ids_data!["a"];

hint(&mut vm, &mut exec_scopes, &ids_data).unwrap();

let a: u64 = exec_scopes.get("a").unwrap();
assert_eq!(a, 42);
```

But out of tree, things become more complex / impossible.

```
// let mut vm = vm!();   // vm! is private
let mut vm = VirtualMachine::new(false);   // but easily replaced by a one-liner, so this is fine

// vm.segments = segments![((1, 0), 42)];   // segments! is also private
vm.add_memory_segment();
vm.add_memory_segment();
vm.insert_value(Relocatable::from((1, 0), 42);  // A bit more verbose, but OK

// vm.run_context.fp = 1;
// Incrementing FP is impossible because `run_context` is `pub(crate)`.

// let ids_data = ids_data!["a"];   // ids_data is private as well
let ids_data = HashMap::from([("a".to_string(), HintReference::new_simple(-1))];
```

So we're in a situation where AP/FP manipulations are impossible because of VM members being `pub(crate)` and the utils macros are not public so they need to be reimplemented manually.

The first point is the one that actually is blocking, so I would like your opinion on the matter. What's a good solution there? Making VM fields `pub` instead of `pub(crate)`? Adding accessors? Another solution I have not thought of?

**Describe the solution you'd like**
Ideally, it should be possible to write unit tests the same way independently of where the implementation lives.

**Additional context**
This is related to the implementation of hints for the bootloader and Starknet OS.

odesenfans:
This PR makes all macros in the `test_utils` module public to allow for easier testing of hints (and other features) living in other crates/repositories.

Fixes #1595.

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/utils.rs
+++ b/vm/src/utils.rs
@@ -232,7 +232,7 @@ pub mod test_utils {
             $crate::types::relocatable::MaybeRelocatable::from(($val1, $val2))
         };
         ($val1 : expr) => {
-            $crate::types::relocatable::MaybeRelocatable::from(crate::Felt252::from($val1 as i128))
+            $crate::types::relocatable::MaybeRelocatable::from($crate::Felt252::from($val1 as i128))
         };
     }
     pub use mayberelocatable;
@@ -240,7 +240,8 @@ pub mod test_utils {
     #[macro_export]
     macro_rules! references {
         ($num: expr) => {{
-            let mut references = crate::stdlib::collections::HashMap::<usize, HintReference>::new();
+            let mut references =
+                $crate::stdlib::collections::HashMap::<usize, HintReference>::new();
             for i in 0..$num {
                 references.insert(i as usize, HintReference::new_simple((i as i32 - $num)));
             }
@@ -295,27 +296,27 @@ pub mod test_utils {
         //Program with builtins
         ( $( $builtin_name: expr ),* ) => {{
             let shared_program_data = SharedProgramData {
-                data: crate::stdlib::vec::Vec::new(),
+                data: $crate::stdlib::vec::Vec::new(),
                 hints_collection: HintsCollection::new(&BTreeMap::new(), 0).unwrap(),
                 main: None,
                 start: None,
                 end: None,
-                error_message_attributes: crate::stdlib::vec::Vec::new(),
+                error_message_attributes: $crate::stdlib::vec::Vec::new(),
                 instruction_locations: None,
-                identifiers: crate::stdlib::collections::HashMap::new(),
+                identifiers: $crate::stdlib::collections::HashMap::new(),
                 reference_manager: Program::get_reference_list(&ReferenceManager {
-                    references: crate::stdlib::vec::Vec::new(),
+                    references: $crate::stdlib::vec::Vec::new(),
                 }),
             };
             Program {
                 shared_program_data: Arc::new(shared_program_data),
-                constants: crate::stdlib::collections::HashMap::new(),
+                constants: $crate::stdlib::collections::HashMap::new(),
                 builtins: vec![$( $builtin_name ),*],
             }
         }};
         ($($field:ident = $value:expr),* $(,)?) => {{
 
-            let program_flat = crate::utils::test_utils::ProgramFlat {
+            let program_flat = $crate::utils::test_utils::ProgramFlat {
                 $(
                     $field: $value,
                 )*
@@ -427,9 +428,9 @@ pub mod test_utils {
             {
                 let ids_names = vec![$( $name ),*];
                 let references = references!(ids_names.len() as i32);
-                let mut ids_data = crate::stdlib::collections::HashMap::<crate::stdlib::string::String, HintReference>::new();
+                let mut ids_data = $crate::stdlib::collections::HashMap::<$crate::stdlib::string::String, HintReference>::new();
                 for (i, name) in ids_names.iter().enumerate() {
-                    ids_data.insert(crate::stdlib::string::ToString::to_string(name), references.get(&i).unwrap().clone());
+                    ids_data.insert($crate::stdlib::string::ToString::to_string(name), references.get(&i).unwrap().clone());
                 }
                 ids_data
             }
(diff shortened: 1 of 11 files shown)


vm/src/utils.rs
```diff
@@ -76,7 +76,7 @@ pub mod test_utils {
             Into::<num_bigint::BigInt>::into($val)
         };
     }
-    pub(crate) use bigint;
+    pub use bigint;
```
fmoletta, line 79: This is not really vm-related, so I don't think it should be included in the vm's public api. Moreover, we may choose to remove the usage of BigInt from the VM altogether and remove this macro too.

vm/src/utils.rs
```diff
@@ -87,15 +87,15 @@ pub mod test_utils {
...
+    pub use bigint_str;
 
     #[macro_export]
     macro_rules! biguint {
         ($val : expr) => {
             Into::<num_bigint::BigUint>::into($val as u128)
         };
     }
-    pub(crate) use biguint;
+    pub use biguint;
```
fmoletta, line 98: This is not really vm-related, so I don't think it should be included in the vm's public api. Moreover, we may choose to remove the usage of BigUint from the VM altogether and remove this macro too.

vm/src/utils.rs
```diff
@@ -106,7 +106,7 @@ pub mod test_utils {
             num_bigint::BigUint::parse_bytes($val.as_bytes(), $opt).expect("Couldn't parse bytes")
         };
     }
-    pub(crate) use biguint_str;
+    pub use biguint_str;
```
fmoletta, line 109: This is not really vm-related, so I don't think it should be included in the vm's public api. Moreover, we may choose to remove the usage of BigUint from the VM altogether and remove this macro too.

vm/src/utils.rs
```diff
@@ -155,8 +158,9 @@ pub mod test_utils {
             }
         };
     }
-    pub(crate) use memory_from_memory;
+    pub use memory_from_memory;
```
fmoletta, line 161: This is an auxiliary macro used to implement the `memory` macro, it shouldn't be made public

vm/src/utils.rs
```diff
@@ -189,17 +193,19 @@ pub mod test_utils {
             }
         };
     }
-    pub(crate) use memory_inner;
+    pub use memory_inner;
```
fmoletta, line 196: This is an auxiliary macro used to implement the memory macro, it shouldn't be made public

vm/src/utils.rs
```diff
@@ -263,12 +273,14 @@ pub mod test_utils {
...
         };
     }
-    pub(crate) use cairo_runner;
+    pub use cairo_runner;
 
-    pub(crate) use crate::stdlib::{collections::BTreeMap, sync::Arc};
+    pub use crate::stdlib::{collections::BTreeMap, sync::Arc};
     pub(crate) use crate::types::program::HintsCollection;
-    pub(crate) use crate::types::program::Program;
+    pub use crate::types::program::Program;
```
fmoletta, line 280: `Program` is already public, we should use the full path in the macro rather than re-exporting it here

vm/src/utils.rs
odesenfans: Makes sense. I made it public out of simplicity because keeping it `pub(crate)` in a now public module induces a few unused import warnings. I'll find a solution.

vm/src/utils.rs
odesenfans: These macros need to be public to make the `memory` and `segments` macros usable. Users need to import them as well.

odesenfans:
This PR should be mergeable. Most of the macros need to be public as they are used in public macros. I addressed all the comments that could be.

vm/src/vm/vm_memory/memory.rs
```diff
@@ -92,8 +92,8 @@ impl AddressSet {
 }
 
 pub struct Memory {
-    pub(crate) data: Vec<Vec<Option<MemoryCell>>>,
```
juanbono, line 95: Why did you made these 2 fields public? they are private to make it easier to refactor the representation and improve it. There are many methods for retrieving and inspecting the Memory and pretty-printing it (Display instance).

vm/src/vm/vm_memory/memory_segments.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
 pub struct MemorySegmentManager {
     pub segment_sizes: HashMap<usize, usize>,
     pub segment_used_sizes: Option<Vec<usize>>,
-    pub(crate) memory: Memory,
+    pub memory: Memory,
```
juanbono, line 18: this should not be made public.

juanbono (changes requested):
See comments

vm/src/vm/vm_memory/memory_segments.rs
odesenfans: Some of the macros in `test_utils` require access to this field. As the goal is to make the macros usable out of tree, the easy solution is to make this field public. I understand the concern though. Any suggestion on how to work around the issue?

vm/src/vm/vm_memory/memory.rs
odesenfans: Same, some macros access these fields directly so they must be public to use the macros out of tree. I'll try to think of another solution.

vm/src/vm/vm_memory/memory_segments.rs
Oppen: > Any suggestion on how to work around the issue?

It depends on what you want to do in the tests. Maybe you could just use the `insert` family of functions.

fmoletta:
Hello! Making these macros public in the current state involves making a lot of fields internal to the VM public which is not desirable at all. I think we should address this task in a different way. What these macros currently do can be implemented using the vm's public api, so we should re-implement or implement new macros that use the public api instead.

odesenfans:
Ok @fmoletta then let's close this PR, no worries.

Issue #1593, opened 2024-01-30
[Cairo1] Append returned values to the Output builtin segment


pefontana:
When executing cairo1-run crate with proof mode, we need to add some instructions in the `create_entry_code` function, that appends the returned values of main to the Output Builtin segment.

For a first iteration, we can assume the return value is just a Felt

PR #1602, opened 2024-02-06
[Cairo1] Add a test to check that the output segment contains the return values when running in proof-mode

fmoletta:
Tests the behaviour added in #1593 
Depends on refactor done in #1601


juanbono (approved):
Nice

pefontana (approved):

PR #1616, opened 2024-02-09
Proofreading the docs

Ursulafe:
# Proofread of python_vm's README

## Description

Came across few errors while reading the docs (for python_vm), so I corrected them.
Doesn't require testing or changelog entry I believe.

(For line 197 I checked compiler/encode.py was the correct path)


Oppen:
Sorry, we're note accepting spellchecks from external contributors at this time.

PR #1620, opened 2024-02-09
Add `Labels` section to CONTRIBUTING.md + update changelog skip labels

fmoletta:
Adds a `Labels` section with short descriptions of commonly-used PR labels to CONTRIBUTING.md.
Removes `coverage` and adds `tests` & `documentation` labels to changelog workflow's skipLabels list.
Resaoning:
*  `tests` & `coverage` labels are often used for the same purposes  so it doesn't make sense to have both. (And having PRs with only improving coverage rather than testing as an objective has lead to unuseful tests being added in the past).
* There is no reason for documentation updates to be reflected in the changelog

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -13,4 +13,4 @@ jobs:
     steps:
     - uses: dangoslen/changelog-enforcer@v3
       with:
-        skipLabels: pipelines,coverage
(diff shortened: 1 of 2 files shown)


juanbono (approved):

Oppen:
I think only code owners should be applying the labels.

pefontana (approved):

PR #1632, opened 2024-02-24
Fix: set up builtins initial stack based on program builtins

### LINKED ISSUE #1631: Builtin pointers not allocated when running a program with `allow-missing-builtins`
**Describe the bug**
For an experiment, I tried to execute the Starknet bootloader with layout `starknet` and allowing missing builtins to suppress the warning about the absence of the `keccak` builtin in the layout. This bypasses the builtin check successfully but yields the following error:

```
$ ./target/debug/stone-prover-cli prove --with-bootloader dependencies/cairo-programs/cairo0/fibonacci/fibonacci.json --verifier=l1 --layout=starknet --allow-missing-builtins --output-file /tmp/proof.json
info  execution in progress...
error failed to run Cairo program: starkware/cairo/bootloaders/bootloader/bootloader.cairo:60:5: While trying to retrieve the implicit argument 'poseidon_ptr' in:
/home/olivier/git/moonsong-labs/starkware/cairo-lang/src/starkware/cairo/bootloaders/simple_bootloader/run_simple_bootloader.cairo:21:5: While expanding the reference 'poseidon_ptr' in:
    poseidon_ptr,
    ^**********^
starkware/cairo/bootloaders/bootloader/bootloader.cairo:37:5: Error at pc=0:553:
Couldn't compute operand op1. Unknown value for memory cell 1:9
Cairo traceback (most recent call last):
<start>:3:1: (pc=0:2)
```

This is somewhat surprising because the missing builtin is `keccak`, not `poseidon`. Upon further inspection, I realized that the Cairo program expects pointers to the builtins in the main function:

```
func main{
    output_ptr: felt*,
    pedersen_ptr: HashBuiltin*,
    range_check_ptr,
    ecdsa_ptr,
    bitwise_ptr,
    ec_op_ptr,
    keccak_ptr,
    poseidon_ptr,
}() {
```

My best guess so far is that the issue happens because the builtin pointers are generated from the list of builtins in the layout, and not the list of builtins required by the program.

**Expected behavior**
The pointers for each builtin should be initialized based on what the program requires, not on the builtins available in the layout.

59254f4c53594217e0182e04008e56df5c187889

odesenfans:
Problem: when using `allow-missing-builtins` for a program, some builtins requested by the program may not be part of the layout. This results in an incorrect configuration of the initial stack of the program as the program will expect pointers for all builtins, even though it may not access them later down the line. An example of this is any program that invokes another program (ex: the bootloader).

Solution: copy the Python VM implementation. We now iterate over the program builtins and set the initial stack of the builtin if it is set up in the VM and a default value of (0, 0) otherwise.

Fixes #1631.

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/vm/runners/cairo_runner.rs
+++ b/vm/src/vm/runners/cairo_runner.rs
@@ -296,7 +296,7 @@ impl CairoRunner {
                         instance_def.ratio,
                         included,
                     )
-                        .into(),
+                    .into(),
                 );
             }
         }
@@ -613,10 +613,10 @@ impl CairoRunner {
             self.initial_ap = self.initial_fp;
             return Ok(self.program_base.as_ref().ok_or(RunnerError::NoProgBase)?
                 + self
-                .program
-                .shared_program_data
-                .end
-                .ok_or(RunnerError::NoProgramEnd)?);
+                    .program
+                    .shared_program_data
+                    .end
+                    .ok_or(RunnerError::NoProgramEnd)?);
         }
 
         let return_fp = vm.segments.add();
@@ -690,11 +690,11 @@ impl CairoRunner {
     ) -> Result<(), VirtualMachineError> {
         let references = &self.program.shared_program_data.reference_manager;
         #[cfg(not(feature = "extensive_hints"))]
-            let hint_data = self.get_hint_data(references, hint_processor)?;
+        let hint_data = self.get_hint_data(references, hint_processor)?;
         #[cfg(feature = "extensive_hints")]
-            let mut hint_data = self.get_hint_data(references, hint_processor)?;
+        let mut hint_data = self.get_hint_data(references, hint_processor)?;
         #[cfg(feature = "extensive_hints")]
-            let mut hint_ranges = self
+        let mut hint_ranges = self
             .program
             .shared_program_data
             .hints_collection
@@ -707,9 +707,9 @@ impl CairoRunner {
                 hint_processor,
                 &mut self.exec_scopes,
                 #[cfg(feature = "extensive_hints")]
-                    &mut hint_data,
+                &mut hint_data,
                 #[cfg(not(feature = "extensive_hints"))]
-                    self.program
+                self.program
                     .shared_program_data
                     .hints_collection
                     .get_hint_range_for_pc(vm.run_context.pc.offset)
@@ -718,7 +718,7 @@ impl CairoRunner {
                     })
                     .unwrap_or(&[]),
                 #[cfg(feature = "extensive_hints")]
-                    &mut hint_ranges,
+                &mut hint_ranges,
                 &self.program.constants,
             )?;
 
@@ -741,18 +741,18 @@ impl CairoRunner {
     ) -> Result<(), VirtualMachineError> {
         let references = &self.program.shared_program_data.reference_manager;
         #[cfg(not(feature = "extensive_hints"))]
-            let hint_data = self.get_hint_data(references, hint_processor)?;
+        let hint_data = self.get_hint_data(references, hint_processor)?;
         #[cfg(feature = "extensive_hints")]
-            let mut hint_data = self.get_hint_data(references, hint_processor)?;
+        let mut hint_data = self.get_hint_data(references, hint_processor)?;
         #[cfg(feature = "extensive_hints")]
-            let mut hint_ranges = self
+        let mut hint_ranges = self
             .program
             .shared_program_data
             .hints_collection
             .hints_ranges
             .clone();
         #[cfg(not(feature = "extensive_hints"))]
-            let hint_data = &self
+        let hint_data = &self
             .program
             .shared_program_data
             .hints_collection
@@ -771,11 +771,11 @@ impl CairoRunner {
                 hint_processor,
                 &mut self.exec_scopes,
                 #[cfg(feature = "extensive_hints")]
-                    &mut hint_data,
+                &mut hint_data,
                 #[cfg(not(feature = "extensive_hints"))]
-                    hint_data,
+                hint_data,
                 #[cfg(feature = "extensive_hints")]
-                    &mut hint_ranges,
+                &mut hint_ranges,
                 &self.program.constants,
             )?;
         }
@@ -835,7 +835,7 @@ impl CairoRunner {
                     (rc_max - rc_min) as usize,
                 ))),
             )
-                .into());
+            .into());
         }
 
         Ok(())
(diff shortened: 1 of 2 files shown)


odesenfans:
A few points of attention for the review:
1. `get_builtins_final_stack` appears to be used in tests only, I wonder if that function is still relevant. I duplicated the change made in `read_return_values` as it seemed pertinent, but I may be missing some context.
2. The `get_output_unordered_builtins` test does not seem to make sense so I disabled it. If I understand this correctly the test checks if a program can specify its builtins out of order, but that seems like a silly idea. Might have misunderstood though, I didn't want to disassemble the program to check.

juanbono (changes requested):
Can you take a look to the failing tests? they are failing with the wasm feature on

odesenfans:
I pushed a fix. The borrow checker was being a bit zealous for some reason.

fmoletta (commented):
Looks good! Could you add a test for this specific case so we don't run into the same issue in the future?

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -3217,6 +3243,9 @@ mod tests {
     }
 
     /// Test that `get_output()` works when the `output` builtin is not the first one.
+    // TODO review: this test swaps builtins, but builtins should appear in the order
+    //              specified in the program (i.e. ordered).
+    #[ignore = "this test does not seem to make sense"]
```
pefontana, line 3248: Is this test now failing???
If it fails can we correct it?

pefontana (changes requested):
Great @odesenfans, I think if we change that two little things we can merge it!

vm/src/vm/runners/cairo_runner.rs
odesenfans: Yes, this test is failing. I'd need to disassemble it or find the source to be sure, but it feels like this test was built ad hoc for something that should not happen (the builtins stack being out of order). If you look at the previous stack initialization code, it looks like this:

```
        for builtin_runner in vm.builtin_runners.iter() {
            stack.append(&mut builtin_runner.initial_stack());
        }
```

In the test, they swap the builtins so the output is `[<bitwise_base>, <output_base>]`. The only way this can work is if the program is aware of that and reverts the order of builtins. IMO this should never happen, hence why I disabled the test and suggested its removal. But I might be missing something. I think @azteca1998 wrote this test, maybe he could explain what it does in more details?

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -5506,9 +5535,72 @@ mod tests {
...
+            .unwrap();
+
+        let initial_stack = cairo_runner.prepare_builtins_initial_stack(&vm);
+        assert_eq!(initial_stack, vec![]);
+    }
+
+    #[rstest]
+    #[case("small", true)]
+    #[case("all_cairo", false)]
+    fn prepare_builtins_initial_stack_missing_builtins(
```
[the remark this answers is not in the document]
odesenfans: I decided to split the test in two and use `#[test]`.

vm/src/vm/runners/cairo_runner.rs
fmoletta: A different builtin order is used for starknet contracts, maybe we could use [initialize_all_builtins](https://github.com/lambdaclass/cairo-vm/blob/6ae7b32b0cb8cf758e52fbf795d6689b81a7ca0e/vm/src/vm/runners/cairo_runner.rs#L350) in the test  instead of `initialize_builtins` (and remove output from the program builtins) to test a proper use case where ouput is not the first builtin

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -5506,9 +5534,76 @@ mod tests {
...
+            .initialize(&mut vm, allow_missing_builtins)
+            .unwrap();
+
+        let initial_stack = cairo_runner.prepare_builtins_initial_stack(&vm);
+        assert_eq!(initial_stack, vec![]);
+    }
+
+    fn prepare_stack_for_program_with_builtin(layout: &str, allow_missing_builtins: bool) {
+        let proof_mode = false;
+        let trace_enabled = true;
```
fmoletta, line 5566: This test is missing the `#[test]` attribute

vm/src/vm/runners/cairo_runner.rs
fmoletta: This test was added because the write_output function used to asume that the output builtin was always the first one, which is not always the case in a starknet context. 
For the starknet context we use [initialize_function_runner](https://github.com/lambdaclass/cairo-vm/blob/6ae7b32b0cb8cf758e52fbf795d6689b81a7ca0e/vm/src/vm/runners/cairo_runner.rs#L1217), that initializes first the builtins in the order that they appear in the program, but then also initializes the remaining builtins following starknet's builtins order, which is different from the vm's expected builtin order

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -488,18 +489,38 @@ impl CairoRunner {
...
+            .collect();
+
+        // For each builtin in the program, set up the corresponding initial stack.
+        // Builtins required in the program but not set up in the VM (i.e. when specifying
+        // `allow_missing_builtins`) are set to (0, 0).
+        for builtin_name in self.program.builtins.iter() {
+            if let Some(builtin_runner) = vm_builtin_map.get(builtin_name.name()) {
+                stack.append(&mut builtin_runner.initial_stack());
+            } else {
+                stack.push(MaybeRelocatable::from((0, 0)));
```
fmoletta, line 509: This value is 0 in the `cairo_lang` implementation, not (0, 0)

juanbono:
@odesenfans can you check the comments by @fmoletta ?

vm/src/vm/runners/cairo_runner.rs
odesenfans: After testing a few solutions, here's what's possible.

1. This PR makes it impossible to run a standard program with builtins in the wrong order. Builtins can be specified in any order on the VM struct. We check that the builtins are specified in the standard order in the program and return a `DisorderedBuiltins` error if the order is not respected. The swap of builtins on the VM is now meaningless.

2. As you mentioned `initialize_function_runner`, running it instead of `initialize_builtins` works. I chose this solution for the moment.

fmoletta:
This has been fixed by #1703

PR #1643, opened 2024-03-02
New tracer pr

apoorvsadana:
Description
- Track the ap, fp, pc and the state of the memory in each step of the execution.
- A new crate cairo-vm-tracer which starts a server on the localhost where you can see an web UI. The UI allows you to interact with your Cairo programs and see what's happening inside the VM.
- Most functionality in existing crates has been put behind feature flags with_tracer and tracer


docs/tracer/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Running the tracer
+
+Once the build in the above step is complete, you can use the `cairo-vm-cli` as you do normally with an additional argument `--tracer true`. This tells the VM to start a server where you can visualize the exection of your cairo program.
+
+```bash
+target/release/cairo-vm-cli <path_to_compiled_cairo_json> --layout <layout> --memory_file <path_to_store_memory_binary> --trace_file <path_to_store_tracer_binary> --tracer true
+```
+
+### NOTE
+> The `--memory_file` and `--trace_file` arguments are compulsary at the moment as the current code relocates the memory and trace only when these arguments are supplied. However, edits can be made in future versions to relocate if only the `--tracer` option is specified.
```
fmoletta, line 22:
```suggestion
> The `--memory_file` and `--trace_file` arguments are compulsory at the moment as the current code relocates the memory and trace only when these arguments are supplied. However, edits can be made in future versions to relocate if only the `--tracer` option is specified.
```
Typo

fmoletta: Also, this requirement can be removed by setting the `trace_enabled` & `relocate_mem` flags to true when creating the `CairoRunConfig` in the `run` function of the vm's cli crate.

cairo-vm-cli/src/main.rs
```diff
@@ -54,6 +64,9 @@ struct Args {
     cairo_pie_output: Option<String>,
     #[structopt(long = "allow_missing_builtins")]
     allow_missing_builtins: Option<bool>,
+    #[structopt(long = "tracer")]
+    #[cfg(feature = "with_tracer")]
+    tracer: Option<bool>,
```
fmoletta, line 69: Could we simplify this to `bool`?

pefontana (changes requested):
Amaizing @apoorvsadana !!!!
I try it and works perfect!
Just some points:
* Run the make clippy command that is failing in the CI
* Is necessary to push the .idea folder, if you cant, can we move it inside cairo-vm-tracer/

cairo-vm-cli/src/main.rs
apoorvsadana: fixed

pefontana (approved):
NICEEEEEEE!

fmoletta (approved):
🚀

PR #1646, opened 2024-03-04
[Cairo 1] Add flag to append return values to output segment when not running in proof_mode

fmoletta:
* Adds the flag `append_return_values` to both the CLI and `Cairo1RunConfig` struct.
  * Enabling flag will add the output builtin and the necessary instructions to append the return values to the output builtin's memory segment.

### DIFF
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -102,6 +102,7 @@ pub fn cairo_run_program(
         main_func,
         initial_gas,
         cairo_run_config.proof_mode,
+        cairo_run_config.proof_mode || cairo_run_config.append_return_values,
         cairo_run_config.args,
     )?;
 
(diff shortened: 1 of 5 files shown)


pefontana (approved):

cairo1-run/src/cairo_run.rs
```diff
@@ -334,6 +341,39 @@ fn create_proof_mode_header(builtin_count: i16, return_type_size: i16) -> Vec<In
     ctx.instructions
 }
 
+// Create specific instructions to append the return values to the output segment when not running in proof_mode
+// Call the firt program instruction, appends the return values to the output builtin's memory segment and then returns
```
azteca1998, line 345: If this is intended to be the documentation, then it should be prefixed with the triple slash `///`.

azteca1998 (approved):

igaray (approved):

PR #1651, opened 2024-03-06
Show only layout builtins in air private input

fmoletta:
When converting to serializable format, the air private input contained all the builtins instead of only the ones present in the layout, this worked fine when running with `starknet_with_keccak` layout but produces a different output from cairo_lang when using a different layout that doesn't use all the builtins


pefontana (approved):
LGTM!!!

azteca1998 (approved):

PR #1652, opened 2024-03-06
Run for extra step in proof mode

fmoletta:
Add missing line equivalent to https://github.com/starkware-libs/cairo-lang/blob/efa9648f57568aad8f8a13fbf027d2de7c63c2c0/src/starkware/cairo/lang/vm/cairo_run.py#L343


pefontana:
Lets merge this into #1653, so we merge the fix and the test together.

PR #1653, opened 2024-03-06
Add workflow to compare vm outputs for factorial on all layouts

fmoletta:
Depends on #1651 
Also Fixes:
- Recursive layout is now usable
- Add missing extra step when running in proof mode


pefontana (approved):
Nice!
I will add the workflow to required

azteca1998 (approved):

PR #1657, opened 2024-03-11
Fix error handling in `initialize_state`

fmoletta:
`initialize_state` now returns the correct errors when either `program_base` or `execution_base` are `None`

### DIFF
--- a/vm/src/vm/runners/cairo_runner.rs
+++ b/vm/src/vm/runners/cairo_runner.rs
@@ -446,19 +446,18 @@ impl CairoRunner {
                 .map_err(RunnerError::MemoryInitializationError)?;
 
             // Mark all addresses from the program segment as accessed
-            let base = self
-                .program_base
-                .unwrap_or_else(|| Relocatable::from((0, 0)));
             for i in 0..self.program.shared_program_data.data.len() {
-                vm.segments.memory.mark_as_accessed((base + i)?);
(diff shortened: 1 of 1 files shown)


juanbono (approved):

Oppen (approved):

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -446,19 +446,18 @@ impl CairoRunner {
...
+        } else {
+            return Err(RunnerError::NoProgBase);
         }
         if let Some(exec_base) = self.execution_base {
             vm.segments
                 .load_data(exec_base, &stack)
                 .map_err(RunnerError::MemoryInitializationError)?;
         } else {
-            return Err(RunnerError::NoProgBase);
+            return Err(RunnerError::NoExecBase);
```
Oppen, line 460: Given both return errors if `None`, another implementation could be:
```rust
let prog_base = self.program_base.ok_or(RunnerError::NoProgBase)?;
let exec_base = self.execution_base.ok_or(RunnerError::NoExecBase)?;
// And then the logic without extra indentation or error handling.
```

Oppen: That said, do those always error out? Maybe it could make sense to make them not optional and fail earlier if they're missing?

PR #1662, opened 2024-03-13
Simplify `get_offset_value_reference` and add unit tests

tcoratger:
# Simplify `get_offset_value_reference` and add unit tests

## Description

- The logical test on `base_addr` and `offset` to know if we should return null is implicitly already provided by the Relocatable addition in the last step of the `get_offset_value_reference` function, so that I don't think it's needed is as an overhead.
- Two unit tests are added to test edge configurations.

## Checklist
- [ ] Linked to Github Issue
- [X] Unit tests added

### DIFF
--- a/vm/src/hint_processor/hint_processor_utils.rs
+++ b/vm/src/hint_processor/hint_processor_utils.rs
@@ -165,10 +165,6 @@ fn get_offset_value_reference(
         apply_ap_tracking_correction(vm.get_ap(), var_ap_trackig, hint_ap_tracking)?
     };
 
-    if offset.is_negative() && base_addr.offset < offset.unsigned_abs() as usize {
-        return None;
-    }
-
     if *deref {
         vm.get_maybe(&(base_addr + *offset).ok()?)
     } else {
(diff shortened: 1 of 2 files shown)


fmoletta (approved):

vm/src/hint_processor/hint_processor_utils.rs
```diff
@@ -165,10 +165,6 @@ fn get_offset_value_reference(
         apply_ap_tracking_correction(vm.get_ap(), var_ap_trackig, hint_ap_tracking)?
     };
 
-    if offset.is_negative() && base_addr.offset < offset.unsigned_abs() as usize {
```
pefontana, line 168: Why are we deleting this check?

pefontana (changes requested):

vm/src/hint_processor/hint_processor_utils.rs
tcoratger: Because the logical test on `base_addr` and `offset` to know if we should return null is implicitly already provided by the `Relocatable` addition in the last step of the `get_offset_value_reference` function, so that I don't think it's needed is as an overhead.

fmoletta:
This function no longer exists in the main branch

PR #1666, opened 2024-03-14
perf: cache-friendly `MemoryCell`

Oppen:
Refactors `MemoryCell` to store a simple `[u64; 4]` to use the cache more appropriately.
This structure is now a fixed 32 bytes with 32 bytes alignment, meaning any given cell
in the VM memory will never be split across two different cache lines, and eviction will
be minimized in virtue of being the optimal size for a `Felt` of 252 bits.

The cost of this change is no longer being able to extract the `MemoryCell` value as a
simple reference, but requiring a conversion for the "normal" `MaybeRelocatable`, that
is, it implies extra computation and extra copying to extract values. This behaves well
for Lambdaworks where the data is inline in the array, but quite a hit for the `BigUint`
implementation, which now allocates and frees memory much more often.
Upgrading is recommended only when using Lambdaworks. The tradeoff is deemed acceptable
since only the Lambdaworks (actually `types-rs`) implementation is preserved for the
1.x branch.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1369,6 +1369,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyper_threading"
+version = "0.9.2"
+dependencies = [
+ "cairo-vm",
+ "rayon",
+ "tracing",
+]
+
+[[package]]
 name = "iai-callgrind"
 version = "0.3.1"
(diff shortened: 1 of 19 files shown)


fmoletta:
Already merged #1672

PR #1672, opened 2024-03-18
perf: rework `MemoryCell` for cache efficiency

Oppen:
Store data and metadata inline as a single `[u64; 4]` with `32` bytes alignment, fitting a whole number of cells per cache line to reduce evictions and double sharing and to avoid split line access.

Besides performance, an observable change is that now `Memory::get` always returns a `Cow::Owned` variant because the decoding process always creates a new value.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


juanbono (approved):

fmoletta (approved):
Amazing! 🚀

pefontana (changes requested):
Before merging I think it will be nice to add a workflow that benchmarks the hyper-threading performance, so we can measure the improvement done here

pefontana (approved):

pefontana (changes requested):
We need to check the benchmarks here, they are giving wrong numbers

pefontana (approved):

pefontana:
Benchmarks are going OK now

PR #1679, opened 2024-03-19
Hyper Threading Crate

pefontana:
# Hyper Threading Crate

## Description
New crate to benchmark the performance of Cairo-VM in a hyper-threaded environment

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/examples/hyper_threading/hyper-threading.sh
+++ b/examples/hyper_threading/hyper-threading.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-thread_counts=(1 2 4 5 6 7 8 9 10 11 12 13 14 16 32 )
+thread_counts=(1 2 4 6 8 10 12 16 24 32 )
 binary="target/release/hyper_threading"
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Cairo-VM Changelog
 
 #### Upcoming Changes
+
+* feat: Create hyper_threading crate to benchmark the `cairo-vm` in a hyper-threaded environment [#1679](https://github.com/lambdaclass/cairo-vm/pull/1679)
+
 * feat: add a `--tracer` option which hosts a web server that shows the line by line execution of cairo code along with memory registers [#1265](https://github.com/lambdaclass/cairo-vm/pull/1265)
 
 * feat: Fix error handling in `initialize_state`[#1657](https://github.com/lambdaclass/cairo-vm/pull/1657)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1692,6 +1692,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyper_threading"
+version = "1.0.0-rc1"
+dependencies = [
+ "cairo-vm",
+ "rayon",
+ "tracing",
+]
+
+[[package]]
 name = "iai-callgrind"
 version = "0.3.1"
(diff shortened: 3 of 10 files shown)


examples/hyper_threading/hyper-threading.sh
```diff
@@ -0,0 +1,17 @@
...
+binary="target/release/hyper_threading"
+
+
+cmd="hyperfine -r 1"
+
+# Build the command string with all thread counts
+for threads in "${thread_counts[@]}"; do
+    # For hyperfine, wrap each command in 'sh -c' to correctly handle the environment variable
+    cmd+=" -n \"threads: ${threads}\" 'sh -c \"RAYON_NUM_THREADS=${threads} ${binary}\"'"
+done
```
Oppen, line 13: You can use `hyperfine` own options here:
```shell
  -P, --parameter-scan <VAR> <MIN> <MAX>                                                                                                                                                                                                      
          Perform benchmark runs for each value in the range MIN..MAX. Replaces                                                                                                                                                               
          the string '{VAR}' in each command by the current parameter value.                                                                                                                                                                  
                                                                                                                                                                                                                                              
            Example:  hyperfine -P threads 1 8 'make -j {threads}'                                                                                                                                                                            
                                                                                                                                                                                                                                              
          This performs benchmarks for 'make -j 1', 'make -j 2', …, 'make -j 8'.                                                                                                                                                              
                                                                                                                                                                                                                                              
          To have the value increase following different patterns, use shell                                                                                                                                                                  
          arithmetics.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                              
            Example: hyperfine -P size 0 3 'sleep $((2**{size}))'                                                                                                                                                                             
                                                                                                                                                                                                                                              
          This performs benchmarks with power of 2 increases: 'sleep 1', 'sleep                                                                                                                                                               
          2', 'sleep 4', …                                                                                                                                                                                                                    
          The exact syntax may vary depending on your shell and OS.                                                                                                                                                                           
  -D, --parameter-step-size <DELTA>                                                                                                                                                                                                           
          This argument requires --parameter-scan to be specified as well.                                                                                                                                                                    
          Traverse the range MIN..MAX in steps of DELTA.                                                                                                                                                                                      
                                                                                                                                                                                                                                              
            Example:  hyperfine -P delay 0.3 0.7 -D 0.2 'sleep {delay}'                                                                                                                                                                       
                                                                                                                                                                                                                                              
          This performs benchmarks for 'sleep 0.3', 'sleep 0.5' and 'sleep 0.7'.                                                                                                                                                              
  -L, --parameter-list <VAR> <VALUES>                                                                                                                                                                                                         
          Perform benchmark runs for each value in the comma-separated list                                                                                                                                                                   
          VALUES. Replaces the string '{VAR}' in each command by the current                                                                                                                                                                  
          parameter value.                                                                                             
                                                                                                                       
          Example:  hyperfine -L compiler gcc,clang '{compiler} -O2 main.cpp'                                          
                                                                                                                       
          This performs benchmarks for 'gcc -O2 main.cpp' and 'clang -O2                                               
          main.cpp'.                                                                                                   
                                                                                                                       
          The option can be specified multiple times to run benchmarks for all
          possible parameter combinations.                    
```

Oppen: There's also no need to fire up the shell explicitly. The default behavior is to spawn one, unless you pass the `-N` option.

juanbono (approved):

Oppen (approved):

PR #1681, opened 2024-03-20
Update segment used offsets with output size.

Alon-Ti:
Update output segment used size with return value size when it is known ahead.


fmoletta (commented):
Hello! It looks like this logic is already present in `cairo1-run`'s `cairo_run_program` method (cairo1-run/src/cairo_run.rs lines 202-208). Is this method being used from an outside crate? If so we can consider removing the hidden attribute and porperly documenting the method and it's use case.

fmoletta (approved):

pefontana (approved):

PR #1686, opened 2024-03-24
feat: Reorganized builtins to be in the top of stack at the end of run.

orizi:
# TITLE
Builtins on top of stack at the end of cairo1 runs.

## Description
Removed the extra wrapping function call, and make a single wrapping entrypoint code, that also copies data into the output_ptr if required, and then rewrites all the builtins (used and unused to top of stack).

This hopefully will make the running of the proof more usable.


pefontana (commented):
Thank you for the PR, @orizi !

I've noticed that this change introduces some issues when running programs in cairo-1 with --proof_mode. When we execute a Cairo 1 program with proof_mode, the execution never completes. Do you have any insights on why this might be happening?

```
cd cairo1-run
cp ../cairo_programs/cairo-1-programs/fibonacci.cairo ./

cargo run \
fibonacci.cairo \
--layout=small \
--air_public_input=fibonacci_public_input.json \
--air_private_input=fibonacci_private_input.json \
--trace_file=fibonacci_trace.json \
--memory_file=fibonacci_memory.json \
--print_output \
--proof_mode
```

orizi:
Found the culprit and fixed as well.

fmoletta:
Hello! We noticed some problems when running with proof_mode and when serializing outputs that makes tests fail, I sent a PR to your branch with fixes for these issues https://github.com/orizi/cairo-vm/pull/1

fmoletta (approved):

PR #1690, opened 2024-03-27
Feature: add a method to retrieve the output builtin from the VM

odesenfans:
Problem: the output builtin often needs to be manipulated directly in the Starknet bootloader and OS hints.

Solution: add a `get_output_builtin() method on the `VirtualMachine` struct to retrieve a reference to the output builtin easily.

## Checklist
- [ ] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -973,6 +973,7 @@ dependencies = [
  "clap",
  "itertools 0.11.0",
  "mimalloc",
+ "num-traits 0.2.18",
  "rstest",
  "thiserror",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -973,7 +973,6 @@ dependencies = [
  "clap",
  "itertools 0.11.0",
  "mimalloc",
- "num-traits 0.2.18",
  "rstest",
  "thiserror",
 ]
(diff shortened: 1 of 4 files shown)


vm/src/vm/vm_core.rs
```diff
@@ -942,6 +942,17 @@ impl VirtualMachine {
 
         Err(VirtualMachineError::NoSignatureBuiltin)
     }
+
+    pub fn get_output_builtin(&mut self) -> Result<&mut OutputBuiltinRunner, VirtualMachineError> {
```
juanbono, line 946: do you really need a mutable reference here?

Cargo.lock
Oppen, line 1: Why does this file change?

odesenfans: Good question, this line appeared automatically when recompiling. I can discard the change from the PR but there's probably something behind that.

juanbono (approved):

vm/src/vm/vm_core.rs
juanbono: Ohh I see.

fmoletta, line 946: Can we name it `get_output_builtin_mut` instead? As we have similar named methods that return immutable references to other builtins and it may be confusing

odesenfans: Fixed.

fmoletta (approved):

PR #1693, opened 2024-04-04
style: code consistency

jujube:
https://github.com/lambdaclass/cairo-vm/blob/22a97dcedd58da60ceb2ae64f8c021208edc50b5/vm/src/cairo_run.rs#L74-L76


fmoletta:
This change is already on the main branch

PR #1696, opened 2024-04-04
perf: delete useless lifetime

jujube:


Oppen:
That would be a small cleanup. Is there evidence that this affects performance?

jujube:
no.
The main code is clearer and does not need to display  lifetime.

Oppen:
> no. The main code is clearer and does not need to display lifetime.

I agree the code is slightly clearer without redundant lifetimes. But the question points to how the PR is classified. If there is no evidence of performance changes, why is it marked as perf?

PR #1673, opened 2024-03-18
Add mod builtin

fmoletta:
Implements `ModBuiltinRunner`
Adds `add_mod` & `mul_mod` to layouts `all_cairo` & `dynamic` under `mod_builtin` feature
Removes unused and duplicated code from builtin_runner module

PR #1698, opened 2024-04-04
Add range_check96 builtin

fmoletta:
Refactors RangeCheckBuiltinRunner to accomodate both standard and 96 variants
Adds range_check96 to all_cairo layout
Integration test added in #1699 as it depends on #1673


juanbono (approved):

pefontana (approved):
Amaizing one @fmoletta !!

PR #1700, opened 2024-04-05
fix: add cairo-compile fibonacci.cairo

jujube:
add --proof_mode


fmoletta:
There is already a section above explaining how to compile a cairo program using the cairo compiler, there is no need to add the compilation command in this section

PR #1711, opened 2024-04-12
Add Hyper Threading benchmarks workflow

pefontana:
# Add Hyper Threading benchmarks workflow

## Description

Add a workflow to benchmark the Cairo-vm in a  Hyper-Threading context

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


.github/workflows/hyper_threading_benchmarks.yml
```diff
@@ -0,0 +1,123 @@
...
+      uses: actions/download-artifact@v4
+      with:
+        name: hyper_threading_workflow_script
+        path: ${{ github.workspace }}/
+
+    - name: Download hyper_threading_main_binary
+      uses: actions/download-artifact@v4
+      with:
+        name: hyper_threading_main_binary
+        path: ${{ github.workspace }}/
```
Oppen, line 84: This is unnecessary. Since you're running everything in a single job you still have the files where you copied them. This would only be necessary if we did parallel jobs here.

.github/workflows/hyper_threading_benchmarks.yml
```diff
@@ -0,0 +1,123 @@
...
+        name: hyper_threading_pr_binary
+        path: ${{ github.workspace }}/hyper_threading_pr
+
+    - name: Upload Workflow Script
+      uses: actions/upload-artifact@v4
+      with:
+        name: hyper_threading_workflow_script
+        path: ${{ github.workspace }}/hyper-threading-workflow.sh
+
+
```
Oppen, line 51: Same for this.

.github/workflows/hyper_threading_benchmarks.yml
pefontana: Thank @Oppen !!
Since we are doing a git checkout main (To create the other binary and compare it vs the branch one) I still need to download and upload the PR binary and the workflow.sh. But I was able to remove the download and upload of the main binary 
https://github.com/lambdaclass/cairo-vm/pull/1711/commits/0726f23e5abeccb49a2ee0e01d8f3de69c3345ba

Oppen: Right. We may be able to remove that too in a later PR, when this is in main.

Oppen: Hold on. We can make the copy to the workspace before checking out main. That should solve it.

fmoletta (approved):

PR #1712, opened 2024-04-14
Enabling SegmentArena usage.

orizi:
# Enabling Segment Arena Usage

## Description
Enabled actually using the segment-arena builtin, and used it in cairo-1 runner.
Without these changes not actually usable.

## Checklist
- [V] This change requires new documentation.
  - [V] CHANGELOG has been updated.


fmoletta (changes requested):
Adding the `segment_arena` builtin directly to the layouts breaks a lot of our tests (and breaks compatibility with cairo_lang). Could we solve this in another way? Maybe we could enable it via a flag instead?

PR #1715, opened 2024-04-16
Use an enum to represent layout name

fmoletta:
Uses an enum instead of a String to represent cairo layouts.
This makes layout matching faster and improves parsing of cli arguments
String representation is still left in `PublicInput` struct, this can be changed once #1716 is merged


vm/src/types/layout_name.rs
```diff
@@ -0,0 +1,68 @@
...
+    starknet,
+    starknet_with_keccak,
+    recursive_large_output,
+    all_solidity,
+    all_cairo,
+    dynamic,
+}
+
+impl LayoutName {
+    pub fn to_str(&self) -> &'static str {
```
Oppen, line 26: Since it's a fieldless enum, consider deriving `Copy` and making the methods take the value:
```suggestion
    pub fn to_str(self) -> &'static str {
```

juanbono (approved):
Great!

pefontana (approved):
LGTM!

PR #1717, opened 2024-04-16
[WIP] Set max combination size for check-all-features to speed up CI

fmoletta:
The workflow `Make sure all features work (vm)` is already taking close to 50 minutes. As there is no way to partition the workflow into sets of features, this PR aims to fix this problem by setting up a max combination of features to reduce the cases to check.
Follow-up work: Investigate other workflows/ solutions that can allow us to properly partition this workflow


gabrielbosio:
Last execution of `Make sure all features work (vm)` in `main` took ~25 min so I think this change is not needed anymore

PR #1718, opened 2024-04-16
Remove unsafe impl of `Add<usize> for &'a Relocatable`

fmoletta:
The impl of `Add<usize> for  Relocatable` should be used instead, which uses `checked_add` and handles potential overflows
Closes #1696


Oppen (approved):

Oppen:
I think the `unwrap_or_default` calls in the relocation functions becomes a silent failure. Maybe they should bubble up an error instead.

juanbono (approved):

PR #1721, opened 2024-04-17
Added segment arena cairo1 runner validation.

orizi:
# Added Cairo1 run minibootloader for segment arena validation.

## Description
Using the cairo1 repo casm builder for easier CASM writing for writing the validation.
Added finalization of dicts in cairo1 hint processor.
Used both to add the validation in the runner.
Updated the tests - as some of the tests returned a Dictionary from main - which isn't sound for proving code (as no squash occured).

## Checklist
- [V] This change requires new documentation.
  - [V] CHANGELOG has been updated.

### DIFF
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -1,5 +1,11 @@
 use cairo_lang_casm::{
-    casm, casm_extend, hints::Hint, inline::CasmContext, instructions::Instruction,
+    builder::{CasmBuilder, Var},
+    casm, casm_build_extend,
+    cell_expression::CellExpression,
+    deref, deref_or_immediate,
+    hints::Hint,
+    inline::CasmContext,
+    instructions::{Instruction, InstructionBody},
 };
 use cairo_lang_sierra::{
     extensions::{
@@ -14,7 +20,7 @@ use cairo_lang_sierra::{
         starknet::syscalls::SystemType,
         ConcreteType, NamedType,
     },
-    ids::ConcreteTypeId,
+    ids::{ConcreteTypeId, GenericTypeId},
     program::{Function, GenericArg, Program as SierraProgram},
     program_registry::ProgramRegistry,
 };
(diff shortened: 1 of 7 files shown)


fmoletta (changes requested):
The cairo programs returning non-squashed dicts are mainly used to test output serialization, and correspond to issues raised by users of the crate. I would prefer to keep them like this to ensure that the same issues don't pop up in the future.

orizi:
The serialization of such dictionaries *should not be supported* - it is never a valid run.
If you want these as tests - the tests should make sure you fail.

orizi:
Specifically - the related issue should have the same effect for checking on SquashedDict - i simply add that a regular `Dict` should never be returned - as it didn't pass a squash - and therefore not provable.

cairo1-run/src/cairo_run.rs
```diff
@@ -356,97 +362,112 @@ fn create_entry_code(
     // The builtins in the formatting expected by the runner.
     let (builtins, builtin_offset) =
         get_function_builtins(&signature.param_types, copy_to_output_builtin);
-    let mut ctx = casm! {};
+    let mut ctx = CasmBuilder::default();
```
pefontana, line 365: @orizi Can you add some documentation here?
Specially in the casm_extend! cases?

cairo1-run/src/cairo_run.rs
```diff
@@ -483,98 +504,122 @@ fn create_entry_code(
...
+        | BuiltinName::mul_mod => unreachable!(),
+    };
+    if copy_to_output_builtin && got_segment_arena {
+        for (i, builtin) in builtins.iter().enumerate() {
+            let var = get_var(builtin);
+            let local = ctx.add_var(CellExpression::Deref(deref!([fp + i.to_i16().unwrap()])));
+            casm_build_extend!(ctx, assert local = var;);
+        }
+        let segment_arena_ptr = get_var(&BuiltinName::segment_arena);
+        casm_build_extend! {ctx,
```
pefontana, line 550: Here too, 
A little documentation here will be really usefull

cairo_programs/cairo-1-programs/dict_with_struct.cairo
pefontana, line 1: Since this Dict return programs are still running without proof_mode, can we made some test for the old version of the cairo program that runs OK without proof_mode and fails when proof_mode?

cairo1-run/src/cairo_run.rs
```diff
@@ -483,98 +504,122 @@ fn create_entry_code(
...
-        casm_extend!(ctx, ret;);
+        casm_build_extend!(ctx, ret;);
     }
-
-    assert_eq!(before_final_call + final_call_size, ctx.current_code_offset);
-
-    Ok((ctx, builtins))
+    let result = ctx.build(["FUNCTION"]);
+    let [call_inst] = result.branches[0].1.as_slice() else {
+        panic!("Expected a single relocation");
```
pefontana, line 602: Can we avoid the panics here? If it is to much work just create an issue and we will handle it

pefontana (changes requested):

cairo1-run/src/cairo_run.rs
orizi: this cannot ever actually panic - given that the builder code is correct (it will either always crash if we've written bad code here) or never crash (if the code here is fine - which it is)

cairo_programs/cairo-1-programs/dict_with_struct.cairo
orizi: it shouldn't work at any version - it is a legal cairo function - but not a sensible run.
even when not running in proof mode i would not expect to be able to run unprovable programs, unless these are actually a test.

cairo1-run/src/cairo_run.rs
orizi: do note this is *exactly* the same code as before - only where the code may somewhat be understandable by variable name.

add some additional doc.

pefontana (approved):
Great @orizi !
Thanks for the work!
We just tested proving some dict programs with these changes + the gas related ones + Stone Prover and work OK

fmoletta (approved):

PR #1722, opened 2024-04-17
Use `BuiltinName` enum instead of string representation

fmoletta:
Replaces all uses of `String` & `&'static str` representing builtin names with `BuiltinName` enum.
Implements methods to convert `BuiltinName` to and from str with and without suffix, adding documentation & examples to reduce confusion for users due to the breaking changes (To see all breaking changes, checkout the CHANGELOG entry for this PR).
Implements generic serialization & deserialization for HashMaps containing `BuiltinName` that use a suffixed version of the builtin names.

This PR will make enforce valid builtin names across the codebase, removing the need for extra validations, and will be particularly useful for ensuring that cairo pies are valid before running them (#1720)


Oppen (approved):

cairo1-run/src/cairo_run.rs
```diff
@@ -724,13 +716,13 @@ fn finalize_builtins(
     for (id, size) in ret_types_and_sizes {
         if let Some(ref name) = id.debug_name {
             let builtin_name = match &*name.to_string() {
```
juanbono, line 718:
```suggestion
            let builtin_name = match name.as_str() {
```

juanbono: this can avoid an small allocation

juanbono (approved):

PR #1726, opened 2024-04-19
Implement `extend_additional_data` for `BuiltinRunner`

fmoletta:
Needed for #1720

### DIFF
--- a/vm/src/vm/runners/builtin_runner/mod.rs
+++ b/vm/src/vm/runners/builtin_runner/mod.rs
@@ -499,6 +499,7 @@ impl BuiltinRunner {
         }
     }
 
+    /// Returns data stored internally by builtins needed to re-execute from a cairo pie
     pub fn get_additional_data(&self) -> BuiltinAdditionalData {
         match self {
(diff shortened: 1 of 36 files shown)


vm/src/vm/runners/builtin_runner/mod.rs
```diff
@@ -508,6 +508,18 @@ impl BuiltinRunner {
         }
     }
 
+    pub fn extend_additional_data(
```
juanbono, line 511: wen docs

vm/src/vm/runners/builtin_runner/output.rs
```diff
@@ -124,6 +124,19 @@ impl OutputBuiltinRunner {
         })
     }
 
+    pub fn extend_additional_data(
```
juanbono, line 127: I think we should have some docs about what is the intent of this function

juanbono (approved):

juanbono (approved):
LGTM

pefontana (approved):

PR #1729, opened 2024-04-23
Implement `CairoPie::read_zip_file`

fmoletta:
Implements custom deserialization for structs using custom serialization.
Implement method to deserialize a CairoPie from a zip file
Move custom serialization functions into sub-modules along with their added deserialization counterparts to make use of the `serde(with)` attribute and make code cleaner

### DIFF
--- a/vm/src/vm/runners/cairo_pie.rs
+++ b/vm/src/vm/runners/cairo_pie.rs
@@ -186,7 +186,6 @@ pub(super) mod serde_impl {
     use crate::types::builtin_name::BuiltinName;
     use num_integer::Integer;
     use num_traits::Num;
-    use serde::ser::SerializeMap;
 
     use super::CAIRO_PIE_VERSION;
     use super::{CairoPieMemory, SegmentInfo};
@@ -199,7 +198,11 @@ pub(super) mod serde_impl {
         Felt252,
     };
     use num_bigint::BigUint;
-    use serde::{de::Error, ser::SerializeSeq, Deserialize, Deserializer, Serialize, Serializer};
+    use serde::{
+        de::Error, ser::SerializeMap, ser::SerializeSeq, Deserialize, Deserializer, Serialize,
+        Serializer,
+    };
+    use serde_json::Number;
 
     pub const ADDR_BYTE_LEN: usize = 8;
     pub const FIELD_BYTE_LEN: usize = 32;
@@ -250,7 +253,6 @@ pub(super) mod serde_impl {
 
     pub mod program_data {
         use super::*;
-        use crate::types::relocatable::MaybeRelocatable;
 
         pub fn serialize<S>(values: &[MaybeRelocatable], serializer: S) -> Result<S::Ok, S::Error>
         where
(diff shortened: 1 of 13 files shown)


vm/src/vm/runners/cairo_pie.rs
```diff
@@ -134,65 +133,187 @@ impl CairoPie {
...
+        let additional_data: CairoPieAdditionalData = serde_json::from_slice(&additional_data)?;
+
+        Ok(CairoPie {
+            metadata,
+            memory,
+            execution_resources,
+            additional_data,
+            version,
+        })
+    }
```
Oppen, line 181: This could have a big memory peak from all the `read_to_end` calls. Consider deserializing from the readers directly, maybe with a `BufReader`.

Oppen: Alternatively, use and clear a single buffer so at least we don't have all the data in temporaries at the same time.

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -272,61 +393,117 @@ mod serde_impl {
...
+        ) -> Result<HashMap<Relocatable, (Felt252, Felt252)>, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            let number_map = Vec::<((Number, Number), (Number, Number))>::deserialize(d)?;
+            let mut res = HashMap::new();
+            for ((index, offset), (r, s)) in number_map.into_iter() {
+                let addr = Relocatable::from((
+                    index.as_u64().ok_or(D::Error::custom("Invalid address"))? as isize,
+                    offset.as_u64().ok_or(D::Error::custom("Invalid address"))? as usize,
```
Oppen, line 468: Are these errors allocated on the heap? Maybe use `ok_or_else` so we only pay for them when actually needed?

Oppen:
There seems to be a significant performance regression.

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -272,61 +393,117 @@ mod serde_impl {
...
+        }
+
+        pub fn deserialize<'de, D>(
+            d: D,
+        ) -> Result<HashMap<Relocatable, (Felt252, Felt252)>, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            let number_map = Vec::<((Number, Number), (Number, Number))>::deserialize(d)?;
+            let mut res = HashMap::new();
```
Oppen, line 464:
```suggestion
            let mut res = HashMap::with_capacity(number_map.len());
```

fmoletta:
> There seems to be a significant performance regression.

This is quite strange, the benchmarks don't produce or consume Cairo PIEs so they shouldn't be affected by this PR

Oppen:
> This is quite strange, the benchmarks don't produce or consume Cairo PIEs so they shouldn't be affected by this PR

Whatever it was it seems to have been reversed by new commits. Maybe it was throttling in the worker.

juanbono (approved):

Oppen (approved):

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -134,65 +133,176 @@ impl CairoPie {
...
+        let reader = std::io::BufReader::new(zip_reader.by_name("version.json")?);
+        let version: CairoPieVersion = serde_json::from_reader(reader)?;
+
+        let reader = std::io::BufReader::new(zip_reader.by_name("metadata.json")?);
+        let metadata: CairoPieMetadata = serde_json::from_reader(reader)?;
+
+        let mut memory = vec![];
+        zip_reader.by_name("memory.bin")?.read_to_end(&mut memory)?;
+        let memory = CairoPieMemory::from_bytes(&memory)
+            .ok_or_else(|| std::io::Error::from(std::io::ErrorKind::InvalidData))?;
```
Oppen, line 155:
```suggestion
        let reader = std::io::BufReader::new(zip_reader.by_name("memory.bin")?);
        let memory = CairoPieMemory::from_reader(reader)
            .ok_or_else(|| std::io::Error::from(std::io::ErrorKind::InvalidData))?;
```

PR #1739, opened 2024-04-26
`cairo1-run` CLI: Allow loading arguments from file

fmoletta:
Adds the `--args_file` that takes a filename and processes the arguments in the file as if they were sent via `--args`


igaray (approved):

pefontana:
nice one!

PR #1741, opened 2024-04-29
[Cairo 1 Run] Refactor integration tests + check that return values taken from output segment are correct

fmoletta:
After PR #1686 the return values are now fetched from the output segment instead of the execution segment when either the `append_return_values` or `proof_mode` flags are enabled, this makes our `check_append_ret_values_to_output_segment` not as useful as it no longer checks that the correct values are being copied to the output segment.
A way to fix it would be to instead compare the values from the execution segment to the ones on the output segment, but after PR #1721 when the segment arena is used the execution segment now contains the values produced by the arena validation where the return values used to be found, so we can no longer use it for comparison.
As a result of these two changes, the best way to test that the correct values are copied to the output segment are to 1: test that the output segment indeed has the return values outputted and no more values after them (current behaviour), and 2: test that the return values outputted when running with `--append_return_values` match the ones outputted by a normal run (this can be accomplished by adding a third case with `--append_return_values`  enabled to our in integration tests.
This PR adds this third test case and also refactors the integration tests to into one test with multiple cases & values so adding new checks and argument combinations is easier.

### DIFF
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -1170,24 +1170,24 @@ mod tests {
         compile_prepared_db(&mut db, main_crate_ids, compiler_config).unwrap()
     }
 
-    fn main_hash_panic_result(sierra_program: &SierraProgram) -> bool {
+    fn return_value_size(sierra_program: &SierraProgram) -> i16 {
+        let sierra_program_registry =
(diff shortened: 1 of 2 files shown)


pefontana (approved):
nice one!

igaray (approved):

PR #1744, opened 2024-04-30
Add hint `U256InvModN` to `Cairo1HintProcessor`

fmoletta:
Take hit code from cairo repo + add test

### DIFF
--- a/vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs
+++ b/vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs
@@ -253,62 +253,19 @@ impl Cairo1HintProcessor {
                 s_or_r1,
                 t_or_k0,
                 t_or_k1,
-            })) => {
-                let pow_2_128 = BigInt::from(u128::MAX) + 1u32;
-                let b0 = get_val(vm, b0)?.to_bigint();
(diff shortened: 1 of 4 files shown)


vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs
```diff
@@ -242,6 +242,73 @@ impl Cairo1HintProcessor {
                 vm, dividend0, dividend1, dividend2, dividend3, divisor0, divisor1, quotient0,
                 quotient1, quotient2, quotient3, remainder0, remainder1,
             ),
+            Hint::Core(CoreHintBase::Core(CoreHint::U256InvModN {
```
pefontana, line 245: Can we move the implementation inside a Cairo1HintProcessor method? like we do with the rest of the hints

pefontana (changes requested):

vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs
fmoletta: Updated

pefontana (approved):

juanbono (approved):

PR #1756, opened 2024-05-08
perf: Change `CairoRunError::VmException` to `Box<VmException>` 

fmoletta:
PR #1720 Added a small error variant to the `CairoRunError` which brought a huge performance regression. This is due to the `VmException` variant having a big size, making all other variants equally as big. This PR solves this issue by wrapping the `VmException` contained in its corresponding variant, and adds a test to ensure that the size of CairoRunError doesn't surpass 32 bytes

### DIFF
--- a/vm/src/cairo_run.rs
+++ b/vm/src/cairo_run.rs
@@ -72,7 +72,13 @@ pub fn cairo_run_program(
 
     cairo_runner
         .run_until_pc(end, &mut vm, hint_executor)
-        .map_err(|err| VmException::from_vm_error(&cairo_runner, &vm, err))?;
+        .map_err(|err| {
+            CairoRunError::VmException(Box::new(VmException::from_vm_error(
+                &cairo_runner,
+                &vm,
+                err,
+            )))
+        })?;
 
     if cairo_run_config.proof_mode {
         cairo_runner.run_for_steps(1, &mut vm, hint_executor)?;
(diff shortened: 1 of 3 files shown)


vm/src/cairo_run.rs
```diff
@@ -153,7 +159,13 @@ pub fn cairo_run_pie(
 
     cairo_runner
         .run_until_pc(end, &mut vm, hint_processor)
```
Oppen, line 161: I think I identified why this doesn't work. By this point the `Result<_, VirtualMachineError>` is already returned. What we need to make smaller is the `VirtualMachineError` itself.
Also, `ProgramError` is potentially big and contributes to `CairoRunError`'s size as well. This is because of the `IO` and `Parse` variants as well as all of the ones containing `String`.
It shouldn't matter for the common case as this happens once at the beginning and once at the end. However, `VirtualMachineError` is likely to be returned in several places.

Oppen: `RunnerError::PageNotOnSegment(Relocatable, usize)` is one candidate to boxing.

fmoletta: `VirtualMachineError` is 32 bits already, should we make it even smaller?

Oppen: Bytes, but yeah, it's used in many places. I'm not sure the error is the culprit though.

PR #1759, opened 2024-05-09
[Cairo 1] Load arguments into VM instead of creating them via instructions

fmoletta:
**Changes**
Adds the function `load_arguments` which loads the input arguments (if any) into the VM after initialization and removes the instructions that create these arguments in `create_entry_code`. This change has no effect on the execution segment.
Adds test to make sure that the program_hash does not depend on the arguments used to run it.

**Motive & Context**
This PR solves the issue of the program_hash changing depending on the arguments passed to the vm when running it. This happens due to how cairo 1 handles arguments. Cairo 0 doesn't have arguments when running a program, this was circumvented by Cairo 1 by adding instructions to the original program that declare the arguments as constants and insert them into the execution segment (via tempvars) before calling the main function (they also add the builtin and segment arena pointers). While this allows running cairo 1 programs with arguments, it also makes the program (and therefore the program_hash) vary depending on the arguments used, as the arguments are now part of the instructions.
This PR is the later of 3 approaches to solve this issue, the previous three attempts consisted on:

- Removing the create_entry_code altogether #1746  (This run into issues due to the segment arena builtin)

- Replacing all instructions that create values in the execution segment with loading them directly into the VM before the run #1751 (This worked similarly to this PR, but aside from loading the input arguments, it also loaded the builtin bases & segment arena values and removed the instructions that created them. While this solution worked (aka it allowed proving using Stone Prover), it involves more values being _magically_ created, which can reduce the integrity of the proof (see discussion & possible issues))

**Discussion & possible issues**
As arguments are now loaded into the memory directly instead of being created by a cairo instruction, this could be seen as a `hint` being the one that creates those values in memory, which means that we won't be able to prove that it was those specific input values that produced an output based on a program as the prover doesn't currently take these arguments as public inputs.


Oppen (approved):

pefontana (approved):
Nice one!

Okm165:
Really good stuff this enables private inputs to the circuits! Before everything was visible in bytecode

raphaelDkhn:
@fmoletta Does this mean that no Cairo1 program supports public input? So it's not possible to prove that a program was executed with specific public inputs with Cairo1?

raphaelDkhn:
@fmoletta 

Since program arguments are no longer included in the `public inputs` for the prover, one way to verify that a circuit has been executed correctly with specific arguments would to return those arguments in the program's output, right?. This way, we ensure that the arguments become part of the `public inputs`, similar to how `serialize_word` in CairoZero.

Is it correct?

PR #1764, opened 2024-05-13
Serialize `Array<Felt252>` return value into output segment

fmoletta:
* Checks that only `PanicResult<Array<Felt252>>` or `Array<Felt252>` can be returned by the program when running with either `--proof_mode` or `--append_return_values`.
* Serializes return values under the previous conditions following the format:
`PanicResult<Array<Felt252>>` -> [panic_flag, array_len, arr[0], arr[1],.., arr[n]]
`t<Array<Felt252>` -> [array_len, arr[0], arr[1],.., arr[n]]
* Creates mirror versions of all test programs modified to return a serialized output so we can keep the same level of testing


cairo1-run/src/cairo_run.rs
```diff
@@ -738,54 +804,124 @@ fn get_function_builtins(
...
         let inner_args = &return_type_info.long_id.generic_args[1];
-        let inner_type = match inner_args {
-            GenericArg::Type(type_id) => type_id,
-            _ => unreachable!(),
+        let inner_type = {
+            let inner_type = match inner_args {
+                GenericArg::Type(type_id) => type_id,
+                _ => unreachable!(),
+            };
+            // The inner type contains a single-element tuple so we ned to get rid of it too
```
juanbono, line 862: typo: need

cairo1-run/src/error.rs
```diff
@@ -59,4 +59,6 @@ pub enum Error {
         param_index: usize,
         arg_index: usize,
     },
+    #[error("Only programs returing `Array<Felt252>` can be currently proven. Try serializing the final values before returning them")]
```
juanbono, line 62: typo: returning

juanbono (approved):

pefontana (approved):

PR #1770, opened 2024-05-22
Serialize inputs into output segment

fmoletta:
Depends on #1764 
* Checks that only `Array<Felt252>` can be received by the program main function when running with either  `--proof_mode` or `--append_return_values`.
* Copies the  input value to the output segment right after the output in the format [array_len, arr[0], arr[1],.., arr[n]].

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 #### Upcoming Changes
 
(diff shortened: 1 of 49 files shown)


juanbono (approved):
Great!

pefontana (approved):

PR #1775, opened 2024-05-24
[Cairo 1] Filter implicit arguments from return value

fmoletta:
Currently, we take the last ret_type from the function's signature as the function's return type. This works well for programs that have a return values, or programs that don't return anything but don't use any implicit arguments. But when we run a program with implicit arguments (such as builtin pointers), the last ret_type will be an implicit return value's type, and the cli will panick when trying to serialize the program output.
This PR checks if the last ret_type is an implicit type (aka builtins, gas, system type) and sets the return type to None if true.

### DIFF
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -133,7 +133,16 @@ pub fn cairo_run_program(
     let initial_gas = 9999999999999_usize;
 
     // Fetch return type data
-    let return_type_id = main_func.signature.ret_types.last();
+    let return_type_id = match main_func.signature.ret_types.last() {
+        // We need to check if the last return type is indeed the function's return value and not an implicit return value
+        return_type @ Some(concrete_ty)
+            if get_info(&sierra_program_registry, concrete_ty)
+                .is_some_and(|info| is_implicit_generic_id(&info.long_id.generic_id)) =>
+        {
+            return_type
+        }
+        _ => None,
+    };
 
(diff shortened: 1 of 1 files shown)


pefontana (approved):

juanbono (approved):

cairo1-run/src/cairo_run.rs
```diff
@@ -885,6 +886,19 @@ fn check_only_array_felt_input_type(
...
+fn is_implicit_generic_id(generic_ty: &GenericTypeId) -> bool {
+    generic_ty == &SegmentArenaType::ID
+        || generic_ty == &GasBuiltinType::ID
+        || generic_ty == &BitwiseType::ID
+        || generic_ty == &EcOpType::ID
+        || generic_ty == &PedersenType::ID
+        || generic_ty == &PoseidonType::ID
+        || generic_ty == &RangeCheckType::ID
+        || generic_ty == &SegmentArenaType::ID
+        || generic_ty == &SystemType::ID
```
Oppen, line 900: Nit: a shorter way to express the same:
```suggestion
    [
        SegmentArenaType::ID,
        GasBuiltinType::ID,
        BitwiseType::ID,
        EcOpType::ID,
        PedersenType::ID,
        PoseidonType::ID,
        RangeCheckType::ID,
        SegmentArenaType::ID,
        SystemType::ID,
    ].contains(generic_ty)
```

pefontana: better than clippy

PR #1781, opened 2024-06-03
Fixed CairoPie output builtin data deserlialization.

Alon-Ti:
Public memory pages were deserialized as structs, whereas CairoPies have them as JSON arrays of length 2.

PR #1783, opened 2024-06-04
Don't validate memory values

Alon-Ti:
Remove validation of values in CairoPie Memory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* fix: Remove validation of CairoPie memory values [#1783](https://github.com/lambdaclass/cairo-vm/pull/1783)
+
 * fix: Handle `GasBuiltin` in cairo1-run crate [#1789](https://github.com/lambdaclass/cairo-vm/pull/1789)
   * Load `initial_gas` into vm instead of creating it via instructions.
   * Fix bug affecting programs with input arguments and gas builtin.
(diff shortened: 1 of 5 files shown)


juanbono (changes requested):
Can you add some tests to this?

Alon-Ti:
@juanbono, I want to add a pathological CairoPie end to end test that will cover this and the previous issue against python VM output (see slack), I can do that once #1781 is merged.

Alon-Ti:
@juanbono I amended an existing test to cover this, will add a test CairoPie with all the edge cases once I'm sure all our test CairoPies run.

cairo_programs/relocate_segments.cairo
```diff
@@ -12,14 +12,14 @@ func main() {
...
-    assert array[1] = 51;
+    assert array[1] = cast(array, felt) + 100;
 
     // Realocate temporary_array into the array segment
     relocate_segment(src_ptr=temporary_array, dest_ptr=array);
 
     // Assert that the realocated temporary_array gets their values from the array segment
     assert temporary_array[0] = 50;
-    assert temporary_array[1] = 51;
+    assert temporary_array[1] = cast(array, felt) + 100;
```
fmoletta, line 22: Could we have this as a separate test instead?

Oppen (approved):

juanbono (approved):

fmoletta (approved):

PR #1785, opened 2024-06-06
Fixed serializations for range_check96 and empty signautre additional

Alon-Ti:
Fix bug where an empty signature array in bulitin additional data was deserialized as Hash, and name of range_check96 builtin.


vm/src/vm/runners/cairo_pie.rs
```diff
@@ -72,10 +72,11 @@ pub struct OutputBuiltinAdditionalData {
     pub attributes: Attributes,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Clone, Debug, Eq)]
 #[serde(untagged)]
 pub enum BuiltinAdditionalData {
     // Contains verified addresses as contiguous index, value pairs
+    Empty([usize; 0]),
```
Oppen, line 79: Why not just use an empty variant here?

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -85,6 +86,45 @@ pub enum BuiltinAdditionalData {
...
+                if let BuiltinAdditionalData::Output(other_data) = other {
+                    data == other_data
+                } else {
+                    false
+                }
+            }
+            BuiltinAdditionalData::None => matches!(other, BuiltinAdditionalData::None),
+        }
+    }
+}
```
Oppen, line 126: I think this could be simplified by:
1. Creating an `is_empty` method that checks only that for all variants.
2. Then doing that check first (with early return `true` if both are empty and early return `false` if one is but the other isn't) and otherwise requires both the variant and data to match, which can be done with a single `match` for both arguments.

Oppen: E.g.:
```suggestion
impl BuiltinAdditionalData {
    fn is_empty(&self) -> bool {
        use BuiltinAdditionalData::*;
        match self {
            Empty => true,
            Hash(data) => data.is_empty(),
            Signature(data) => data.is_empty(),
            Output(data) => data.is_empty(),
            None => true,
        }
    }
}

impl PartialEq for BuiltinAdditionalData {
    fn eq(&self, other: &BuiltinAdditionalData) -> bool {
        use BuiltinAdditionalData::*;
        if matches!(self, BuiltinAdditionalData::None) {
            return matches!(other, BuiltinAdditionalData::None);
        }
        match (self.is_empty(), other.is_empty()) {
            (true, true) => { return true; },
            (true, false)|(false, true) => { return false; },
            _ => (),
        }
        match (self, other) {
            (Hash(data), Hash(other_data)) if data == other_data => true,
            (Signature(data), Signature(other_data)) if data == other_data => true,
            (Output(data), Output(other_data)) if data == other_data => true,
            _ => false,
        }
    }
}
```

juanbono (approved):

vm/src/vm/runners/cairo_pie.rs
Alon-Ti: I need this variant to deserialize a JSON empty list, changing `usize` to the empty type.

vm/src/vm/runners/cairo_pie.rs
Alon-Ti: This doesn't distinguish an empty hash list from an empty signature list, but I agree that that's a good sacrifice for clarity.

Oppen: Huh, I misread the original code as (intentionally) doing that. Now that I read it again I see you use if let to distinguish.

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -85,6 +87,35 @@ pub enum BuiltinAdditionalData {
...
+            self.is_empty() & other.is_empty()
+        } else {
+            match (self, other) {
+                (Self::Hash(data), Self::Hash(other_data)) => data == other_data,
+                (Self::Signature(data), Self::Signature(other_data)) => data == other_data,
+                (Self::Output(data), Self::Output(other_data)) => data == other_data,
+                (Self::None, Self::None) => true,
+                _ => false,
+            }
+        }
```
Oppen, line 115: I think this should fix the error you found in my previous suggestion:
```suggestion
        match (self, other) {
            (Self::Hash(data), Self::Hash(other_data)) => data == other_data,
            (Self::Signature(data), Self::Signature(other_data)) => data == other_data,
            (Self::Output(data), Self::Output(other_data)) => data == other_data,
            (Self::None, Self::None) => true,
            (Self::Empty, x)|(x, Self::Empty) => x.is_empty(),
            _ => false,
        }
```
Emptiness for same variant instances is handled by the equality check, so we only need to ask that the other variant is empty if we know one is `Empty`. I'm not 100% sure it's legal as written due to the swapping order, worst case you need to write two arms instead.

Oppen (approved):

fmoletta (approved):

PR #1792, opened 2024-06-21
Fix WASM32 compilation for Cairo1-run and cairo-vm

wraitii:
# Fix WASM32 compilation for Cairo1-run and cairo-vm

## Description

Running the following fails on `main`:
```cargo build --bin cairo1-run --target wasm32-unknown-unknown --release --no-default-features```

There are 4 issues:
 - `zip` is not wasm-compatible by default. This can be circumvented with appropriate feature flags.
 - `cairo1-run/src/cairo_run.rs` sets an initial gas to 9999999999999_usize which is bigger than `usize` on 32 bit platforms.  I'm _not_ sure it's actually safe to downgrade this to merely `999999999`.
 - Two lines in `vm/src/lib.rs` and `vm/src/vm/runners/cairo_pie.rs` pessimistically assume `wasm <=> no_std`,  when the former works just fine - we can be more precise.

I also remove a comment to check WASM compatibility with some libraries, as they are indeed compatible.

This PR fixes both `cairo1-run` and `cairo-vm` for WASM targets.

### DIFF
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -141,6 +141,10 @@ pub fn cairo_run_program(
 
     let main_func = find_function(sierra_program, "::main")?;
 
+    #[cfg(target_pointer_width = "32")]
+    let initial_gas = 999999999_usize;
+
+    #[cfg(not(target_pointer_width = "32"))]
     let initial_gas = 9999999999999_usize;
 
     // Fetch return type data
--- a/vm/src/vm/runners/cairo_pie.rs
+++ b/vm/src/vm/runners/cairo_pie.rs
@@ -373,7 +373,7 @@ pub(super) mod serde_impl {
 
     use super::CAIRO_PIE_VERSION;
     use super::{CairoPieMemory, Pages, PublicMemoryPage, SegmentInfo};
-    #[cfg(any(target_arch = "wasm32", no_std, not(feature = "std")))]
+    #[cfg(any(no_std, not(feature = "std")))]
     use crate::alloc::string::ToString;
     use crate::stdlib::prelude::{String, Vec};
     use crate::{
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -37,7 +37,7 @@ test_utils = ["std", "dep:arbitrary", "starknet-types-core/arbitrary", "starknet
 extensive_hints = []
 
 [dependencies]
-zip = {version = "0.6.6", optional = true }
+zip = {version = "0.6.6", optional = true, default_features = false, features=["deflate"] }
 num-bigint = { workspace = true }
 rand = { workspace = true }
 num-traits = { workspace = true }
@@ -67,8 +67,6 @@ bitvec = { workspace = true }
 cairo-lang-starknet = { workspace = true, optional = true }
 cairo-lang-starknet-classes = { workspace = true, optional = true }
 cairo-lang-casm = { workspace = true, optional = true }
-
-# TODO: check these dependencies for wasm compatibility
 ark-ff = { workspace = true, optional = true }
 ark-std = { workspace = true, optional = true }
 
--- a/vm/src/lib.rs
+++ b/vm/src/lib.rs
@@ -12,7 +12,7 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![deny(warnings)]
 #![forbid(unsafe_code)]
-#![cfg_attr(any(target_arch = "wasm32", not(feature = "std")), no_std)]
+#![cfg_attr(any(not(feature = "std")), no_std)]
 
 #[cfg(feature = "std")]
 include!("./with_std.rs");

### CODE REVIEW

File: cairo1-run/src/cairo_run.rs [hunk 1]
```diff
@@ -141,6 +141,10 @@ pub fn cairo_run_program(
 
     let main_func = find_function(sierra_program, "::main")?;
 
+    #[cfg(target_pointer_width = "32")]
+    let initial_gas = 999999999_usize;
```
Oppen, line 145: If this value works, why not just use this?
Otherwise, why not change `initial_gas` to a `u64` instead of `usize`?

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -373,7 +373,7 @@ pub(super) mod serde_impl {
 
     use super::CAIRO_PIE_VERSION;
     use super::{CairoPieMemory, Pages, PublicMemoryPage, SegmentInfo};
-    #[cfg(any(target_arch = "wasm32", no_std, not(feature = "std")))]
+    #[cfg(any(no_std, not(feature = "std")))]
```
Oppen, line 376: I don't get this change.

Oppen:
> * Two lines in `vm/src/lib.rs` and `vm/src/vm/runners/cairo_pie.rs` pessimistically assume `wasm <=> no_std`,  when the former works just fine - we can be more precise.

Those really shouldn't be handled there in the first place. We have the `stdlib` module to abstract those differences away.

cairo1-run/src/cairo_run.rs
wraitii: > If this value works, why not just use this?
It compiles, but that doesn't tell me it's 'working' - I haven't really checked what it's being used for, and this is a billion gas - perhaps an achievable amount (where 1 trillion wasn't). If it's used to initialise a "Max" value, it's perhaps unsafe - would need someone to know why 999999999999 was chosen in the first place.

> Otherwise, why not change initial_gas to a u64 instead of usize?
Much smaller change / laziness, but I can update the PR to change this to u64 for sure

wraitii:
> > * Two lines in `vm/src/lib.rs` and `vm/src/vm/runners/cairo_pie.rs` pessimistically assume `wasm <=> no_std`,  when the former works just fine - we can be more precise.
> 
> Those really shouldn't be handled there in the first place. We have the `stdlib` module to abstract those differences away.

I'm not sure if this means I need to update the PR or if it's a general comment, could you clarify ?

---

Should I add this as a fix to the changelog? The PR template wasn't clear on this (this isn't a change that needs to be 'documented', per se).

cairo1-run/src/cairo_run.rs
Oppen: > It compiles, but that doesn't tell me it's 'working' - I haven't really checked what it's being used for, and this is a billion gas - perhaps an achievable amount (where 1 trillion wasn't). If it's used to initialise a "Max" value, it's perhaps unsafe - would need someone to know why 999999999999 was chosen in the first place.

IIUC that's just a "should be enough to not fail due to gas exhaustion" value, nothing else. I personally prefer to use the `MAX` constant for whatever type I'm using to make it more obvious that this is just a sloppy infinity value.

Oppen:
> I'm not sure if this means I need to update the PR or if it's a general comment, could you clarify ?

Just a general comment. Your fix for those is OK. We should clean up at some later point.

Oppen:
> Should I add this as a fix to the changelog? The PR template wasn't clear on this (this isn't a change that needs to be 'documented', per se).

Just the gist, something like "compilation of cairo-1-run and cairo-vm on wasm32 has been fixed" if the latest release was already broken. If only main broke and you fixed it before it reached a release, then it shouldn't affect users reading the changelog.

The intention of the template (we may need to rephrase it) is to document functional changes mostly. We consider non-functional changes thing like refactors, cleanups, or improvements to the build system or the CI, same for simply adding tests, while bug fixes for reported issues, significant performance changes or added/broken interfaces need to be in the changelog.

Oppen (approved):

Oppen:
Oh BTW, congrats! 🎉 
I just noticed it's your first PR. It's on the right track for merging.

Oppen:
Also, I played a bit of 0AD like 10 years ago or so. I didn't do much because IIRC there were no campaigns I had nobody to play with. Nice to cross paths with a contributor.

wraitii:
> Just the gist, [...]

Added a line, can't hurt much anyways.


> Also, I played a bit of 0AD like 10 years ago or so. I didn't do much because IIRC there were no campaigns I had nobody to play with. Nice to cross paths with a contributor.

Ah that's fun 😄 we have support for campaigns nowadays, but IIRC still none of them. Though a small multiplayer user base exists.

wraitii:
Hey @Oppen , bumping this - from my perspective this is still ready to merge

Oppen:
> Hey @Oppen , bumping this - from my perspective this is still ready to merge

I agree. We need a second reviewer to approve it. I'll remind the team we have this PR waiting.

gabrielbosio:
superseded by #2315

PR #1795, opened 2024-06-25
Fix: make coverage command

FrancoGiachetta:
# Change coverage command

## Description

This PR changes make coverage command so that it executes the following command: 

`cargo llvm-cov --html --workspace --features "test_utils, cairo-1-hints"`

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,7 +275,7 @@ clippy:
 	cargo clippy --manifest-path fuzzer/Cargo.toml --all-targets
 
 coverage:
-	cargo llvm-cov report --lcov --output-path lcov.info
+	cargo llvm-cov --html --workspace --features "test_utils, cairo-1-hints"
 
 coverage-clean:
 	cargo llvm-cov clean
--- a/Makefile
+++ b/Makefile
@@ -274,7 +274,7 @@ clippy:
 	cargo clippy --workspace --all-features --benches --examples --tests -- -D warnings
 	cargo clippy --manifest-path fuzzer/Cargo.toml --all-targets
 
-coverage:
+coverage: cairo_proof_programs cairo_test_programs cairo_1_test_contracts cairo_2_test_contracts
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -275,7 +275,7 @@ clippy:
 	cargo clippy --manifest-path fuzzer/Cargo.toml --all-targets
 
 coverage:
-	cargo llvm-cov report --lcov --output-path lcov.info
+	cargo llvm-cov --html --workspace --features "test_utils, cairo-1-hints"
```
pefontana, line 278: This line is OK, but let me clarify a few things:

The idea is that a user should be able to run the commands:
- `make deps`
- `make coverage`

These commands should execute successfully. However, at present, this is not the case because running the `make coverage` command requires the compilation of several programs.

Check the `make test` command:
```makefile
test: cairo_proof_programs cairo_test_programs cairo_1_test_contracts cairo_2_test_contracts
```

The names listed after the colon are targets that the make command runs before executing the test command.

Let me know if you need any more help!

pefontana (changes requested):

FrancoGiachetta:
Changes were done @pefontana

pefontana (approved):
nice one!

pefontana:
@FrancoGiachetta Put the "Tests" label to the PR so the changelog check doesn't fail

Oppen:
I'm not sure this was broken in the first place, though I failed to document properly how it works. `make test` produces coverage data, what `make coverage` currently does is consolidate it into a report we can upload to codecov.io. That's so we can run the different features in parallel in the CI. The original way took us a long time for each PR.
Is this no longer working?

FrancoGiachetta:
Hi! The idea behind this change is so that we can run `make coverage` without the need to run `make test` before that.

edg-l (approved):

PR #1796, opened 2024-06-26
Fix Zero segment location.

Alon-Ti:
The zero segment location has been changed to be after the rest of the segments.

### DIFF
--- a/vm/src/cairo_run.rs
+++ b/vm/src/cairo_run.rs
@@ -173,7 +173,12 @@ pub fn cairo_run_pie(
         }
     }
     // Load previous execution memory
-    let n_extra_segments = pie.metadata.extra_segments.len();
+    let has_zero_segment = if cairo_runner.vm.segments.has_zero_segment() {
+        0
+    } else {
+        1
+    };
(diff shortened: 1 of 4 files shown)


Oppen:
I don't understand, what was wrong with the previous location?
Also, why is it initialized separate from all other builtin runner segments and after the entrypoint?
Lastly, whatever those reasons are, could they get documented as code comments where it matters? Otherwise, we may inadvertently break it in the future. Tests for the desired behavior would also be good.

vm/src/cairo_run.rs
```diff
@@ -173,7 +173,12 @@ pub fn cairo_run_pie(
         }
     }
     // Load previous execution memory
-    let n_extra_segments = pie.metadata.extra_segments.len();
+    let has_zero_segment = if cairo_runner.vm.segments.has_zero_segment() {
+        0
+    } else {
+        1
+    };
```
Oppen, line 180:
```suggestion
    let has_zero_segment = cairo_runner.vm.segments.has_zero_segment() as usize;
```

PR #1815, opened 2024-08-05
Only initialize program builtins

### LINKED ISSUE #1814: Add CairoRunner:: initialize_program_builtins method
Add a `initialize_program_builtins` method to the CairoRunner struct.
The goal of this method, is to initialize just the `&self.program.builtins`
This function should be the same as the `initialize_all_builtins` but with some modifications:

Remove these chuncks of code:

```
        let starknet_preset_builtins = vec![
            BuiltinName::pedersen,
            BuiltinName::range_check,
            BuiltinName::output,
            BuiltinName::ecdsa,
            BuiltinName::bitwise,
            BuiltinName::ec_op,
            BuiltinName::keccak,
            BuiltinName::poseidon,
        ];
```

```
        for builtin_name in starknet_preset_builtins {
            if !self.program.builtins.contains(&builtin_name) {
                initialize_builtin(builtin_name, &mut self.vm, add_segment_arena_builtin)
            }
        }
```

note: We can avoid the segment arena flag, because it is present in `self.program.builtins`

* Remove the old `initialize_all_builtins` method, is not necessary

JulianGCalderon:
# Only initialize program builtins

closes #1814

## Description

Replace initialize_all_builtins for initialize_program_builtins

It also removes/changes tests that used the former function to make sense with the new function

## Checklist
- [x] Linked to Github Issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Upcoming Changes
 
-* fix(BREAKING): Only initialize program builtins instead of a preset list
+* fix(BREAKING): Replace `CairoRunner` method `initialize_all_builtins` with `initialize_program_builtins`. Now it only initializes program builtins instead of all of them
 
 #### [1.0.0] - 2024-08-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* fix(BREAKING): Only initialize program builtins instead of a preset list
+
 #### [1.0.0] - 2024-08-01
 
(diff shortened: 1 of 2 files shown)


pefontana (approved):
Nice one @JulianGCalderon !

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* fix(BREAKING): Only initialize program builtins instead of a preset list
```
fmoletta, line 5: As this is a breaking change I think we should mention which part of the public api was changed. I would add something like: Replace `CairoRunner` method `initialize_all_builtins` with `initialize_program_builtins`.

JulianGCalderon: Done!

fmoletta (approved):
🚀

PR #1832, opened 2024-09-03
Disable unnecessary features of `zip` dependency

kkovaacs:
We've discovered that using the `cairo-vm` crate in our projects brings in a lot of unnecessary dependencies and in the end leads to our binary being linked with the `libbz2` library.

Turns out that neither of these is actually required and this happens because the `zip` crate has a large number of features enabled by default.

This change disables all `zip` features except `deflate` which the code actually uses.


Oppen (approved):
Thanks! Excellent catch! 🎉

pefontana (approved):
Nice one @kkovaacs !

PR #1833, opened 2024-09-03
Cairo-lang2.8.0

FrancoGiachetta:
# Update cairo-lang to 2.8.0

## Description

This PR updates cairo-lang creates to 2.8.0. It also updates rust to version 1.80.0 since now cairo-lang uses the `lazy-cell` library, which was stabilized at that version. It also removes some features used before `alloc` and `std`, which seem to be deprecated in rust 1.80.0

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/air_private_input.rs
+++ b/vm/src/air_private_input.rs
@@ -212,7 +212,7 @@ mod tests {
         assert_matches::assert_matches,
     };
 
-    #[cfg(any(target_arch = "wasm32", no_std, not(feature = "std")))]
+    #[cfg(any(target_arch = "wasm32", not(feature = "std")))]
     use crate::alloc::string::ToString;
 
     #[cfg(feature = "std")]
--- a/vm/src/tests/cairo_pie_test.rs
+++ b/vm/src/tests/cairo_pie_test.rs
@@ -6,7 +6,7 @@ use crate::{
 #[cfg(target_arch = "wasm32")]
 use wasm_bindgen_test::*;
 
-#[cfg(all(not(feature = "std"), feature = "alloc"))]
+#[cfg(all(not(feature = "std")))]
 use alloc::{
     string::{String, ToString},
     vec::Vec,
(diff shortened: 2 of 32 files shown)


JulianGCalderon (approved):

vm/src/air_private_input.rs
```diff
@@ -215,7 +215,7 @@ mod tests {
         assert_matches::assert_matches,
     };
 
-    #[cfg(any(target_arch = "wasm32", no_std, not(feature = "std")))]
```
pefontana, line 218: why do we change this?

vm/src/tests/cairo_pie_test.rs
```diff
@@ -6,12 +6,6 @@ use crate::{
 #[cfg(target_arch = "wasm32")]
 use wasm_bindgen_test::*;
 
-#[cfg(all(not(feature = "std"), feature = "alloc"))]
```
pefontana, line 9: same here, these imports where unused?

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -378,7 +378,7 @@ pub(super) mod serde_impl {
 
     use super::CAIRO_PIE_VERSION;
     use super::{CairoPieMemory, Pages, PublicMemoryPage, SegmentInfo};
-    #[cfg(any(target_arch = "wasm32", no_std, not(feature = "std")))]
+    #[cfg(any(target_arch = "wasm32", not(feature = "std")))]
```
pefontana, line 381: same here

pefontana (changes requested):

vm/src/air_private_input.rs
FrancoGiachetta: It seems it is not a valid attribute for `#[cfg(...)]` in rust 1.80.0

vm/src/tests/cairo_pie_test.rs
FrancoGiachetta: Correct. The weren't being used and they also were duplicated.

pefontana (changes requested):
Nice @FrancoGiachetta !
Can you update the readme Doc
`- [Rust 1.76.0 or newer](https://www.rust-lang.org/tools/install)`
I think that with that we could merge it

FrancoGiachetta:
Done!

pefontana (approved):

PR #1845, opened 2024-09-25
Always use a normal segment for first SegmentArena segment

notlesh:
(Note that this is basically a clone of https://github.com/Moonsong-Labs/cairo-vm/pull/45. We'd love to get this accepted upstream if possible.)

Problem: Starknet OS expects SegmentArena segments to all be temporary except the first segment, which must be normal. This is so that relocation rules for all segments can chain back into the first.

Solution: Use temporary segments only if it's not the first.

## Description

The Starknet OS verifies the continuity of the segments in `SegmentArena`s [here](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/starknet/builtins/segment_arena/segment_arena.cairo#L90), which requires relocation rules to be properly set up.

The requirements for this to work are that each `src_ptr` is a temporary segment and each `dest_ptr` is a non-temporary segment (or a temporary segment with an existing rule mapping it to a non-temporary segment).

`cairo-vm` uses a number of hints to accomplish this, and these are injected by the sierra-to-casm compiler. However, it relies on [one additional hint](https://github.com/Moonsong-Labs/cairo-vm/blob/19b07b4c0c59265a2d79b6b0bead9c7490ede6e9/vm/src/hint_processor/cairo_1_hint_processor/hint_processor.rs#L59) which is not injected anywhere in the case of SNOS.

This hint seems to ensure the same condition needed in SNOS: that the initial segment is a normal segment and that all other segments are temporary segments with relocation rules pointing back to the initial one. However, it creates the normal segment inside this last hint rather than when the SegmentArena is initialized.

SNOS handles this by simply allocating the initial segment as a normal one [as explained here](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/starknet/builtins/segment_arena/segment_arena.cairo#L12). This seems to alleviate the need to later recreate the segment in `relocate_all_segments()`, allowing the implementation to work in SNOS without the hint mentioned above.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


pefontana:
Hi @notlesh 
Thanks for the contribution!
There is just a minor error in the CI in the `make check-fmt` check
Also, it would be possible to add a little test for this case?

notlesh:
> Hi @notlesh Thanks for the contribution! There is just a minor error in the CI in the `make check-fmt` check Also, would it be possible to add a small program or a little test for this fix?

I've wanted to produce a better way to test this, but it's complicated because the failure occurs in a different repo.

One idea for a test case is to show that the dictionaries do not have a temporary segment as their first segment prior to the `RelocateAllDictionaries` cheatcode being called -- this is fundamentally what was broken and what is fixed here.

This should act as a regression test (reproducing the problem and demonstrating that it's fixed) but it doesn't reproduce the exact error. Let me know what you think and if you want more details or want to know how you can reproduce the exact error yourself.

pefontana:
> This should act as a regression test (reproducing the problem and demonstrating that it's fixed) but it doesn't reproduce the exact error. Let me know what you think and if you want more details or want to know how you can reproduce the exact error yourself.

What do you mean by it `doesn't reproduce the exact error`? Can we add some Cairo program that plays with dictionaries triggering the error? O maybe just add some unit tests


Also, there are merge conflicts that need to let me approve the CI run, can you solve them?

notlesh:
> What do you mean by it `doesn't reproduce the exact error`? Can we add some Cairo program that plays with dictionaries triggering the error? O maybe just add some unit tests
> 
> Also, there are merge conflicts that need to let me approve the CI run, can you solve them?

Sorry for the long delay, I'd like to push this forward soon (resolve conflicts and add a test). In the meantime, I was wondering if you could point me to an example of the sort of test you have in mind, as I'm unsure of where to start with this.

gabrielbosio:
> Let me know what you think and if you want more details or want to know how you can reproduce the exact error yourself.

Hi, @notlesh! To test this, I think it would be helpful to know what's the code that's being problematic to see why is this scenario difficult to reproduce, because I'm not sure if there's a Cairo program that can be used for testing because the hint that calls `new_default_dict` already enforces the temporary segments to be all of the segments except the first one.
Once we have the code we can think of something smaller like a unit test.

notlesh:
> > Let me know what you think and if you want more details or want to know how you can reproduce the exact error yourself.
> 
> Hi, @notlesh! To test this, I think it would be helpful to know what's the code that's being problematic to see why is this scenario difficult to reproduce, because I'm not sure if there's a Cairo program that can be used for testing because the hint that calls `new_default_dict` already enforces the temporary segments to be all of the segments except the first one. Once we have the code we can think of something smaller like a unit test.

Hey, @gabrielbosio , thanks for taking a look.

The code that is failing is the [first link mentioned in the description](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/starknet/builtins/segment_arena/segment_arena.cairo#L90), specifically the hint that calls `memory.add_relocation_rule()`:

```cairo
// Helper function for validate_segment_arena.
func _verify_continuity(infos: SegmentInfo*, n_segments_minus_one: felt) {
    if (n_segments_minus_one == 0) {
        // If there is only one segment left, there is no need to check anything.
        return ();
    }

    // Enforce an empty cell between two consecutive segments so that the start of a segment
    // is strictly bigger than the end of the previous segment.
    // This is required for proving the soundness of this construction, in the case where a segment
    // has length zero.

    // Note: the following code was copied from relocate_segment() for efficiency reasons.
    let src_ptr = infos[1].start;
    let dest_ptr = infos[0].end + 1;
    %{ memory.add_relocation_rule(src_ptr=ids.src_ptr, dest_ptr=ids.dest_ptr) %} // <--- FAILS HERE
    assert src_ptr = dest_ptr;

    return _verify_continuity(infos=&infos[1], n_segments_minus_one=n_segments_minus_one - 1);
}
```

This fails because `src_ptr` must be a temporary segment. `cairo-vm` currently does this, basically:
* Allocate everything as a temporary segment during execution
* At the end, in a hint injected by the Sierra-CASM compiler (IIRC), [this is fixed by creating a single normal segment](https://github.com/Moonsong-Labs/cairo-vm/blob/19b07b4c0c59265a2d79b6b0bead9c7490ede6e9/vm/src/hint_processor/cairo_1_hint_processor/dict_manager.rs#L124) before processing all of the other segments

The current code works so long as this hint is called, but that's not possible in Cairo0 code -- which the OS is written in, hence the reason it doesn't work for SNOS.

This PR fixes this by ensuring that the required normal segment is allocated up front instead of fixed later on, which I think works for all cases.

Let me know if this is any clearer than it was, I know the problem is pretty detailed. I'd also be happy to jump on a call and walk through it with you. Thanks again!

gabrielbosio:
> This PR fixes this by ensuring that the required normal segment is allocated up front instead of fixed later on, which I think works for all cases.

To make sure that this works for all cases, I'm thinking of adding at least two tests:

- Check that this works for Cairo 0. This could be achieved by calling the `validate_segment_arena` without failing.
- Check that this works for Cairo 1. This could be achieved with a Cairo program that uses the `alloc_felt_252_dict` hint. For example, this program uses it:
  
  ```
  use dict::{felt252_dict_entry_finalize, Felt252DictTrait};

  fn main() -> (felt252, felt252) {
      let mut dict = felt252_dict_new::<felt252>();
      dict.insert(1, 64);
      let val = dict.get(1);
      dict.insert(1, 75);
      let val2 = dict.get(1);
      dict.squash();
      return (val, val2);
  }
  ```
  
  And then run this program and check the segments the hint creates are the same than 
  `main`. I've run this program in both `main` and this branch and they produced the same segments. The segments created by the hint were the ones that had index >= 6 so I think you can compare those instead of the entire memory.

gabrielbosio:
Also, there's an error from clippy

gabrielbosio:
@notlesh any news on this?

whichqua:
Hey @gabrielbosio, I am taking on the remaining parts of this task, specifically ensuring our fix is tested appropriately.

### Initial takeaways
1. For Cairo 1, there are already existing tests that use the `alloc_felt_252_dict` as seen [here](https://github.com/lambdaclass/cairo-vm/blob/7dde1c6a6b88693fdd8b99ee8d8bbaa6f2800de9/cairo_programs/cairo-1-contracts/felt252_dict_entry_update.cairo#L7) and [here](https://github.com/lambdaclass/cairo-vm/blob/7dde1c6a6b88693fdd8b99ee8d8bbaa6f2800de9/cairo_programs/cairo-1-contracts/assert_le_find_small_arcs.cairo#L7). The tests are still passing and there is not much we are gonna add on this.
2. For Cairo 0, we are going to add a test and this is not straight forward but I do have an attempt.

#### Our Cairo0 approach:
- We have added a Cairo0 test that creates some segments and runs some transformations on these segments as seen on [this file](https://github.com/lambdaclass/cairo-vm/blob/bc9ea69aae39793b5bcd0dbd3a7f3bd1ce3b69bc/cairo_programs/test_segment_arena.cairo) which is based on [this test](https://github.com/starkware-libs/cairo-lang/blob/8276ac35830148a397e1143389f23253c8b80e93/src/starkware/starknet/builtins/segment_arena/segment_arena_test.cairo#L16) on the `cairo-lang` repo.
- This test has some hints that are not available on `cairo-vm`, and hence my test is a little more customized as seen [here](https://github.com/lambdaclass/cairo-vm/blob/bc9ea69aae39793b5bcd0dbd3a7f3bd1ce3b69bc/vm/src/tests/cairo_run_test.rs#L1464).
I would like to hear your opinion on this, specifically if this approach is what your team expects. 

Let me know your thoughts.

cairo_programs/test_segment_arena.cairo
```diff
@@ -0,0 +1,170 @@
+from starkware.cairo.common.alloc import alloc
```
gabrielbosio, line 1: If this test file is based on [this one](https://github.com/starkware-libs/cairo-lang/blob/8276ac35830148a397e1143389f23253c8b80e93/src/starkware/starknet/builtins/segment_arena/segment_arena_test.cairo) I would add a comment with a link to that file.

cairo_programs/test_segment_arena.cairo
```diff
@@ -0,0 +1,170 @@
...
+        assert finalize_segment(segment3 + 2) = segment3;
+        assert finalize_segment(segment2 + 1) = segment2;
+    }
+    validate_segment_arena(segment_arena=&segment_arena[-1]);
+    return (segment0, segment_arena_start[-1].infos);
+}
+
+func main{}() {
+    let (res, segments) = test_segment_arena();
+    // assert res[0] = 1;
```
gabrielbosio, line 168: Why is this commented?

cairo_programs/test_segment_arena.cairo
```diff
@@ -0,0 +1,170 @@
...
+    ret;
+}
```
gabrielbosio, line 170: Please add a final newline

vm/src/tests/cairo_run_test.rs
```diff
@@ -1351,3 +1360,123 @@ fn cairo_run_data_availability_reduced_mul() {
...
+
+    let mut hint_executor = BuiltinHintProcessor::new(extra_hints, run_resources);
+    let cairo_run_config = CairoRunConfig {
+        layout: LayoutName::all_cairo,
+        relocate_mem: true,
+        trace_enabled: true,
+        proof_mode: false,
+        ..Default::default()
+    };
+    cairo_run(program_data, &cairo_run_config, &mut hint_executor).expect("Execution failed");
```
gabrielbosio, line 1481: Given that [the Python test used as reference](https://github.com/starkware-libs/cairo-lang/blob/8276ac35830148a397e1143389f23253c8b80e93/src/starkware/starknet/builtins/segment_arena/segment_arena_test.py#L23-L44) runs some assertions after running the program, I think it would be nice if this test includes those assertions too.

vm/src/tests/cairo_run_test.rs
```diff
@@ -1351,3 +1360,123 @@ fn cairo_run_data_availability_reduced_mul() {
...
+    use crate::any_box;
+    use crate::hint_processor::builtin_hint_processor::hint_utils::insert_value_into_ap;
+    use crate::hint_processor::hint_processor_definition::HintReference;
+    use crate::serde::deserialize_program::ApTracking;
+    use crate::types::exec_scope::ExecutionScopes;
+    use crate::vm::errors::hint_errors::HintError;
+    use crate::vm::vm_core::VirtualMachine;
+    use crate::Felt252;
+    use indoc::indoc;
+    use std::collections::HashMap;
```
gabrielbosio, line 1376: Please move these imports outside the test function

vm/src/tests/cairo_run_test.rs
```diff
@@ -1,4 +1,13 @@
-use crate::{tests::*, types::layout_name::LayoutName};
+use std::rc::Rc;
```
FrancoGiachetta, line 1: Here there, I know this is still in process. But this line could lead to a criptic error when running wasm tests. Instead of using `std`, try with `crate::stdlib::rc::Rc`.

vm/src/tests/cairo_run_test.rs
```diff
@@ -1351,3 +1360,123 @@ fn cairo_run_data_availability_reduced_mul() {
...
+        segment_index_to_arena_index.insert(start.segment_index, n_segments);
+
+        exec_scopes.data[0].insert(
+            "segment_index_to_arena_index".to_string(),
+            any_box!(segment_index_to_arena_index),
+        );
+
+        Ok(())
+    }
+
```
gabrielbosio, line 1463: I suggest to move these hints to a helper file in the same directory than this file.

gabrielbosio (changes requested):
Hi @whichqua I think the proposal makes sense, broadly speaking. I left some suggestions and comments regarding specific details.

cairo_programs/test_segment_arena.cairo
whichqua: I need to do the actual assertions here.

vm/src/tests/cairo_run_test.rs
whichqua: Sure, I will have that.

vm/src/tests/cairo_run_test.rs
whichqua: I ended up moving the test into its own file as its pretty different with everything else in that file.
See https://github.com/lambdaclass/cairo-vm/pull/1845/commits/3c2c6bc23ee7270a986dd465f9a287b9ae2ee7ca

vm/src/tests/cairo_run_test.rs
whichqua: Resolved in https://github.com/lambdaclass/cairo-vm/pull/1845/commits/3c2c6bc23ee7270a986dd465f9a287b9ae2e

cairo_programs/test_segment_arena.cairo
whichqua: Resolved in https://github.com/lambdaclass/cairo-vm/pull/1845/commits/3c2c6bc23ee7270a986dd465f9a287b9ae2e

cairo_programs/test_segment_arena.cairo
whichqua: Resolved in https://github.com/lambdaclass/cairo-vm/pull/1845/commits/3c2c6bc23ee7270a986dd465f9a287b9ae2e

vm/src/tests/cairo_run_test.rs
whichqua: Resolve in https://github.com/lambdaclass/cairo-vm/pull/1845/commits/e4ef596e1b9f11f854dc3f8248a4cd35d66a2735

vm/src/hint_processor/builtin_hint_processor/segments.rs
```diff
@@ -19,11 +19,35 @@ pub fn relocate_segment(
...
-    #[cfg(feature = "extensive_hints")]
-    let dest_ptr = crate::hint_processor::builtin_hint_processor::hint_utils::get_maybe_relocatable_from_var_name("dest_ptr", vm, ids_data, ap_tracking)?;
-
+    // Bugfix: this is a workaround for an issue in the vm related to the way it computes
+    // `dest_ptr` in `segment_arena.cairo`. For some reason it sets `dest_ptr` to 0 (felt) instead
+    // of a valid relocatable value.
+    // As this hint is used in other places, we need to determine if we are in `segments_arena`
+    // first. We check this by determining if the "infos" variable is declared.
+    // If it is the case, we simply recompute `dest_ptr` manually.
+    let dest_ptr = if let Ok(infos) = get_ptr_from_var_name("infos", vm, ids_data, ap_tracking) {
```
whichqua, line 28: Maybe this can be improved with something like:
```rs
   fn is_segment_arena(ids_data: &HashMap<String, HintReference>) -> bool {
        return ids_data.contains_key("infos")
            && ids_data
                .get("infos")
                .expect("Must contain infos")
                .cairo_type
                .expect("Must have a type")
                .ends_with("SegmentInfo*");
    }
```

whichqua:
@gabrielbosio This is ready for review.

vm/src/hint_processor/builtin_hint_processor/segments.rs
FrancoGiachetta: I would avoid trying to match strings when checking for types. There could be future updates which changes `SegmentInfo*` with a different name. This would end up not only in a silent error, but a very difficult one to spot.

vm/src/hint_processor/builtin_hint_processor/segments.rs
```diff
@@ -19,11 +19,35 @@ pub fn relocate_segment(
...
+
+        #[cfg(not(feature = "extensive_hints"))]
+        {
+            (infos_0_end + 1u32)?
+        }
+
+        #[cfg(feature = "extensive_hints")]
+        {
+            crate::types::relocatable::MaybeRelocatable::RelocatableValue((infos_0_end + 1u32)?)
+        }
```
FrancoGiachetta, line 39: In this cases, you can use this syntax:
```rust
if cfg(not(feature = "extensive_hints")) {
...
} else {
...
}
```
Makes it more readable.

vm/src/hint_processor/builtin_hint_processor/segments.rs
```diff
@@ -19,11 +19,35 @@ pub fn relocate_segment(
...
+    } else {
+        #[cfg(not(feature = "extensive_hints"))]
+        {
+            get_ptr_from_var_name("dest_ptr", vm, ids_data, ap_tracking)?
+        }
+
+        #[cfg(feature = "extensive_hints")]
+        {
+            crate::hint_processor::builtin_hint_processor::hint_utils::get_maybe_relocatable_from_var_name("dest_ptr", vm, ids_data, ap_tracking)?
```
FrancoGiachetta, line 48: Here also, you can use the syntax I stated in the previous comment.

FrancoGiachetta (changes requested):

vm/src/hint_processor/cairo_1_hint_processor/dict_manager.rs
```diff
@@ -123,9 +123,26 @@ impl DictManagerExecScope {
     /// Relocates all dictionaries into a single segment
     /// Does nothing if use_temporary_segments is set to false
     pub fn relocate_all_dictionaries(&mut self, vm: &mut VirtualMachine) -> Result<(), HintError> {
-        if self.use_temporary_segments {
-            let mut prev_end = vm.add_memory_segment();
-            for tracker in &self.trackers {
+        // We expect the first segment to be a normal one, which doesn't require relocation. So
+        // there is nothing to do unless there are at least two segments.
+        if self.use_temporary_segments && !self.trackers.is_empty() {
+            let first_segment = self.trackers.first().expect("Trackers cannot be empty");
```
FrancoGiachetta, line 129: We should be returning an error instead of panicking in these cases which are not tests. This function already returns one, so you can use `?`.

FrancoGiachetta (changes requested):

vm/src/hint_processor/cairo_1_hint_processor/dict_manager.rs
whichqua: Hmmm, in this case we already have a guard:
```
!self.trackers.is_empty()
```
We know that this will never panic and if it did it would be a rust bug. On the other hand using ? will force me to define a new error variant in `HintError`, let me know if this is a better approach.

vm/src/hint_processor/builtin_hint_processor/segments.rs
whichqua: Noted!

vm/src/hint_processor/cairo_1_hint_processor/dict_manager.rs
FrancoGiachetta: You can return a: `Err(HintError::CustomHint(<string>))`

whichqua:
@gabrielbosio @FrancoGiachetta Would be nice if we had the workflows running.

whichqua:
@gabrielbosio I need your thoughts how to approach [this failure](https://github.com/lambdaclass/cairo-vm/actions/runs/13994255026/job/39198994454?pr=1845#step:6:1). The hints are custom and are not implemented on the vm.

FrancoGiachetta:
There's a way to add custom hints and add them during the vm initialization (which might suit your case since it's a test). Check this [link](https://github.com/lambdaclass/cairo-vm/tree/main/examples/custom_hint). It's an example of how to use it.

whichqua:
@FrancoGiachetta That is not the problem here. Our test contains a hint that is not in the vm

FrancoGiachetta:
Yes, I understand that. Custom hints are an easy way of defining hints that are no implemented in the vm. I think that should be the solution to the issue here.

whichqua:
@FrancoGiachetta The hints themselves are already implemented [here](https://github.com/lambdaclass/cairo-vm/pull/1845/files#diff-7499c392237ea6422c3314c36699efb87957444f8b4a6a28a968ac508cf014d3R44)

FrancoGiachetta:
Mmm, I see. Perhaps it's not well formated?

whichqua:
I dont think so, I can successfully run `make test` without any problem. The failures come from [this workflow](https://github.com/lambdaclass/cairo-vm/blob/3aed53229422fb1d74cdd15b027296bd620f458a/.github/workflows/rust.yml#L527).

FrancoGiachetta:
Okay. The issue seems to be that since your program is in cairo_programs/ it's being taken into account by that workflow. Since it uses the cairo-vm-cli to run it, that custom hints are not defined. I can only suggest you to move that code to a different location to see if that solves the problem. Create a new folder inside cairo-programs/ an place you program there. Let me know if that helps.

Makefile
```diff
@@ -218,7 +222,7 @@ cargo-deps:
 	cargo install --version 1.19.0 hyperfine
 	cargo install --version 0.9.49 cargo-nextest --locked
 	cargo install --version 0.5.9 cargo-llvm-cov
-	cargo install --version 0.12.1 wasm-pack --locked
+	# cargo install --version 0.12.1 wasm-pack --locked
```
FrancoGiachetta, line 225: why is this comment?

whichqua, line 225: Hmm, missed removing that comment.
```suggestion
	cargo install --version 0.12.1 wasm-pack --locked
```

gabrielbosio:
@whichqua looks good, I'm waiting for the CI in a clone of this PR (#2029) to make sure the jobs that fail here pass in the clone.

gabrielbosio:
@whichqua I've run the CI and the codecov job fails. It seems that you need to cover the cases where `relocate_all_dictionaries` returns `Err`.
[Here's](https://app.codecov.io/gh/lambdaclass/cairo-vm/pull/2029?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=lambdaclass) the link to the codecov analysis.

gabrielbosio:
Also, all the commits in this PR must be signed before merging.

whichqua:
> It seems that you need to cover the cases where relocate_all_dictionaries returns Err.

I think I can add unit tests on the called methods. Would that be enough?

gabrielbosio:
I think it would 👍

whichqua:
@gabrielbosio I had to work though the resigning of some past commits but ended up settling for a force push. Do you need @notlesh to also sign in this case?

whichqua:
Any updates on this @gabrielbosio?

gabrielbosio:
> Do you need @notlesh to also sign in this case?

It seems it's enough with what you did. I'm running the CI jobs in #2048 to check if coverage is OK.

whichqua:
@gabrielbosio Seems everything is ✅

gabrielbosio (approved):

FrancoGiachetta (approved):

PR #1851, opened 2024-10-14
Sort_ecdsa_and_mod_builtins_private_inputs_by_idx

YairVaknin-starkware:
# TITLE
Sort_ecdsa_and_mod_builtins_private_inputs_by_idx

## Description
Making sure to sort the resulted list of those builtins in the air_private_input_file by index, so it is aligned with the python vm's outputs.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 #### Upcoming Changes
 
+* fix: [#1851](https://github.com/lambdaclass/cairo-vm/pull/1851):
+  * Fix unsorted signature and mod builtin outputs in air_private_input.
+
 * feat(BREAKING): [#1824](https://github.com/lambdaclass/cairo-vm/pull/1824)[#1838](https://github.com/lambdaclass/cairo-vm/pull/1838):
     * Add support for dynamic layout
     * CLI change(BREAKING): The flag `cairo_layout_params_file` must be specified when using dynamic layout.
--- a/vm/src/vm/runners/builtin_runner/signature.rs
+++ b/vm/src/vm/runners/builtin_runner/signature.rs
@@ -210,7 +210,17 @@ impl SignatureBuiltinRunner {
 
     pub fn air_private_input(&self, memory: &Memory) -> Vec<PrivateInput> {
         let mut private_inputs = vec![];
-        for (addr, signature) in self.signatures.borrow().iter() {
+
+        // Collect and sort the signatures by their index before the loop
+        let binding = self.signatures.borrow();
+        let mut sorted_signatures: Vec<_> = binding.iter().collect();
+        sorted_signatures.sort_by_key(|(addr, _)| {
+            addr.offset
+                .checked_div(CELLS_PER_SIGNATURE as usize)
+                .unwrap_or_default()
+        });
+
+        for (addr, signature) in sorted_signatures {
             if let (Ok(pubkey), Some(msg)) = (
                 memory.get_integer(*addr),
                 (*addr + 1_usize)
@@ -238,14 +248,6 @@ impl SignatureBuiltinRunner {
                 }))
             }
         }
-        private_inputs.sort_by_key(|input| match input {
-            PrivateInput::Signature(sig) => sig.index,
-            _ => unreachable!(
-                "Unexpected variant in private_inputs: {:?}.
-            This private input of the signature builtin is always a Signature variant.",
-                input
-            ),
-        });
         private_inputs
     }
 }
--- a/vm/src/vm/runners/builtin_runner/signature.rs
+++ b/vm/src/vm/runners/builtin_runner/signature.rs
@@ -240,7 +240,11 @@ impl SignatureBuiltinRunner {
         }
         private_inputs.sort_by_key(|input| match input {
             PrivateInput::Signature(sig) => sig.index,
-            _ => unreachable!(),
+            _ => unreachable!(
+                "Unexpected variant in private_inputs: {:?}.
+            This private input of the signature builtin is always a Signature variant.",
+                input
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
pefontana, line 1: @YairVaknin-starkware Can you restore this format changes?

YairVaknin-starkware: Done.

pefontana (commented):
Great @YairVaknin-starkware !
Is there a possibility to add a test? I don't know what these changes are for
If we could add an integration test, if not a unit test, it would be good

YairVaknin-starkware:
> Great @YairVaknin-starkware ! Is there a possibility to add a test? I don't know what these changes are for If we could add an integration test, if not a unit test, it would be good
@pefontana It's just sorting the array in the resulting air_private_input json according to indices, so comparing  air_private_inputs using [https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/tests/air_private_input_comparator.py](url) for xample doesn't fail unnecessarily if either of those two builtins are included (when comparing to a python cairo run). 
Does it still sound to you like it's necessary to add test?

vm/src/vm/runners/builtin_runner/signature.rs
```diff
@@ -238,6 +238,10 @@ impl SignatureBuiltinRunner {
                 }))
             }
         }
+        private_inputs.sort_by_key(|input| match input {
+            PrivateInput::Signature(sig) => sig.index,
+            _ => unreachable!(),
```
pefontana, line 243: How much work is to add an error handling here?
If it is not possible at least add an error message to the user

pefontana:
> > Great @YairVaknin-starkware ! Is there a possibility to add a test? I don't know what these changes are for If we could add an integration test, if not a unit test, it would be good
> 
> @pefontana It's just sorting the array in the resulting air_private_input json output file according to indices, so comparing air_private_inputs using [https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/tests/air_private_input_comparator.py](url) for example doesn't fail unnecessarily if either of those two builtins are included (when comparing to a python cairo run). Does it still sound to you like it's necessary to add test?

Okey! no problem, it is okey to don't add test in this case
It would be great if we can remove that unreachable and add some error handling

vm/src/vm/runners/builtin_runner/signature.rs
YairVaknin-starkware: Added a msg here and also add a test for the signature's `air_public_input` fn.

YairVaknin-starkware: > How much work is to add an error handling here?
> 
> If it is not possible at least add an error message to the user and a line documenting why it is not possible to get into the unreachable

I can add error handling, but I'm not sure how elegant it will be. Lmk if it's important, or if it's okay the way it is now.

YairVaknin-starkware: Also, currently `air_private_input` doesn't return a Result type for any builtin variant. Do we want to `panic` or refactor the code that uses it? Because this case can never happen, I think if we do add an error `panic` is probably okay(?).

JulianGCalderon:
Hi @YairVaknin-starkware! It would be great if you could remove the unreachable. I think there is a way to do it without changing the signature.

The `air_private_input` function iterates `self.signatures` and builds `Vec<PrivateInput>` with only the variant `PrivateInput::Signature`. Then, it is sorted by asserting that the variant is indeed `Signature`.

Instead of doing this, you could:
1. Iterate `self.signatures` and build a `Vec<PrivateInputSignature>` (the inner concrete type).
2. Sort the vector without needing an unreachable, as we don't have an enum anymore.
3. Build the `Vec<PrivateInput>` while preserving the order.

Let me know your thoughts on this approach.

YairVaknin-starkware:
> Hi @YairVaknin-starkware! It would be great if you could remove the unreachable. I think there is a way to do it without changing the signature.
> 
> The `air_private_input` function iterates `self.signatures` and builds `Vec<PrivateInput>` with only the variant `PrivateInput::Signature`. Then, it is sorted by asserting that the variant is indeed `Signature`.
> 
> Instead of doing this, you could:
> 
> 1. Iterate `self.signatures` and build a `Vec<PrivateInputSignature>` (the inner concrete type).
> 2. Sort the vector without needing an unreachable, as we don't have an enum anymore.
> 3. Build the `Vec<PrivateInput>` while preserving the order.
> 
> Let me know your thoughts on this approach.

HI @JulianGCalderon! I wanted to avoid this method since it would be inefficient to iterate twice. Please take a look at my newest commit here and tell me what you think about this approach.

YairVaknin-starkware:
> Hi @YairVaknin-starkware! It would be great if you could remove the unreachable. I think there is a way to do it without changing the signature.
> 
> The `air_private_input` function iterates `self.signatures` and builds `Vec<PrivateInput>` with only the variant `PrivateInput::Signature`. Then, it is sorted by asserting that the variant is indeed `Signature`.
> 
> Instead of doing this, you could:
> 
> 1. Iterate `self.signatures` and build a `Vec<PrivateInputSignature>` (the inner concrete type).
> 2. Sort the vector without needing an unreachable, as we don't have an enum anymore.
> 3. Build the `Vec<PrivateInput>` while preserving the order.
> 
> Let me know your thoughts on this approach.

This might be the 

> JulianGCalderon

Hi @JulianGCalderon. I opted to sort the signatures before iterating. Please lmk what you think.

JulianGCalderon (approved):

pefontana (approved):

PR #1861, opened 2024-10-30
serde: add `size` field to `Identifier`

tcoratger:


pefontana:
@tcoratger Why do you need this?

tcoratger:
> @tcoratger Why do you need this?

@pefontana because we are building a serde service to be able to exploit Cairo VM result towards rust types and for this, we need to do in rust the equivalent of this in Python (i.e. accessing the `size` item for struct):

https://github.com/kkrt-labs/keth/blob/5dd36473e600936d3a633e58c4b1540480c9a04d/cairo/tests/utils/serde.py#L44

ClementWalter:
regardless of our usage in keth, it's part of the definition of a `Struct` in https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/compiler/identifier_definition.py#L76-L103, so there is no reason not to deserialize it in rust imho

and it's true that we actually is it in python, and so need it to migrate in rust

pefontana (approved):

pefontana:
Sure @tcoratger @ClementWalter !
We can add this

JulianGCalderon (approved):

ClementWalter:
are the ❌ blocking for merging? looks unrelated

pefontana:
> are the ❌ blocking for merging? looks unrelated

nono, everything is fine, if the CI run OK we can merge it!

PR #1862, opened 2024-10-30
Convert relocation table to HashMap<usize, MaybeRelocatable>

### LINKED ISSUE #1860: Memory operations should sometimes accept an Integer
**Describe the bug**

The pythonic implementation of some memory operations accepts either an integer or a relocatable in some isolated cases. These cases are used by `cairo-lang` and break the OS when not supported.

_This issue replaces #1856 which assumed that the downstream problem was caused by the way `cast(0, felt*)` was used. However, the behavior described there is consistent with the way the pythonic implementation works (see code below), so I will close it as "works as intended."_

There are two specific functions used in the OS that depend on support for integers, discussed below.

#### add_relocation_rule()

[Source](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/cairo/lang/vm/memory_dict.py#L111), called from [here](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/cairo/common/segments.cairo#L6).

While the python code clearly declares `dest_ptr: RelocatableValue`, neither function does any checks against it (except `assert src_ptr = dest_ptr;` in [the latter](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/cairo/common/segments.cairo#L9C5-L9C31); note that this assert makes it impossible to work around the issue).

Consider the following code snippet which demonstrates that the relocation system works transparently when `dest_ptr` is a `felt` and not a `felt*`:

```cairo
from starkware.cairo.common.segments import relocate_segment

func main() {
    alloc_locals;

    local temp_segment: felt*;
    local temp_segment_2: felt*;

    %{
        ids.temp_segment = segments.add_temp_segment()
        ids.temp_segment_2 = segments.add_temp_segment()

        # the following prints "-1:0"
        print("temp_segment before relocation:", ids.temp_segment)

        # the following prints "-2:0"
        print("temp_segment_2 before relocation:", ids.temp_segment_2)

    %}

    relocate_segment(src_ptr=temp_segment, dest_ptr=ptr_from_cast_0);
    relocate_segment(src_ptr=temp_segment_2, dest_ptr=ptr_from_cast_1);

    %{
        # the following prints "0"
        print("temp_segment after relocation:", ids.temp_segment)

        # the following prints "1"
        print("temp_segment_2 after relocation:", ids.temp_segment_2)
    %}

    ret;
}
```

#### get_range()

[Source](https://github.com/starkware-libs/cairo-lang/blob/a86e92bfde9c171c0856d7b46580c66e004922f3/src/starkware/cairo/lang/vm/memory_dict.py#L237)

Similarly, this function is happy to operate on `addr` of type `felt` so long as `size == 0`. Unlike the `add_relocation_rule()` issue, this one may be easy to workaround.

Consider the following code snippet which clearly shows that `get_range` can be called as described above:

```cairo
func main() {
    alloc_locals;

    let ptr_from_cast_0: felt* = cast(0, felt*);
    let ptr_from_cast_1: felt* = cast(1, felt*);

    %{
        # note that ptr_from_cast_0 and ptr_from_cast_1 are both internally treated as felt
        # and not felt*:
        # the following prints "0"
        print("ptr_from_cast_0:", ids.ptr_from_cast_0)
        # the following prints "1"
        print("ptr_from_cast_1:", ids.ptr_from_cast_1)

        # get_range(0,0) works
        range = memory.get_range(ids.ptr_from_cast_0, 0)
        # the following prints "[]"
        print("get_range(0,0)", range)

        # get_range(1,0) works
        range = memory.get_range(ids.ptr_from_cast_1, 0)
        # the following prints "[]"
        print("get_range(1,0)", range)

        # get_range(0,1) fails
        # range = memory.get_range(ids.ptr_from_cast_0, 1)

        # get_range(1,1) fails
        # range = memory.get_range(ids.ptr_from_cast_1, 1)
    %}
}
```

**To Reproduce**
Compile (with `cairo-compile`) the above code snippets and run with `cairo-run`, e.g.:

```bash
cairo-compile test_cast.cairo --output test_cast_compiled.json
cairo-run --program=test_cast_compiled.json --print_output --print_info --relocate_prints
```

**Expected behavior**

Changing the underlying types used by `cairo-vm` to be `MaybeRelocatable` could fix this. It seems like a step backwards in sanity, but it does support the behavior that is required by the OS.

I'll propose a concrete suggestion (and I'm happy to implement it, although I'd love feedback as well):

* Modify the `relocation_rules` table [here](https://github.com/lambdaclass/cairo-vm/blob/159f67da19964cc54a95423a69470a26e534a13d/vm/src/vm/vm_memory/memory.rs#L164) to be a `HashMap<usize, MaybeRelocatable>`
* Keep existing accessors for adding a relocation rule (that is, the functions which accept `Relocatable`)
* Add new functions which accept a `MaybeRelocatable` and give them an obnoxious name like `fn add_relocation_rule_maybe_relocatable()`

git commit `58363ad8065ed891e3b14a8191b707677c7c7cb5b9d10030822506786d8d8108` (roughly `v1.0.1`)

**Additional context**

These requirements come from the OS code in [cairo-lang](https://github.com/starkware-libs/cairo-lang). They are very much corner cases as well (triggered in fewer than 1% of blocks on Sepolia testnet), which suggests that the behavior may not be intentional. It could be argued that these are bugs in the OS code... :shrug:

notlesh:
# Convert the relocation table to `MaybeRelocatable`s in order to support ints

## Description

Fixes #1860.

This PR converts the relocation table from a `HashMap<usize, Relocatable>` to a `HashMap<usize, MaybeRelocatable>` in order to support integers.

The original `add_relocation_rule` fns have been left untouched so that they still only support `Relocatable`s and a new `fn add_relocation_rule_maybe_relocatable()` has been added to support using a `MaybeRelocatable`. This should prevent this change from affecting most users.

See #1860 for more details on the motivation for this change.

## Checklist
- [x] Linked to Github Issue
- [x] Unit tests added

### DIFF
--- a/vm/src/vm/vm_core.rs
+++ b/vm/src/vm/vm_core.rs
@@ -1012,6 +1012,20 @@ impl VirtualMachine {
         self.segments.memory.add_relocation_rule(src_ptr, dst_ptr)
     }
 
+    /// Add a new relocation rule.
+    ///
+    /// Will return an error if any of the following conditions are not met:
+    ///   - Source address's segment must be negative (temporary).
+    ///   - Source address's offset must be zero.
+    ///   - There shouldn't already be relocation at the source segment.
+    pub fn add_relocation_rule_maybe_relocatable(
+        &mut self,
+        src_ptr: Relocatable,
+        dst: MaybeRelocatable,
+    ) -> Result<(), MemoryError> {
+        self.segments.memory.add_relocation_rule_maybe_relocatable(src_ptr, dst)
+    }
+
     pub fn gen_arg(&mut self, arg: &dyn Any) -> Result<MaybeRelocatable, MemoryError> {
         self.segments.gen_arg(arg)
     }
--- a/vm/src/vm/vm_core.rs
+++ b/vm/src/vm/vm_core.rs
@@ -1012,7 +1012,11 @@ impl VirtualMachine {
         self.segments.memory.add_relocation_rule(src_ptr, dst_ptr)
     }
 
-    /// Add a new relocation rule.
+    /// Add a new relocation rule, allowing for dst to be a `MaybeRelocatable`.
+    ///
+    /// Relocating memory to anything other than a `Relocatable` is generally not useful, but it
+    /// does make the implementation consistent with the pythonic version. In most cases it is
+    /// advisable to use `add_relocation_rule()`, which only accepts a Relocatable.
     ///
     /// Will return an error if any of the following conditions are not met:
     ///   - Source address's segment must be negative (temporary).
--- a/vm/src/vm/vm_core.rs
+++ b/vm/src/vm/vm_core.rs
@@ -1011,10 +1011,8 @@ impl VirtualMachine {
     pub fn add_relocation_rule(
         &mut self,
         src_ptr: Relocatable,
-        #[cfg(not(feature = "extensive_hints"))]
-        dst_ptr: Relocatable,
-        #[cfg(feature = "extensive_hints")]
-        dst_ptr: MaybeRelocatable,
+        #[cfg(not(feature = "extensive_hints"))] dst_ptr: Relocatable,
+        #[cfg(feature = "extensive_hints")] dst_ptr: MaybeRelocatable,
     ) -> Result<(), MemoryError> {
         self.segments.memory.add_relocation_rule(src_ptr, dst_ptr)
     }
--- a/vm/src/vm/vm_core.rs
+++ b/vm/src/vm/vm_core.rs
@@ -1000,6 +1000,10 @@ impl VirtualMachine {
 
     /// Add a new relocation rule.
     ///
+    /// When using feature "extensive_hints" the destination is allowed to be an Integer (via
+    /// MaybeRelocatable). Relocating memory to anything other than a `Relocatable` is generally
+    /// not useful, but it does make the implementation consistent with the pythonic version.
+    ///
     /// Will return an error if any of the following conditions are not met:
     ///   - Source address's segment must be negative (temporary).
     ///   - Source address's offset must be zero.
@@ -1007,27 +1011,12 @@ impl VirtualMachine {
     pub fn add_relocation_rule(
         &mut self,
         src_ptr: Relocatable,
+        #[cfg(not(feature = "extensive_hints"))]
         dst_ptr: Relocatable,
+        #[cfg(feature = "extensive_hints")]
+        dst_ptr: MaybeRelocatable,
     ) -> Result<(), MemoryError> {
         self.segments.memory.add_relocation_rule(src_ptr, dst_ptr)
-    }
-
-    /// Add a new relocation rule, allowing for dst to be a `MaybeRelocatable`.
-    ///
-    /// Relocating memory to anything other than a `Relocatable` is generally not useful, but it
-    /// does make the implementation consistent with the pythonic version. In most cases it is
-    /// advisable to use `add_relocation_rule()`, which only accepts a Relocatable.
-    ///
-    /// Will return an error if any of the following conditions are not met:
-    ///   - Source address's segment must be negative (temporary).
-    ///   - Source address's offset must be zero.
-    ///   - There shouldn't already be relocation at the source segment.
-    pub fn add_relocation_rule_maybe_relocatable(
-        &mut self,
-        src_ptr: Relocatable,
-        dst: MaybeRelocatable,
-    ) -> Result<(), MemoryError> {
-        self.segments.memory.add_relocation_rule_maybe_relocatable(src_ptr, dst)
     }
 
     pub fn gen_arg(&mut self, arg: &dyn Any) -> Result<MaybeRelocatable, MemoryError> {
--- a/vm/src/vm/vm_memory/memory.rs
+++ b/vm/src/vm/vm_memory/memory.rs
@@ -161,7 +161,7 @@ pub struct Memory {
     pub(crate) temp_data: Vec<Vec<MemoryCell>>,
     // relocation_rules's keys map to temp_data's indices and therefore begin at
     // zero; that is, segment_index = -1 maps to key 0, -2 to key 1...
-    pub(crate) relocation_rules: HashMap<usize, Relocatable>,
+    pub(crate) relocation_rules: HashMap<usize, MaybeRelocatable>,
     pub validated_addresses: AddressSet,
     validation_rules: Vec<Option<ValidationRule>>,
 }
@@ -249,13 +249,16 @@ impl Memory {
     // Version of Memory.relocate_value() that doesn't require a self reference
     fn relocate_address(
         addr: Relocatable,
-        relocation_rules: &HashMap<usize, Relocatable>,
+        relocation_rules: &HashMap<usize, MaybeRelocatable>,
     ) -> Result<MaybeRelocatable, MemoryError> {
         if addr.segment_index < 0 {
             // Adjust the segment index to begin at zero, as per the struct field's
             // comment.
             if let Some(x) = relocation_rules.get(&(-(addr.segment_index + 1) as usize)) {
-                return Ok((*x + addr.offset)?.into());
+                return Ok(match x {
+                    MaybeRelocatable::RelocatableValue(r) => (*r + addr.offset)?.into(),
+                    MaybeRelocatable::Int(i) => i.into(),
+                });
             }
         }
         Ok(addr.into())
@@ -289,6 +292,14 @@ impl Memory {
         for index in (0..self.temp_data.len()).rev() {
             if let Some(base_addr) = self.relocation_rules.get(&index) {
                 let data_segment = self.temp_data.remove(index);
+
+                let base_addr = match base_addr {
+                    MaybeRelocatable::RelocatableValue(addr) => addr,
+                    MaybeRelocatable::Int(_) => {
+                        continue;
+                    },
+                };
+
                 // Insert the to-be relocated segment into the real memory
                 let mut addr = *base_addr;
                 if let Some(s) = self.data.get_mut(addr.segment_index as usize) {
@@ -310,7 +321,6 @@ impl Memory {
         self.relocation_rules.clear();
         Ok(())
     }
-
     /// Add a new relocation rule.
     ///
     /// Will return an error if any of the following conditions are not met:
@@ -321,6 +331,20 @@ impl Memory {
         &mut self,
         src_ptr: Relocatable,
         dst_ptr: Relocatable,
+    ) -> Result<(), MemoryError> {
+        self.add_relocation_rule_maybe_relocatable(src_ptr, MaybeRelocatable::RelocatableValue(dst_ptr))
+    }
+
+    /// Add a new relocation rule.
+    ///
+    /// Will return an error if any of the following conditions are not met:
+    ///   - Source address's segment must be negative (temporary).
+    ///   - Source address's offset must be zero.
+    ///   - There shouldn't already be relocation at the source segment.
+    pub(crate) fn add_relocation_rule_maybe_relocatable(
+        &mut self,
+        src_ptr: Relocatable,
+        dst: MaybeRelocatable,
     ) -> Result<(), MemoryError> {
         if src_ptr.segment_index >= 0 {
             return Err(MemoryError::AddressNotInTemporarySegment(
@@ -338,7 +362,7 @@ impl Memory {
             return Err(MemoryError::DuplicatedRelocation(src_ptr.segment_index));
         }
 
-        self.relocation_rules.insert(segment_index, dst_ptr);
+        self.relocation_rules.insert(segment_index, dst);
         Ok(())
     }
 
@@ -646,8 +670,8 @@ pub(crate) trait RelocateValue<'a, Input: 'a, Output: 'a> {
     fn relocate_value(&self, value: Input) -> Result<Output, MemoryError>;
 }
 
-impl RelocateValue<'_, Relocatable, Relocatable> for Memory {
-    fn relocate_value(&self, addr: Relocatable) -> Result<Relocatable, MemoryError> {
+impl RelocateValue<'_, Relocatable, MaybeRelocatable> for Memory {
+    fn relocate_value(&self, addr: Relocatable) -> Result<MaybeRelocatable, MemoryError> {
         if addr.segment_index < 0 {
             // Adjust the segment index to begin at zero, as per the struct field's
             // comment.
@@ -655,10 +679,15 @@ impl RelocateValue<'_, Relocatable, Relocatable> for Memory {
                 .relocation_rules
                 .get(&(-(addr.segment_index + 1) as usize))
             {
-                return (*x + addr.offset).map_err(MemoryError::Math);
+                return Ok(match x {
+                    MaybeRelocatable::RelocatableValue(r) => {
+                        (*r + addr.offset).map_err(MemoryError::Math)?.into()
+                    },
+                    MaybeRelocatable::Int(i) => i.into(),
+                });
             }
         }
-        Ok(addr)
+        Ok(addr.into())
     }
 }
 
--- a/vm/src/vm/vm_memory/memory.rs
+++ b/vm/src/vm/vm_memory/memory.rs
@@ -757,6 +757,7 @@ impl<'a> RelocateValue<'a, &'a MaybeRelocatable, Cow<'a, MaybeRelocatable>> for 
         Ok(match value {
             MaybeRelocatable::Int(_) => Cow::Borrowed(value),
             MaybeRelocatable::RelocatableValue(addr) => {
+                #[allow(clippy::useless_conversion)]
                 Cow::Owned(self.relocate_value(*addr)?.into())
             }
         })
--- a/vm/src/vm/vm_memory/memory.rs
+++ b/vm/src/vm/vm_memory/memory.rs
@@ -161,6 +161,9 @@ pub struct Memory {
     pub(crate) temp_data: Vec<Vec<MemoryCell>>,
     // relocation_rules's keys map to temp_data's indices and therefore begin at
     // zero; that is, segment_index = -1 maps to key 0, -2 to key 1...
+    #[cfg(not(feature = "extensive_hints"))]
+    pub(crate) relocation_rules: HashMap<usize, Relocatable>,
+    #[cfg(feature = "extensive_hints")]
     pub(crate) relocation_rules: HashMap<usize, MaybeRelocatable>,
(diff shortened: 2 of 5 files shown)


pefontana:
Thanks for the contribution @notlesh !
We will take a look at the changes

vm/src/vm/vm_core.rs
```diff
@@ -1012,6 +1012,24 @@ impl VirtualMachine {
...
+    /// Add a new relocation rule, allowing for dst to be a `MaybeRelocatable`.
+    ///
+    /// Relocating memory to anything other than a `Relocatable` is generally not useful, but it
+    /// does make the implementation consistent with the pythonic version. In most cases it is
+    /// advisable to use `add_relocation_rule()`, which only accepts a Relocatable.
+    ///
+    /// Will return an error if any of the following conditions are not met:
+    ///   - Source address's segment must be negative (temporary).
+    ///   - Source address's offset must be zero.
+    ///   - There shouldn't already be relocation at the source segment.
```
Oppen, line 1024: Shouldn't or mustn't? If the former, in which scenario would it be allowed? If the latter, can this function check it and return an error?

Oppen:
This seems to cause a significant slowdown [see workflow](https://github.com/lambdaclass/cairo-vm/actions/runs/11615606688/job/32394842519?pr=1862#step:8:15). Is there any way to avoid the hit for more conforming programs? Is it possible to fix the calling code instead?

vm/src/vm/vm_core.rs
notlesh: "Mustn't" would be more appropriate here, although this verbiage was copied verbatim (it's already used a couple times in the codebase), so I hesitate to make it different.

This condition is checked in `fn Memory::add_relocation_rule_maybe_relocatable()`, which this fn calls: https://github.com/lambdaclass/cairo-vm/blob/159f67da19964cc54a95423a69470a26e534a13d/vm/src/vm/vm_memory/memory.rs#L338

notlesh:
> This seems to cause a significant slowdown [see workflow](https://github.com/lambdaclass/cairo-vm/actions/runs/11615606688/job/32394842519?pr=1862#step:8:15). Is there any way to avoid the hit for more conforming programs? Is it possible to fix the calling code instead?

Good catch, that's a significant performance decrease. I'm a bit surprised by it, actually, because it shouldn't be adding significant overhead. I'll try to run locally to understand it better.

I agree that such a performance impact is not really acceptable for such a corner case... :sweat_smile: 

We have tried fixing the calling code; we come up with [a workaround](https://github.com/keep-starknet-strange/snos/pull/393) but unfortunately it doesn't always work. The OS makes an assert after calling `add_relocation_rule()` which makes workarounds very tricky:

```cairo
%{ memory.add_relocation_rule(src_ptr=ids.src_ptr, dest_ptr=ids.dest_ptr) %}
assert src_ptr = dest_ptr;
```

In this particular case, `dest_ptr` is an `Int(0)` which leaves us with only two options:
* `src_ptr` must be `Int(0)` (which is what the above-mentioned workaround accomplishes)
* `src_ptr` must have a relocation rule which effectively makes it an `Int(0)` (which this PR accoplishes)

Oppen:
> > This seems to cause a significant slowdown [see workflow](https://github.com/lambdaclass/cairo-vm/actions/runs/11615606688/job/32394842519?pr=1862#step:8:15). Is there any way to avoid the hit for more conforming programs? Is it possible to fix the calling code instead?
> 
> Good catch, that's a significant performance decrease. I'm a bit surprised by it, actually, because it shouldn't be adding significant overhead. I'll try to run locally to understand it better.
> 
> I agree that such a performance impact is not really acceptable for such a corner case... 😅
> 
> We have tried fixing the calling code; we come up with [a workaround](https://github.com/keep-starknet-strange/snos/pull/393) but unfortunately it doesn't always work. The OS makes an assert after calling `add_relocation_rule()` which makes workarounds very tricky:
> 
> ```cairo
> %{ memory.add_relocation_rule(src_ptr=ids.src_ptr, dest_ptr=ids.dest_ptr) %}
> assert src_ptr = dest_ptr;
> ```
> 
> In this particular case, `dest_ptr` is an `Int(0)` which leaves us with only two options:
> 
> * `src_ptr` must be `Int(0)` (which is what the above-mentioned workaround accomplishes)
>   
>   * _(this also breaks the way `add_relocation_rule()` is supposed to work)_
> * `src_ptr` must have a relocation rule which effectively makes it an `Int(0)` (which this PR accomplishes)

IIUC, you only need to cover the `Int(0)` case? Maybe we can get away with just special casing that?

notlesh:
> IIUC, you only need to cover the `Int(0)` case? Maybe we can get away with just special casing that?

Yes, this crossed my mind. The cases I've seen the OS fail on come from `cast(0, felt*)` which should mean `Int(0)` is the only case we care about. If so, we could perhaps have a magical value for `segment_index == usize::MAX` or something like that which triggers the desired behavior.

I suspect this overhead would be similar to my original approach, however. I run into compile errors when running `cargo bench`, but I'll try to sort that out and measure the perf locally.

Speaking of perf, I noticed there is a lot of noise in the measurements [here](https://lambdaclass.github.io/cairo-vm/), maybe the delta measured in this PR is within this noise?

Oppen:
> > IIUC, you only need to cover the `Int(0)` case? Maybe we can get away with just special casing that?
> 
> Yes, this crossed my mind. The cases I've seen the OS fail on come from `cast(0, felt*)` which should mean `Int(0)` is the only case we care about. If so, we could perhaps have a magical value for `segment_index == usize::MAX` or something like that which triggers the desired behavior.
> 

Yes, that or the offset. Both are invalid as `Relocatable` fields so using them as sentinels should be alright.
Note that your original approach significantly increases the size of values (`MaybeRelocatable` is 40 bytes), which tends to be bad for cache locality, and in turn has a more complex logic for addition. The special-case approach, on the other hand, only adds one branch AFAICT. I expect some hit, but I expect it to be barely noticeable.

> I suspect this overhead would be similar to my original approach, however. I run into compile errors when running `cargo bench`, but I'll try to sort that out and measure the perf locally.
> 

I'm not sure `cargo bench` will give you useful information. Our benchmarks are end to end done with Hyperfine.

> Speaking of perf, I noticed there is a lot of noise in the measurements [here](https://lambdaclass.github.io/cairo-vm/), maybe the delta measured in this PR is within this noise?

Erm, not quite. Those measurements are really not looked at. As you noticed, criterion is too noisy for meaningful comparison of commits in a CI. That page is obsolete and we no longer refer to it for performance because of that.
What we use are the inline comments that the CI itself generated (@pefontana copied its output here because for external contributors we can't publish the comment due to lack of credentials), which runs in the same machine the base and the merged refs for proper comparison under Hyperfine.
That said, sometimes there is noise in the Hyperfine run, but most of the time is within 1-2%. When in doubt we can trigger the workflow again. If you suspect this is the case we might run it again and see if it was just a temporary artifact.

pefontana:
Yes let me post the benchmarks again, it seems like a ~15% performance degradation
```
Benchmark Results for unmodified programs :rocket:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base big_factorial | 2.542 ± 0.077 | 2.462 | 2.686 | 1.00 |
| Head big_factorial | 2.845 ± 0.013 | 2.829 | 2.868 | 1.12 ± 0.03 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base big_fibonacci | 2.439 ± 0.023 | 2.407 | 2.471 | 1.00 |
| Head big_fibonacci | 2.826 ± 0.052 | 2.802 | 2.971 | 1.16 ± 0.02 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base blake2s_integration_benchmark | 9.317 ± 0.432 | 9.047 | 10.517 | 1.00 |
| Head blake2s_integration_benchmark | 10.417 ± 0.138 | 10.272 | 10.744 | 1.12 ± 0.05 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base compare_arrays_200000 | 2.602 ± 0.043 | 2.554 | 2.693 | 1.00 |
| Head compare_arrays_200000 | 2.973 ± 0.032 | 2.915 | 3.023 | 1.14 ± 0.02 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base dict_integration_benchmark | 1.701 ± 0.024 | 1.675 | 1.753 | 1.00 |
| Head dict_integration_benchmark | 1.967 ± 0.047 | 1.930 | 2.085 | 1.16 ± 0.03 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base field_arithmetic_get_square_benchmark | 1.441 ± 0.049 | 1.410 | 1.571 | 1.00 |
| Head field_arithmetic_get_square_benchmark | 1.620 ± 0.023 | 1.603 | 1.683 | 1.12 ± 0.04 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base integration_builtins | 9.329 ± 0.077 | 9.171 | 9.415 | 1.00 |
| Head integration_builtins | 10.505 ± 0.084 | 10.400 | 10.634 | 1.13 ± 0.01 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base keccak_integration_benchmark | 9.619 ± 0.125 | 9.476 | 9.916 | 1.00 |
| Head keccak_integration_benchmark | 10.887 ± 0.183 | 10.744 | 11.278 | 1.13 ± 0.02 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base linear_search | 2.642 ± 0.094 | 2.558 | 2.865 | 1.00 |
| Head linear_search | 2.973 ± 0.036 | 2.937 | 3.058 | 1.12 ± 0.04 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base math_cmp_and_pow_integration_benchmark | 1.757 ± 0.023 | 1.734 | 1.801 | 1.00 |
| Head math_cmp_and_pow_integration_benchmark | 2.022 ± 0.036 | 1.996 | 2.122 | 1.15 ± 0.03 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base math_integration_benchmark | 1.717 ± 0.018 | 1.695 | 1.743 | 1.00 |
| Head math_integration_benchmark | 1.952 ± 0.009 | 1.940 | 1.968 | 1.14 ± 0.01 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base memory_integration_benchmark | 1.440 ± 0.038 | 1.408 | 1.518 | 1.00 |
| Head memory_integration_benchmark | 1.651 ± 0.024 | 1.632 | 1.697 | 1.15 ± 0.03 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base operations_with_data_structures_benchmarks | 1.803 ± 0.017 | 1.785 | 1.832 | 1.00 |
| Head operations_with_data_structures_benchmarks | 2.054 ± 0.031 | 2.025 | 2.115 | 1.14 ± 0.02 |

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| Base pedersen | 597.0 ± 6.8 | 590.4 | 610.0 | 1.00 |
| Head pedersen | 653.2 ± 9.3 | 647.7 | 679.2 | 1.09 ± 0.02 |

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| Base poseidon_integration_benchmark | 704.0 ± 5.6 | 695.4 | 712.2 | 1.00 |
| Head poseidon_integration_benchmark | 768.1 ± 8.3 | 757.5 | 784.7 | 1.09 ± 0.01 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base secp_integration_benchmark | 2.073 ± 0.029 | 2.048 | 2.143 | 1.00 |
| Head secp_integration_benchmark | 2.247 ± 0.026 | 2.221 | 2.296 | 1.08 ± 0.02 |

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| Base set_integration_benchmark | 713.2 ± 11.0 | 701.9 | 735.5 | 1.00 |
| Head set_integration_benchmark | 737.5 ± 2.4 | 734.4 | 741.1 | 1.03 ± 0.02 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| Base uint256_integration_benchmark | 5.027 ± 0.057 | 4.972 | 5.110 | 1.00 |
| Head uint256_integration_benchmark | 5.819 ± 0.157 | 5.690 | 6.222 | 1.16 ± 0.03 |
```

pefontana:
Hi @notlesh !
We discussed a litle wit the team, what do you think about this:
To use the OS you already use the feature flag `extensive_hints` no?
Can you move this changes into that feature flag? With this we wont get any performance degradation and also the code will look simpler.
Also I think it will be nice if you could have the two implementations for the functions, to keep the things cleaner, like:
fn relocate_value(&self, addr: Relocatable) -> Result<MaybeRelocatable, MemoryError> {

notlesh:
> Hi @notlesh ! We discussed a litle wit the team, what do you think about this: To use the OS you already use the feature flag `extensive_hints` no? Can you move this changes into that feature flag? With this we wont get any performance degradation and also the code will look simpler. Also I think it will be nice if you could have the two implementations for the functions, to keep the things cleaner, like:
> 
> ```
> #[cfg(feature = "extensive_hints")]
>     fn relocate_address(
>         addr: Relocatable,
>         relocation_rules: &HashMap<usize, Relocatable>,
> #[cfg(feature = "extensive_hints")]
>     fn relocate_address(
>         addr: Relocatable,
>         relocation_rules: &HashMap<usize, MaybeRelocatable>,
>         
> ```
> 
> Tell me what you think

Hey, I think those are great ideas. It certainly solves the perf issue and it gives the VM a way to behave exactly like the pythonic impl.

That said, we are also discussing some small changes to the OS itself that would prevent relocation-to-integers, so this change on the VM may not be needed. I'll keep you posted!

pefontana:
> > Hi @notlesh ! We discussed a litle wit the team, what do you think about this: To use the OS you already use the feature flag `extensive_hints` no? Can you move this changes into that feature flag? With this we wont get any performance degradation and also the code will look simpler. Also I think it will be nice if you could have the two implementations for the functions, to keep the things cleaner, like:
> > ```
> > #[cfg(feature = "extensive_hints")]
> >     fn relocate_address(
> >         addr: Relocatable,
> >         relocation_rules: &HashMap<usize, Relocatable>,
> > #[cfg(feature = "extensive_hints")]
> >     fn relocate_address(
> >         addr: Relocatable,
> >         relocation_rules: &HashMap<usize, MaybeRelocatable>,
> >         
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Tell me what you think
> 
> Hey, I think those are great ideas. It certainly solves the perf issue and it gives the VM a way to behave exactly like the pythonic impl.
> 
> That said, we are also discussing some small changes to the OS itself that would prevent relocation-to-integers, so this change on the VM may not be needed. I'll keep you posted!
> 
> Edit: yes, we are using the `extensive_hints` feature

Great @notlesh !
Hope the OS changes works!

notlesh:
@pefontana cf36b4a is a first pass at putting this all behind `extensive_hints`. It could use some clean up, but feel free to run benchmarks against it to confirm it fixes the perf issue.

pefontana:
@notlesh Can you solve merge conflicts, so I can approve the CI run?

notlesh:
> @notlesh the performance hit seems to be fixed, but some builds in the CI are failing. Could you fix them?

I think I took care of them all now.

vm/src/vm/vm_memory/memory.rs
```diff
@@ -676,6 +757,7 @@ impl<'a> RelocateValue<'a, &'a MaybeRelocatable, Cow<'a, MaybeRelocatable>> for
         Ok(match value {
             MaybeRelocatable::Int(_) => Cow::Borrowed(value),
             MaybeRelocatable::RelocatableValue(addr) => {
+                #[allow(clippy::useless_conversion)]
                 Cow::Owned(self.relocate_value(*addr)?.into())
```
gabrielbosio, line 761:
```suggestion
                Cow::Owned(self.relocate_value(*addr)?)
```

notlesh: This results in the following error:

```
error[E0271]: type mismatch resolving `<MaybeRelocatable as ToOwned>::Owned == Relocatable`
   --> vm/src/vm/vm_memory/memory.rs:760:17
    |
760 |                 Cow::Owned(self.relocate_value(*addr)?)
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `MaybeRelocatable`, found `Relocatable`
```

because `fn relocate_value()` now has a different fn signature (return type is either `Relocatable` or `MaybeRelocatable` depending on whether feature `"extensive_hints"` is on or not (respectively)).

gabrielbosio: Ok, you're right. So you can do this instead:

```suggestion
                #[cfg(not(feature = "extensive_hints"))]
                let v = self.relocate_value(*addr)?.into();
                #[cfg(feature = "extensive_hints")]
                let v = self.relocate_value(*addr)?;

                Cow::Owned(v)
```

I think this way shows the reason why you need to call `into()`.

notlesh: Ok, good point, this makes it more clear what's going on :+1:

Added in 72d27c8

JulianGCalderon (approved):

gabrielbosio (approved):

gabrielbosio:
Merge queue bot removed this PR from the queue due to a job that was failing because of a deprecated version of MacOS. We addressed this on the merged PR #1883. Please merge main into this branch so that we can merge this PR.

notlesh:
> @notlesh please merge main into this branch so that it includes those changes and we can merge this PR.

Done, PTAL!

PR #1887, opened 2024-12-05
feat: implement kzg data availability hints

whichqua:
Context: port of the Starknet OS to Rust.

We need an implementation of the hints used by the Starknet OS in KZG commitments for data availability. This hint relies on private primitives in cairo-vm and must be implemented in this repository.

### Related Issues
https://github.com/starkware-libs/cairo-lang/issues/198

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+        &BigInt3::from_var_name("a", vm, ids_data, ap_tracking)?,
+        &SECP_P,
+    );
+    let b = bls_pack(
+        &BigInt3::from_var_name("b", vm, ids_data, ap_tracking)?,
+        &SECP_P,
+    );
+    let (q, r) = (a * b).div_mod_floor(&BLS_PRIME);
+    let q_reloc = get_relocatable_from_var_name("q", vm, ids_data, ap_tracking)?;
+    let res_reloc = get_relocatable_from_var_name("res", vm, ids_data, ap_tracking)?;
```
FrancoGiachetta, line 61: From what I can see in the hint, the var name should be "r" and not "res", right?

vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+fn as_int(value: &BigInt, prime: &BigInt) -> BigInt {
+    let half_prime = prime.clone() / 2u32;
+    if value > &half_prime {
+        value - prime
+    } else {
+        value.clone()
+    }
+}
```
FrancoGiachetta, line 98: The same as the comment above.

vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+    let mut a = Vec::new();
+    for _ in 0..2 {
+        let residue = num.clone() % BLS_BASE.deref();
+        num /= BLS_BASE.deref();
+        a.push(residue);
+    }
+    a.push(num.clone());
+    assert!(num.abs() < BigInt::from_u128(1 << 127).unwrap());
+    a
+}
```
FrancoGiachetta, line 89: Few things here.
1. Looking at the code as a whole, this `num` does not need to be a reference. Both `q` and `r`, which are the values with which the method is called, are never used again. If you used a reference so as to not pass the parameter by value, you are cloning it and so it has the same impact.
2. You could avoid cloning `num` every time here. If you need num to be mutable, can should make the parameter to be mutable. You should also make the assertion before the push operation so that `num` does not get moved.

FrancoGiachetta (changes requested):

vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
whichqua: This part relates to this part of the hint:
```rs
segments.write_arg(ids.res.address_, split(r))"
```
I need to get the relocatable and write the following value:
```rs
    let res_arg: Vec<MaybeRelocatable> = bls_split(r)
        .into_iter()
        .map(|ref n| Felt252::from(n).into())
        .collect::<Vec<MaybeRelocatable>>();
        
```

vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
whichqua: Resolved by https://github.com/lambdaclass/cairo-vm/pull/1887/commits/cbccca596838a6d4cec945be5096a86d0f0e3fb4

vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
whichqua: Resolved by https://github.com/lambdaclass/cairo-vm/pull/1887/commits/cbccca596838a6d4cec945be5096a86d0f0e3fb4

vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
FrancoGiachetta: Oh I see, but where does this `res` variable come from? Where is it declared?

whichqua: Its declared just before the hint is invoked:
https://github.com/starkware-libs/cairo-lang/blob/8e11b8cc65ae1d0959328b1b4a40b92df8b58595/src/starkware/starknet/core/os/data_availability/bls_field.cairo#L42

whichqua: @FrancoGiachetta Any thoughts on merging this?

whichqua:
@FrancoGiachetta Any way we can get this merged?

gabrielbosio (approved):

FrancoGiachetta (approved):

JulianGCalderon:
Hey @whichqua! I have a question about this PR that just came up. Do you remember why is the hint implementation using `SECP_P` instead of `CAIRO_PRIME`? 

Looking at the hint implementation:
```
from starkware.starknet.core.os.data_availability.bls_utils import BLS_PRIME, pack, split
a = pack(ids.a, PRIME)
b = pack(ids.b, PRIME)
q, r = divmod(a * b, BLS_PRIME)
# By the assumption: |a|, |b| < 2**104 * ((2**86) ** 2 + 2**86 + 1) < 2**276.001.
# Therefore |q| <= |ab| / BLS_PRIME < 2**299.
# Hence the absolute value of the high limb of split(q) < 2**127.
segments.write_arg(ids.q.address_, split(q))
segments.write_arg(ids.res.address_, split(r))"#
```

It looks like it should be using the Cairo prime, rather than SECP_PRIME.

JulianGCalderon:
Hey @whichqua, we will be merging the PR #2078 that fixes the implementation to use the Cairo Prime, rather than SECP_PRIME. If this breaks code of yours, let us know please!

PR #1898, opened 2024-12-20
Add support for subtractions contenting references as right hand side operands

FrancoGiachetta:
# Add support for subtractions contenting references as right hand side operands

## Description

The `VM` reads `let` expressions like ` c - tmp` as an addition instead of a subtraction, where `c` is a constant and `tmp` is a `local` variable. This is because currently only additions are supported. This is not an issue with variables declared with the `let` keyword because those are resolved at compile time.

This PR introduces a little refactor to allow subtractions with references as the rhs operand. It does it by saving the sign with which the reference is defined.

## Checklist
- [X] Linked to Github Issue
- [X] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 #### [2.0.0-rc2] - 2024-12-12
 
-* feat: Add support for subtractions contenting references as right hand side operands [#1898](https://github.com/lambdaclass/cairo-vm/pull/1898)
+* feat: Add support for subtractions containing references as right hand side operands [#1898](https://github.com/lambdaclass/cairo-vm/pull/1898)
 
 * fix: Change wildcard getrandom dependency.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 #### Upcoming Changes
 
+#### [2.0.0-rc3] - 2024-12-26
+
+* chore: update cairo-lang dependencies to 2.10.0-rc.0 #[1901](https://github.com/lambdaclass/cairo-vm/pull/1901)
+
 #### [2.0.0-rc2] - 2024-12-12
 
 * feat: Add support for subtractions contenting references as right hand side operands [#1898](https://github.com/lambdaclass/cairo-vm/pull/1898)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### [2.0.0-rc2] - 2024-12-12
 
-* fix: Add support for subtractions contenting references as right hand side operands [#1898](https://github.com/lambdaclass/cairo-vm/pull/1898)
+* feat: Add support for subtractions contenting references as right hand side operands [#1898](https://github.com/lambdaclass/cairo-vm/pull/1898)
 
 * fix: Change wildcard getrandom dependency.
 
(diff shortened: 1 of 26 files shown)


examples/custom_hint/custom_hint.cairo
```diff
@@ -1,6 +1,9 @@
...
-    let a = 17;
-    // Use custom hint to print the value of a
+    alloc_locals;
+    local tmp = 3;
+    let a = 17 - tmp;
+
     %{ print(ids.a) %}
+
     return ();
 }
```
JulianGCalderon, line 9: This file shouldn't be changed, right?

JulianGCalderon:
I think this is OK as it's a simple and fast fix, but we should create an issue to tackle the underlying problem later on: supporting arbitrary expressions like cairo-lang does. I think we would eventually need to define an AST for the expression.

gabrielbosio:
> Since the compiler simplifies the expression whenever it can, there would be no problem in let expressions which only use constants or even variables defined with the let keyword. However, when we use variables defined with the local keyword this is not the case anymore because in doing so we are introducing the use of references.

PR looks good but the description could be rephrased, I think it's not clear what's the issue. It can be simplified to something like:

> The VM reads let expressions as `c - tmp` as an addition instead of a subtraction, where `c` is a constant and `tmp` is a local variable.

Then it can describe why this happens.
Also:

> Thus, only supporting additions would make expressiones like this one: 17 - [fp] to equivalent to 17 + [fp] because the sign of the reference is never taken into account.

Is `[fp]` interpreted as a local variable?

examples/custom_hint/custom_hint.cairo
FrancoGiachetta: You are right. I changed it to match the code the issue and forgot to change it again. Thanks

FrancoGiachetta:
The `fp` points to the first `local` declared, `fp + 1` to the second and so on. I'll change the description to make it clearer.

examples/custom_hint/custom_hint.cairo
FrancoGiachetta: Done

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 #### [2.0.0-rc2] - 2024-12-12
 
+* feat: Add support for subtractions contenting references as right hand side operands [#1898](https://github.com/lambdaclass/cairo-vm/pull/1898)
```
gabrielbosio, line 11: Is `contenting` the right word here? Or is this one:

```suggestion
* feat: Add support for subtractions containing references as right hand side operands [#1898](https://github.com/lambdaclass/cairo-vm/pull/1898)
```

vm/src/types/errors/program_errors.rs
```diff
@@ -9,6 +9,8 @@ pub enum ProgramError {
     IO(#[from] std::io::Error),
     #[error(transparent)]
     Parse(#[from] serde_json::Error),
+    #[error("The \"{0}\" operation not supported")]
```
gabrielbosio, line 12:
```suggestion
    #[error("The \"{0}\" operation is not supported")]
```

gabrielbosio (approved):

JulianGCalderon (approved):

PR #1900, opened 2024-12-24
stavbe: change AirPrivateInputSerializable fields to public

Stavbe:
Change AirPrivateInputSerializable fields to be public, so I can use this struct.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1900)


Stavbe:
Hey :)
I actually need the trace and mem ...

gabrielbosio:
Ok, I'm trying to get a bit more of context here so what's the use case here? Because when calling `AirPrivateInput::to_serializable` one passes both trace and mem paths: https://github.com/lambdaclass/cairo-vm/blob/259487e5c12fe1fd5912ea540b075a5f4a98456a/vm/src/air_private_input.rs#L154-L158

Stavbe:
?

Stavbe (commented):
Sorry, I missed that response.  
Yes, the JSON file (which was created after calling `to_serializable`) contains the `mem` and `trace`.  
Now, when trying to read it (`private.json`), I need to define a new struct to use those fields, which is redundant.  
For example, I read `pub.json` into the `pub struct PublicInput<'a>`.

> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1900#-:-OIAHLuz2ifTKuivTg0s:b3643hu)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @Oppen, and @pefontana)

JulianGCalderon (commented):
Hi @Stavbe! Would exposing only `trace_path` and `memory_path`suffice? Or maybe we could add getter methods (i.e `get_trace_path()`).

Stavbe (commented):
Yes. that could work. Fixed :)

> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1900#-:-OL4p_hH6tcic-kvyuFx:b3of0le)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @noaov1, @Oppen, @pefontana, and @yuvalsw)

JulianGCalderon (approved):

yuvalsw (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1900#-:-OLUPVxQ3cjvq--Le0cl:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @noaov1, @Oppen, and @pefontana)

___
*[`-- commits` line 5 at r3](https://reviewable.io/reviews/lambdaclass/cairo-vm/1900#-OLUPUJfDUMyCPS8C_t2:-OLUPUJg3nUyCqKkuLQa:b9kq1g3):*
needs rebase?



_Code quote:_
```
New commits in r2 on 11/03/2025 at 18:49, probably rebased from r1:
- 61a1c75: stavbe: change AirPrivateInputSerializable fields to public

New commits in r3 on 16/03/2025 at 15:52:
- fda1f4c: Merge branch 'main' into stavbe/air-private-input-fields-to-pub
```

Stavbe (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1900)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @noaov1, @Oppen, and @pefontana)

vm/src/air_private_input.rs
```diff
@@ -12,8 +12,8 @@ use crate::Felt252;
 // Serializable format, matches the file output of the python implementation
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct AirPrivateInputSerializable {
-    trace_path: String,
-    memory_path: String,
+    pub trace_path: String,
+    pub memory_path: String,
```
gabrielbosio, line 16: If the idea is to read these fields and to not write them, I think it would be better to add getters for these fields instead of making them public

gabrielbosio (changes requested):

vm/src/air_private_input.rs
Stavbe: It wouldn't work because I read it from json file.

gabrielbosio: I'm not completely sure how reading the struct from a json file restricts implementing getters instead of making these fields public.
Let me rephrase my suggestion just in case. For example, instead of accessing these fields like this:

```rust
let trace_path = air_private_input_serializable.trace_path;
let memory_path = air_private_input_serializable.memory_path;
```

Shouldn't this also work?

```rust
let trace_path = air_private_input_serializable.get_trace_path();
let memory_path = air_private_input_serializable.get_memory_path();
```

Stavbe: Before being able to call the getter, you need to construct the sturct from the json: 
`fn deserialize_inputs<'a>(
    public_input_string: &'a str,
    private_input_string: &'a str,
) -> Result<(PublicInput<'a>, PrivateInput), VmImportError> {
    {
        Ok((
            sonic_rs::from_str(public_input_string)?,
            sonic_rs::from_str(private_input_string)?,
        ))
    }
    #[cfg(not(feature = "std"))]
    {
        Ok((
            serde_json::from_str(public_input_string)?,
            serde_json::from_str(private_input_string)?,
        ))
    }
`

gabrielbosio: Ok, but if you make `trace_path` and `memory_path` public, wouldn't you also need to construct the struct from the json?

PR #1909, opened 2025-01-05
Fix_no_trace_padding_flow_in_proof_mode

YairVaknin-starkware:
# TITLE

## Description
Avoid using `safe_div_usize` when disabling trace padding, and use `ceil_div` instead.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

- - -
This change is [Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* fix(BREAKING): Fix no trace padding flow in proof mode [#1909](https://github.com/lambdaclass/cairo-vm/pull/1909)
+
 * refactor: Limit ret opcode decodeing to Cairo0's standards. [#1925](https://github.com/lambdaclass/cairo-vm/pull/1925)
 
 #### [2.0.0-rc4] - 2025-01-23
--- a/vm/src/cairo_run.rs
+++ b/vm/src/cairo_run.rs
@@ -34,6 +34,13 @@ pub struct CairoRunConfig<'a> {
     pub dynamic_layout_params: Option<CairoLayoutParams>,
     pub proof_mode: bool,
     pub secure_run: Option<bool>,
+    /// Disable padding of the trace.
+    /// By default, the trace is padded to accommodate the expected builtins-n_steps relationships
+    /// according to the layout.
+    /// When the padding is disabled:
+    /// - It doesn't modify/pad n_steps.
+    /// - It still pads each builtin segment to the next power of 2 (w.r.t the number of used
+    ///   instances of the builtin) compared to their sizes at the end of the execution.
     pub disable_trace_padding: bool,
     pub allow_missing_builtins: Option<bool>,
 }
@@ -75,6 +82,7 @@ pub fn cairo_run_program_with_initial_scope(
         cairo_run_config.dynamic_layout_params.clone(),
         cairo_run_config.proof_mode,
         cairo_run_config.trace_enabled,
+        cairo_run_config.disable_trace_padding,
     )?;
 
     cairo_runner.exec_scopes = exec_scopes;
@@ -162,6 +170,7 @@ pub fn cairo_run_pie(
         cairo_run_config.dynamic_layout_params.clone(),
         false,
         cairo_run_config.trace_enabled,
+        cairo_run_config.disable_trace_padding,
     )?;
 
     let end = cairo_runner.initialize(allow_missing_builtins)?;
@@ -235,6 +244,7 @@ pub fn cairo_run_fuzzed_program(
         cairo_run_config.dynamic_layout_params.clone(),
         cairo_run_config.proof_mode,
         cairo_run_config.trace_enabled,
+        cairo_run_config.disable_trace_padding,
     )?;
 
     let _end = cairo_runner.initialize(allow_missing_builtins)?;
--- a/vm/src/vm/runners/builtin_runner/mod.rs
+++ b/vm/src/vm/runners/builtin_runner/mod.rs
@@ -532,17 +532,29 @@ impl BuiltinRunner {
                 Ok((used, used))
             }
             _ => {
-                let used = self.get_used_cells(&vm.segments)?;
-                let size = self.get_allocated_memory_units(vm)?;
-                if used > size {
-                    return Err(InsufficientAllocatedCellsError::BuiltinCells(Box::new((
-                        self.name(),
-                        used,
-                        size,
-                    )))
-                    .into());
+                let used_cells = self.get_used_cells(&vm.segments)?;
+                if vm.disable_trace_padding {
+                    // If trace padding is disabled, we pad the used cells to still ensure that the
+                    // number of instances is a power of 2.
+                    let num_instances = self.get_used_instances(&vm.segments)?;
+                    let padded_used_cells = if num_instances > 0 {
+                        num_instances.next_power_of_two() * self.cells_per_instance() as usize
+                    } else {
+                        0
+                    };
+                    Ok((used_cells, padded_used_cells))
+                } else {
+                    let size = self.get_allocated_memory_units(vm)?;
+                    if used_cells > size {
+                        return Err(InsufficientAllocatedCellsError::BuiltinCells(Box::new((
+                            self.name(),
+                            used_cells,
+                            size,
+                        )))
+                        .into());
+                    }
+                    Ok((used_cells, size))
                 }
-                Ok((used, size))
             }
         }
     }
@@ -693,11 +705,13 @@ impl From<ModBuiltinRunner> for BuiltinRunner {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::cairo_run::{cairo_run, CairoRunConfig};
     use crate::hint_processor::builtin_hint_processor::builtin_hint_processor_definition::BuiltinHintProcessor;
     use crate::relocatable;
     use crate::types::builtin_name::BuiltinName;
     use crate::types::instance_definitions::mod_instance_def::ModInstanceDef;
     use crate::types::instance_definitions::LowRatio;
+    use crate::types::layout_name::LayoutName;
     use crate::types::program::Program;
     use crate::utils::test_utils::*;
     use crate::vm::errors::memory_errors::InsufficientAllocatedCellsError;
@@ -875,6 +889,111 @@ mod tests {
             .unwrap();
 
         assert_eq!(builtin.get_allocated_memory_units(&cairo_runner.vm), Ok(5));
+    }
+
+    #[test]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+    fn compare_proof_mode_with_and_without_disable_trace_padding() {
+        const PEDERSEN_TEST: &[u8] =
+            include_bytes!("../../../../../cairo_programs/proof_programs/pedersen_test.json");
+        const BIGINT_TEST: &[u8] =
+            include_bytes!("../../../../../cairo_programs/proof_programs/bigint.json");
+        const POSEIDON_HASH_TEST: &[u8] =
+            include_bytes!("../../../../../cairo_programs/proof_programs/poseidon_hash.json");
+
+        let program_files = vec![PEDERSEN_TEST, BIGINT_TEST, POSEIDON_HASH_TEST];
+
+        for program_data in program_files {
+            let config_false = CairoRunConfig {
+                disable_trace_padding: false,
+                proof_mode: true,
+                layout: LayoutName::all_cairo,
+                ..Default::default()
+            };
+            let mut hint_processor_false = BuiltinHintProcessor::new_empty();
+            let runner_false =
+                cairo_run(program_data, &config_false, &mut hint_processor_false).unwrap();
+            let last_step_false = runner_false.vm.current_step;
+
+            assert!(last_step_false.is_power_of_two());
+
+            let config_true = CairoRunConfig {
+                disable_trace_padding: true,
+                proof_mode: true,
+                layout: LayoutName::all_cairo,
+                ..Default::default()
+            };
+            let mut hint_processor_true = BuiltinHintProcessor::new_empty();
+            let runner_true =
+                cairo_run(program_data, &config_true, &mut hint_processor_true).unwrap();
+            let last_step_true = runner_true.vm.current_step;
+
+            // Ensure the last step is not a power of two - true for this specific program, not always.
+            assert!(!last_step_true.is_power_of_two());
+
+            assert!(last_step_true < last_step_false);
+
+            let builtin_runners_false = &runner_false.vm.builtin_runners;
+            let builtin_runners_true = &runner_true.vm.builtin_runners;
+            assert_eq!(builtin_runners_false.len(), builtin_runners_true.len());
+            // Compare allocated instances for each pair of builtin runners.
+            for (builtin_runner_false, builtin_runner_true) in builtin_runners_false
+                .iter()
+                .zip(builtin_runners_true.iter())
+            {
+                assert_eq!(builtin_runner_false.name(), builtin_runner_true.name());
+                match builtin_runner_false {
+                    BuiltinRunner::Output(_) | BuiltinRunner::SegmentArena(_) => {
+                        continue;
+                    }
+                    _ => {}
+                }
+                let (_, allocated_size_false) = builtin_runner_false
+                    .get_used_cells_and_allocated_size(&runner_false.vm)
+                    .unwrap();
+                let (used_cells_true, allocated_size_true) = builtin_runner_true
+                    .get_used_cells_and_allocated_size(&runner_true.vm)
+                    .unwrap();
+                let n_allocated_instances_false = safe_div_usize(
+                    allocated_size_false,
+                    builtin_runner_false.cells_per_instance() as usize,
+                )
+                .unwrap();
+                let n_allocated_instances_true = safe_div_usize(
+                    allocated_size_true,
+                    builtin_runner_true.cells_per_instance() as usize,
+                )
+                .unwrap();
+                assert!(
+                    n_allocated_instances_false.is_power_of_two()
+                        || n_allocated_instances_false == 0
+                );
+                assert!(
+                    n_allocated_instances_true.is_power_of_two() || n_allocated_instances_true == 0
+                );
+                // Checks that the number of allocated instances is different when trace padding is
+                // enabled/disabled. Holds for this specific program, not always (that is, in other
+                // programs, padding may be of size 0, or the same).
+                assert!(
+                    n_allocated_instances_true == 0
+                        || n_allocated_instances_true != n_allocated_instances_false
+                );
+
+                // Since the last instance of the builtin isn't guaranteed to have a full output,
+                // the number of used_cells might not be a multiple of cells_per_instance, so we
+                // make sure that the discrepancy is up to the number of output cells.
+                // This is the same for both cases, so we only check one (true).
+                let n_output_cells = builtin_runner_true.cells_per_instance() as usize
+                    - builtin_runner_true.n_input_cells() as usize;
+                assert!(
+                    used_cells_true + n_output_cells
+                        >= (builtin_runner_true.cells_per_instance() as usize)
+                            * builtin_runner_true
+                                .get_used_instances(&runner_true.vm.segments)
+                                .unwrap()
+                );
+            }
+        }
     }
 
     #[test]
--- a/bench/criterion_benchmark.rs
+++ b/bench/criterion_benchmark.rs
@@ -33,6 +33,7 @@ fn build_many_runners(c: &mut Criterion) {
                     black_box(None),
                     black_box(false),
                     black_box(false),
+                    black_box(false),
                 )
                 .unwrap(),
             );
@@ -51,6 +52,7 @@ fn load_program_data(c: &mut Criterion) {
                     &program,
                     LayoutName::starknet_with_keccak,
                     None,
+                    false,
                     false,
                     false,
                 )
--- a/bench/iai_benchmark.rs
+++ b/bench/iai_benchmark.rs
@@ -37,6 +37,7 @@ fn build_runner() {
         None,
         false,
         false,
+        false,
     )
     .unwrap();
     core::mem::drop(black_box(runner));
@@ -52,6 +53,7 @@ fn build_runner_helper() -> CairoRunner {
         &program,
         LayoutName::starknet_with_keccak,
         None,
+        false,
         false,
         false,
     )
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -258,6 +258,7 @@ pub fn cairo_run_program(
         cairo_run_config.dynamic_layout_params.clone(),
         runner_mode,
         cairo_run_config.trace_enabled,
+        false,
     )?;
     let end = runner.initialize(cairo_run_config.proof_mode)?;
     load_arguments(&mut runner, &cairo_run_config, main_func, initial_gas)?;
--- a/hint_accountant/src/main.rs
+++ b/hint_accountant/src/main.rs
@@ -49,7 +49,7 @@ fn run() {
             whitelists.push(whitelist_file.allowed_hint_expressions);
         }
     }
-    let mut vm = VirtualMachine::new(false);
+    let mut vm = VirtualMachine::new(false, false);
     let mut hint_executor = BuiltinHintProcessor::new_empty();
     let (ap_tracking_data, reference_ids, references, mut exec_scopes, constants) = (
         ApTracking::default(),
--- a/vm/src/hint_processor/builtin_hint_processor/secp/cairo0_hints.rs
+++ b/vm/src/hint_processor/builtin_hint_processor/secp/cairo0_hints.rs
@@ -372,7 +372,7 @@ mod tests {
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_is_on_curve_2() {
-        let mut vm = VirtualMachine::new(false);
+        let mut vm = VirtualMachine::new(false, false);
         vm.set_fp(1);
         let ids_data = non_continuous_ids_data![("is_on_curve", -1)];
         vm.segments = segments![((1, 0), 1)];
@@ -404,7 +404,7 @@ mod tests {
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_compute_q_mod_prime() {
-        let mut vm = VirtualMachine::new(false);
+        let mut vm = VirtualMachine::new(false, false);
 
         let ap_tracking = ApTracking::default();
 
@@ -431,7 +431,7 @@ mod tests {
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_compute_ids_high_low() {
-        let mut vm = VirtualMachine::new(false);
+        let mut vm = VirtualMachine::new(false, false);
 
         let value = BigInt::from(25);
         let shift = BigInt::from(12);
@@ -501,7 +501,7 @@ mod tests {
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_r1_get_point_from_x() {
-        let mut vm = VirtualMachine::new(false);
+        let mut vm = VirtualMachine::new(false, false);
         vm.set_fp(10);
 
         let ids_data = non_continuous_ids_data![("x", -10), ("v", -7)];
@@ -560,7 +560,7 @@ mod tests {
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_reduce_value() {
-        let mut vm = VirtualMachine::new(false);
+        let mut vm = VirtualMachine::new(false, false);
 
         //Initialize fp
         vm.run_context.fp = 10;
@@ -616,7 +616,7 @@ mod tests {
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_reduce_x() {
-        let mut vm = VirtualMachine::new(false);
+        let mut vm = VirtualMachine::new(false, false);
 
         //Initialize fp
         vm.run_context.fp = 10;
--- a/vm/src/tests/cairo_run_test.rs
+++ b/vm/src/tests/cairo_run_test.rs
@@ -1185,6 +1185,7 @@ fn run_program_with_custom_mod_builtin_params(
         cairo_run_config.dynamic_layout_params,
         cairo_run_config.proof_mode,
         cairo_run_config.trace_enabled,
+        cairo_run_config.disable_trace_padding,
     )
     .unwrap();
 
--- a/vm/src/tests/mod.rs
+++ b/vm/src/tests/mod.rs
@@ -116,6 +116,7 @@ fn run_cairo_1_entrypoint(
         None,
         false,
         false,
+        false,
     )
     .unwrap();
 
@@ -219,6 +220,7 @@ fn run_cairo_1_entrypoint_with_run_resources(
         &(contract_class.clone().try_into().unwrap()),
         LayoutName::all_cairo,
         None,
+        false,
         false,
         false,
     )
--- a/vm/src/utils.rs
+++ b/vm/src/utils.rs
@@ -234,7 +234,7 @@ pub mod test_utils {
 
     macro_rules! vm_with_range_check {
         () => {{
-            let mut vm = VirtualMachine::new(false);
+            let mut vm = VirtualMachine::new(false, false);
             vm.builtin_runners = vec![
                 $crate::vm::runners::builtin_runner::RangeCheckBuiltinRunner::<8>::new(
                     Some(8),
@@ -255,12 +255,13 @@ pub mod test_utils {
                 None,
                 false,
                 false,
+                false,
             )
             .unwrap()
         };
         ($program:expr, $layout:expr) => {
             crate::vm::runners::cairo_runner::CairoRunner::new(
-                &$program, $layout, None, false, false,
+                &$program, $layout, None, false, false, false,
             )
             .unwrap()
         };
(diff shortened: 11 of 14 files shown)


yuvalsw (changes requested):
> Reviewed 14 of 14 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 5 unresolved discussions (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @YairVaknin-starkware)

___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 202 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5mccH7RpJlfD3PVul:-OG5mccH7RpJlfD3PVum:bgro98e) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L202)):*
> ```XML
>                     Some(0) => Ok(0),
>                     Some(ratio) => {
>                         if !vm.disable_trace_padding {
> ```

please explain (doc) why it isn't relevant with disable_trace_padding
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 919 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5nNQA8-GNeoSouw6h:-OG5nNQA8-GNeoSouw6i:bbnv25u) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L919)):*
> ```XML
> 
>         // Ensure the last step is not a power of two - true for this specific program, not always.
>         assert!(!last_step_true.is_power_of_two());
> ```

if you know the exact number, maybe best to assert it's that number. (but do comment that the important thing here is to assert it did not round to a power of 2).
Same above - you can check the exact power of 2 you expect.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 929 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5ouln0h29fl1uDqeA:-OG5ouln0h29fl1uDqeB:b-a7hoym) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L929)):*
> ```XML
>         for (builtin_runner_false, builtin_runner_true) in builtin_runners_false
>             .iter()
>             .zip(builtin_runners_true.iter())
> ```

consider using zip_eq
This isn't checking everything if they are not of the same length.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 222 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5kwVK29xMct83RG4g:-OG5kwVK29xMct83RG4h:b-45g21t) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/21171ad58ea4d9e4f0c41ce78713e35061a8ef7a/vm/src/vm/runners/builtin_runner/mod.rs#L222)):*
> ```XML
> 
>                         let allocated_instances = if vm.disable_trace_padding {
>                             div_ceil(numerator, ratio as usize)
> ```

Please add doc saying why it's ok that it's not accurate (not used, right?) and why it's related to no padding.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 228 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5l_5t5ENUJQxLuWyx:-OG5l_5t5ENUJQxLuWyy:b-3qnc53) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/21171ad58ea4d9e4f0c41ce78713e35061a8ef7a/vm/src/vm/runners/builtin_runner/mod.rs#L228)):*
> ```XML
>                         };
> 
>                         Ok(allocated_instances)
> ```

suggestion



_Suggestion:_
```XML
                        let ratio_den = self.ratio_den().or(1);
                        let numerator = vm.current_step * ratio_den as usize;

                        let allocated_instances = if vm.disable_trace_padding {
                            div_ceil(numerator, ratio as usize)
                        } else {
                            safe_div_usize(numerator, ratio as usize)
                                .map_err(|_| MemoryError::ErrorCalculatingMemoryUnits)?
                        };

                        Ok(allocated_instances)
```

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 4 unresolved discussions (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @yuvalsw)

___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 222 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5kwVK29xMct83RG4g:-OGPedHz4hCgB6zvvjng:bb6un1j) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/21171ad58ea4d9e4f0c41ce78713e35061a8ef7a/vm/src/vm/runners/builtin_runner/mod.rs#L222)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

Please add doc saying why it's ok that it's not accurate (not used, right?) and why it's related to no padding.
</blockquote></details>

Done
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 228 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5l_5t5ENUJQxLuWyx:-OGPg4-_Dn-5UwVG93J7:bb6un1j) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/21171ad58ea4d9e4f0c41ce78713e35061a8ef7a/vm/src/vm/runners/builtin_runner/mod.rs#L228)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

suggestion
</blockquote></details>

Done
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 202 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5mccH7RpJlfD3PVul:-OGPduLUB_ms9trdEKev:bb6un1j) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L202)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

please explain (doc) why it isn't relevant with disable_trace_padding
</blockquote></details>

Done
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 919 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5nNQA8-GNeoSouw6h:-OGPeea41tlCYvYqkHcu:b5zc8bn) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L919)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

if you know the exact number, maybe best to assert it's that number. (but do comment that the important thing here is to assert it did not round to a power of 2).
Same above - you can check the exact power of 2 you expect.
</blockquote></details>

I think I prefer it the way it is now. How strongly do you feel about changing it?
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 929 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5ouln0h29fl1uDqeA:-OGPektC1bSYlX3-FAW9:byz9j9d) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L929)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

consider using zip_eq
This isn't checking everything if they are not of the same length.
</blockquote></details>

They don't import the needed crate in cargo.toml, and would like to avoid adding to it. 
So just added an assertion they are of equal lengths.

yuvalsw (changes requested):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 3 unresolved discussions (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @YairVaknin-starkware)

___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 222 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5kwVK29xMct83RG4g:-OGQFQzGDJ7avioBHIlL:b4pxqzr) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/21171ad58ea4d9e4f0c41ce78713e35061a8ef7a/vm/src/vm/runners/builtin_runner/mod.rs#L222)):*
<details><summary><i>Previously, YairVaknin-starkware wrote…</i></summary><blockquote>

Done
</blockquote></details>

also "saying why it's ok that it's not accurate" please
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 919 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5nNQA8-GNeoSouw6h:-OGQGEHK72TbOCAU70gs:b-ts56ja) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L919)):*
<details><summary><i>Previously, YairVaknin-starkware wrote…</i></summary><blockquote>

I think I prefer it the way it is now. How strongly do you feel about changing it?
</blockquote></details>

Not super strong. Just generally - tests should have less code and be more specific so that you can be more certain their code itself is good (so they test what you want). 
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 205 at r3](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGQEkD586FyG0KvMahR:-OGQEkD586FyG0KvMahS:b4s1mrx) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/0e23543f710651df8105d99c9f2530df8ffa863a/vm/src/vm/runners/builtin_runner/mod.rs#L205)):*
> ```XML
>                             // if we disable trace padding, we don't need to check the min_step,
>                             // since we won't have the guarantee of number of steps being padded to
>                             // accommodate each builtin size.
> ```

please move to above the if (as it refers to what doesn't happen in the if...).



_Code quote:_
```XML
                            // if we disable trace padding, we don't need to check the min_step,
                            // since we won't have the guarantee of number of steps being padded to
                            // accommodate each builtin size.
```

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 3 unresolved discussions (waiting on @fmoletta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @yuvalsw)

___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 222 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5kwVK29xMct83RG4g:-OGTkUKE7JeRdS9fQizZ:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/21171ad58ea4d9e4f0c41ce78713e35061a8ef7a/vm/src/vm/runners/builtin_runner/mod.rs#L222)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

also "saying why it's ok that it's not accurate" please
</blockquote></details>

Done.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 919 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OG5nNQA8-GNeoSouw6h:-OGTkYZT4WShAlS-xZTb:b-lw3bxt) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/6cd85de17270e74794ecac2235c565bf0a5c591f/vm/src/vm/runners/builtin_runner/mod.rs#L919)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

Not super strong. Just generally - tests should have less code and be more specific so that you can be more certain their code itself is good (so they test what you want).
</blockquote></details>

I agree. And here this is the thing we want to test.

vm/src/vm/runners/builtin_runner/mod.rs
```diff
@@ -877,6 +891,59 @@ mod tests {
         assert_eq!(builtin.get_allocated_memory_units(&cairo_runner.vm), Ok(5));
     }
 
+    #[test]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+    fn proof_mode_with_and_without_disable_trace_padding() {
```
FrancoGiachetta, line 896: I would split this test in two, one with `disable_trace_padding` and one without. By doing this, it is easier to detect the reason of a test failure (if any).

YairVaknin-starkware: Hi, so I wanted to compare both runs to each other and would like them to fail if they don't match up to each other as expected. I this should be left as a single test.

gabrielbosio:
Please provide more context in the description of the PR. For example, Is this change a bugfix of the VM or an improvement for a specific use? This is important because changing to `div_ceil` would imply that `get_allocated_instances` would not return an error in some cases where `main` would do, making callers of this function to work different.

YairVaknin-starkware:
> Please provide more context in the description of the PR. For example, Is this change a bugfix of the VM or an improvement for a specific use? This is important because changing to `div_ceil` would imply that `get_allocated_instances` would not return an error in some cases where `main` would do, making callers of this function to work different.

Hi, so we would like it to not fail the division when trace padding is disabled, since it won't be a multiple of the number of steps in this case. This check is only done in proof_mode (in finalize_segments), and most of the time it will throw an error when we disable trace padding, which would not let us complete the non padded run.

vm/src/vm/runners/builtin_runner/mod.rs
YairVaknin-starkware: Sure, done.

yuvalsw (changes requested):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 2 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @YairVaknin-starkware)

___
*[a discussion](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGeEKVI1qx_-OsnmaBq:-OGeEKVI1qx_-OsnmaBr:b-hdgb4) (no related file):*
Blocking for now as we need to understand if the other logic we talked about overrides this one or lives along it. If it overrides, then maybe no need to merge this one. You can change this PR to the new logic or send a different one, as you see fit.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* fix: Fix no trace padding flow in proof mode [#1909](https://github.com/lambdaclass/cairo-vm/pull/1909)
```
Oppen, line 5: This is an API breaking change. It must be documented as such.

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: 13 of 14 files reviewed, 3 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @pefontana, and @yuvalsw)

___
*[a discussion](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGeEKVI1qx_-OsnmaBq:-OGyToSb1H3S0ZdSkq19:b-72wupf) (no related file):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

Blocking for now as we need to understand if the other logic we talked about overrides this one or lives along it. If it overrides, then maybe no need to merge this one. You can change this PR to the new logic or send a different one, as you see fit.
</blockquote></details>

Changed to a different logic as discussed

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 9 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @pefontana, and @yuvalsw)

___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 223 at r5](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGyxNTR4xDzw0WfjxUT:-OGzBvi7-W1CP8IszUos:b-jn06df) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/d6d42f5653370e55e0732c432f6e30ea34d768ad/vm/src/vm/runners/builtin_runner/mod.rs#L223)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

How come this is not needed? Aren't the safe_divs failing? Ratio is still irrelevant in the case of disable_trace_padding.
</blockquote></details>

We don't reach here when disabling trace padding. TAL at ` fn get_used_cells_and_allocated_size`.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 223 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGyUHlqEC0m_AJzus3o:-OGzBn7zAh_II-Nhb6GY:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L223)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

please revert this newline (just to avoid having traces here while you didn't actually change anything here).
</blockquote></details>

Done.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 538 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGywNBO6HvJc12GMIe2:-OGzCGVX5FJYJnSR7nOQ:b9vu48y) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L538)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

Sorry for nagging, but can you please doc the effects of disable_trace_padding in its definition? I'd like to avoid the case of proof_mode in which no one knows what exactly it does and thus are worried changing/touching it.
</blockquote></details>

it's defined in the config of the runner, then passed to the vm. In which struct would you like it documented (or both)? Also, not sure I get you remark regarding proof_mode.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 545 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGyXe2A6g6milvlF5Qh:-OGzCfuv8jNAbTWEur3x:bd3ecap) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L545)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

assert that padded_used_cells >= used_cells?
Or maybe even compute the used_cells as used_instances * cells_per_instance, and assert it is equal to used_cells?
The computation seems messy and I am concerned it may have inconsistencies.
Or maybe just compute it that way, and convert the assert with a test (asserting both ways of computation yield the same numbers, for all the builtins).
</blockquote></details>

the `>= ` check is done implicitly. `get_used_instances` just does `div_ceil(used_cells, cells_per_instance())`. Also, the second remark isn't always true for the last instance (some outputs might not be filled).
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 935 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGz7YSb0eWEkE-BCALj:-OGzDz3jCmO6n3HIYva4:bee4zre) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L935)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

this assumes they are in the same order. Makes sense, but consider documenting it.
</blockquote></details>

Yeah, it could be assumed, but you're right that it's much clearer if state explicitly. I added an assert at the start of the loop that that's the case.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 945 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGz6fml1vY9a3SMMdOC:-OGzEMwVDpWGazGX8n3S:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L945)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

any reason this should be Err? If not, just unwrap should do better.
</blockquote></details>

Done.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 966 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGz7vPR9UJvvMGEZuBI:-OGzETdA8ZwZQXs_hH_f:bg4p3i1) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L966)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

unclear. What does it check?
</blockquote></details>

That there was indeed a different outcome in the number of allocated instances (basically, that `true` didn't pad according to the ratio)

yuvalsw (changes requested):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 5 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @pefontana, and @YairVaknin-starkware)

___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 538 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGywNBO6HvJc12GMIe2:-OH26kZQEXz9W1GFmxjq:b-ibjn7m) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L538)):*
<details><summary><i>Previously, YairVaknin-starkware wrote…</i></summary><blockquote>

it's defined in the config of the runner, then passed to the vm. In which struct would you like it documented (or both)? Also, not sure I get you remark regarding proof_mode.
</blockquote></details>

Both. I'd write it in the outer one (runner), and in the vm just refer to it for more details.

re proof_mode - it has many effects, it got to the state where it's hard to tell exactly what they are. I don't want disable_trace_padding to get to the same state.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 967 at r7](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OH29YF12HN0WIFGlJKx:-OH29YF12HN0WIFGlJKy:bwy079p) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/a8a42d8f219786a32eb8d56838752ec3eeee053d/vm/src/vm/runners/builtin_runner/mod.rs#L967)):*
> ```XML
>             // Checks that the number of allocated instances is diffrent when trace padding is
>             // disabled. True for this specific program, not always. That is, instance could could
>             // happen to be the same.
> ```

clearer than before. My suggestion for more clarity:



_Suggestion:_
```XML
            // Checks that the number of allocated instances is diffrent when trace padding is
            // enabled/disabled. Holds for this specific program, not always (that is, in other programs, padding may be of size 0).
```

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 5 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @pefontana, and @yuvalsw)

___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 538 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGywNBO6HvJc12GMIe2:-OH2q0DNB6UcXZjPD5om:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L538)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

Both. I'd write it in the outer one (runner), and in the vm just refer to it for more details.

re proof_mode - it has many effects, it got to the state where it's hard to tell exactly what they are. I don't want disable_trace_padding to get to the same state.
</blockquote></details>

Done.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 967 at r7](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OH29YF12HN0WIFGlJKx:-OH2q0mB7HO__KKcIG8U:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/a8a42d8f219786a32eb8d56838752ec3eeee053d/vm/src/vm/runners/builtin_runner/mod.rs#L967)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

clearer than before. My suggestion for more clarity:
</blockquote></details>

Done.

CHANGELOG.md
YairVaknin-starkware: Done.

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: 10 of 14 files reviewed, 5 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @yuvalsw)

yuvalsw:
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 545 at r6](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OGyXe2A6g6milvlF5Qh:-OH3-iFO5Cru9l-iolh3:b-s9w6q1) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/23e63122a647d6f99d52ab8900565eee56e41a79/vm/src/vm/runners/builtin_runner/mod.rs#L545)):*
<details><summary><i>Previously, YairVaknin-starkware wrote…</i></summary><blockquote>

the `>= ` check is done implicitly. `get_used_instances` just does `div_ceil(used_cells, cells_per_instance())`. Also, the second remark isn't always true for the last instance (some outputs might not be filled).
</blockquote></details>

Then, as discussed, maybe assert (in the test) that (something like) `get_used_instances * cells_per_instance - get_used_cells < cells_per_instance`.

yuvalsw (commented):
> Reviewed 3 of 3 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 4 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @YairVaknin-starkware)

___
*[`vm/src/cairo_run.rs` line 38 at r9](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OH3ATvQ7KVr6bOKyRFZ:-OH3ATvR7AUxuwg8fQ9g:b-amq99l) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/29d1d44840f1943fb9176f6d614a76401676caa5/vm/src/cairo_run.rs#L38)):*
> ```XML
>     pub secure_run: Option<bool>,
>     // disable padding of the trace to accommodate the expected ratios/n_steps relationships w.r.t
>     // the layout.
> ```

Please mention how. It doesn't pad steps at all, but does pad segments to a power of 2...



_Suggestion:_
```XML
    // Disable padding of the trace to accommodate the expected ratios/n_steps relationships w.r.t
    // the layout.
```

yuvalsw (changes requested):
> Reviewed 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 4 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @YairVaknin-starkware)

___
*[`vm/src/cairo_run.rs` line 39 at r10](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OH3fVdE6C0nm38uL4qt:-OH3fVdE6C0nm38uL4qu:b7vmrot) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/c9b8416266984da8fc91dcc0c9fe3f732ee25b49/vm/src/cairo_run.rs#L39)):*
> ```XML
>     // Disable padding of the trace to accommodate the expected ratios-n_steps relationships w.r.t
>     // the layout. It does, however, pad the number of builtin instances to the next power of 2
>     // compared to their values at the end of the execution, but doesn't modify n_steps.
> ```

It confuses with double negations. My suggestion:



_Suggestion:_
```XML
    /// Disable padding of the trace.
    /// By default, the trace is padded to accommodate the expected builtins-n_steps relationships according to
    /// the layout.
    /// When the padding is disabled:
    /// - It doesn't modify/pad n_steps.
    /// - It still pads each builtin segment to the next power of 2 (counting instances of the builtin) compared to their sizes at the end of the execution.
```
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 543 at r10](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OH3hJpz2gdLQ22zKkHy:-OH3hJpz2gdLQ22zKkHz:bfqnc6i) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/c9b8416266984da8fc91dcc0c9fe3f732ee25b49/vm/src/vm/runners/builtin_runner/mod.rs#L543)):*
> ```XML
>                         let n_output_cells =
>                             self.cells_per_instance() as usize - self.n_input_cells() as usize;
>                         assert!(
> ```

1. Please doc what you explained to me f2f - about why it is not exactly equal.
2. why not assert in a test?
3. can the >= become >? Shouldn't there be at least one full output cell ? (not super important of course)

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 4 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @yuvalsw)

___
*[`vm/src/cairo_run.rs` line 39 at r10](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OH3fVdE6C0nm38uL4qt:-OH8rinT0XRP6Ns0ZKoE:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/c9b8416266984da8fc91dcc0c9fe3f732ee25b49/vm/src/cairo_run.rs#L39)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

It confuses with double negations. My suggestion:
</blockquote></details>

Done.
___
*[`vm/src/vm/runners/builtin_runner/mod.rs` line 543 at r10](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909#-OH3hJpz2gdLQ22zKkHy:-OH8uPh50EGa78Be3QfP:bxhz46h) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/c9b8416266984da8fc91dcc0c9fe3f732ee25b49/vm/src/vm/runners/builtin_runner/mod.rs#L543)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

1. Please doc what you explained to me f2f - about why it is not exactly equal.
2. why not assert in a test?
3. can the >= become >? Shouldn't there be at least one full output cell ? (not super important of course)
</blockquote></details>

1. Done.
2. thought it couldn't hurt, but moved it to the test and made it more robust (more programs/more used builtins).
3 . no, for range_check there's no input/output distinction.

yuvalsw (approved):
> Reviewed 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1909)* status: all files reviewed, 3 unresolved discussions (waiting on @fmoletta, @FrancoGiachetta, @gabrielbosio, @igaray, @juanbono, @OmriEshhar1, @Oppen, @pefontana, and @YairVaknin-starkware)

vm/src/cairo_run.rs
```diff
@@ -34,6 +34,13 @@ pub struct CairoRunConfig<'a> {
     pub dynamic_layout_params: Option<CairoLayoutParams>,
     pub proof_mode: bool,
     pub secure_run: Option<bool>,
+    /// Disable padding of the trace.
+    /// By default, the trace is padded to accommodate the expected builtins-n_steps relationships
+    /// according to the layout.
+    /// When the padding is disabled:
+    /// - It doesn't modify/pad n_steps.
+    /// - It still pads each builtin segment to the next power of 2 w.r.t the number of used
+    ///   instances of the builtin) compared to their sizes at the end of the execution.
```
gabrielbosio, line 43: In line 43 there's only a closing parenthesis but not the opening one

gabrielbosio (approved):
Looks good, I've left a small comment

vm/src/cairo_run.rs
YairVaknin-starkware: Fixed.

PR #1945, opened 2025-02-11
Check overflow in cairo pie address calculation

JulianGCalderon:
# Check overflow in cairo pie address calculation

## Description

Uses `checked_add` and `checked_mul` to return explicit errors rather than silently failing in case of an overflow.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/vm/runners/cairo_pie.rs
+++ b/vm/src/vm/runners/cairo_pie.rs
@@ -528,7 +528,9 @@ pub(super) mod serde_impl {
                 .and_then(|n| n.checked_add(ADDR_BASE))
                 .and_then(|n| n.checked_add(*offset as u64))
                 .ok_or_else(|| {
-                    serde::ser::Error::custom("final memory address calculation overflowed")
+                    serde::ser::Error::custom(format!(
+                        "failed to serialize address: {segment}:{offset}"
+                    ))
                 })?;
 
             res.extend_from_slice(mem_addr.to_le_bytes().as_ref());
--- a/vm/src/vm/runners/cairo_pie.rs
+++ b/vm/src/vm/runners/cairo_pie.rs
@@ -858,6 +858,17 @@ mod test {
         );
     }
 
+    #[test]
+    fn serialize_cairo_pie_memory_with_overflow() {
+        let memory = CairoPieMemory(vec![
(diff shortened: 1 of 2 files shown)


vm/src/vm/runners/cairo_pie.rs
```diff
@@ -522,7 +522,15 @@ pub(super) mod serde_impl {
...
 
         for ((segment, offset), value) in values.iter() {
-            let mem_addr = ADDR_BASE + *segment as u64 * OFFSET_BASE + *offset as u64;
+            // mem_addr = ADDR_BASE + segment * OFFSET_BASE + offset
+            let mem_addr = (*segment as u64)
+                .checked_mul(OFFSET_BASE)
+                .and_then(|n| n.checked_add(ADDR_BASE))
+                .and_then(|n| n.checked_add(*offset as u64))
+                .ok_or_else(|| {
+                    serde::ser::Error::custom("final memory address calculation overflowed")
```
ilyalesokhin-starkware, line 531: I think the error should be something like:

Address serialization failed: {segment}:{offset}

JulianGCalderon: Done

vm/src/vm/runners/cairo_pie.rs
```diff
@@ -522,7 +522,17 @@ pub(super) mod serde_impl {
...
+            // mem_addr = ADDR_BASE + segment * OFFSET_BASE + offset
+            let mem_addr = (*segment as u64)
+                .checked_mul(OFFSET_BASE)
+                .and_then(|n| n.checked_add(ADDR_BASE))
+                .and_then(|n| n.checked_add(*offset as u64))
+                .ok_or_else(|| {
+                    serde::ser::Error::custom(format!(
+                        "failed to serialize address: {segment}:{offset}"
+                    ))
+                })?;
```
FrancoGiachetta, line 534: The error can come up from any of three different operations, so it would be nice if we could give a hint of which operation failed. If there's no way to do this, I would the `and_then` methods into variables. This way as least we don't centralize the error in one source.

FrancoGiachetta (changes requested):

gabrielbosio (approved):

PR #1927, opened 2025-01-26
implement Blake2s opcode in runner

ohad-nir-starkware:
# Blake2s opcode runner
## Description
Adding the opcode Blake2s to the VM.
Expects op0 to be a pointer to a sequence of 8 felts and and op1 to be a pointer to a sequence of 16 felts.
Said felts should represent u32 integers, i.e. have value of at most 2**32-1.
The 8 felts of op0 represent a state and the 16 felts of op1 represent a message.
dst should hold the value of the counter.
The "output" consists of 8 felts representing u32 numbers of the output of the (non last block) Blake2s compression.
ap should store a pointer, it points to a sequence of 8 cells which each should either be uninitialised or already contain a value matching that of the output at the same index.
The opcode inserts the aforementioned output into the 8 cells [[ap]], [[ap]+1], ... [[ap]+7] (and yields an error if one of said cells already contains a value differing from the output).

Currently Blake2s has opcode_num 8, meaning encoded_instr can use all 64 bits and InstructionNonZeroHighBits is no longer needed as an error.

The motivation is that it has been decided at Starkware that Blake2s is to be implemented at an opcode, thus it needs to be supported by the runner.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added
- [x] Integration tests added.
- [x] This change requires new documentation.
  - [x] Documentation has been added/updated.
  - [x] CHANGELOG has been updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/1927)

PR #1994, opened 2025-03-10
Added hints felt unpacking for blake.

Alon-Ti:


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 #### [2.0.0] - 2025-02-26
 
+* feat: Support hints for new blake felt serialization library code [#1994](https://github.com/lambdaclass/cairo-vm/pull/1994)
```
FrancoGiachetta, line 7: This is an upcoming feature, it should go under the `Upcoming Changes` label

FrancoGiachetta (changes requested):

FrancoGiachetta:
Hi there! Could you give some context of what these new hints do?

Alon-Ti:
> Hi there! Could you give some context of what these new hints do?

The new blake2s opcode (see #1927 etc) requires that we split felts we would like to hash into 32-bit limbs. These are hints for cairo library code that takes an array of felts and returns an array of 32-bit limbs, with the optimization that felts under 2^63 are encoded in two limbs.

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
```diff
@@ -242,6 +245,87 @@ pub fn blake2s_add_uint256_bigend(
...
+        vm.insert_value(ap, 0)?
+    } else {
+        let val = vm.get_integer(packed_values)?.into_owned();
+        vm.insert_value(
+            ap,
+            if val.to_biguint() < (BigUint::from(1_u32) << 63) {
+                1
+            } else {
+                0
+            },
```
FrancoGiachetta, line 270: This could be simplified as:
```rust
vm.insert_value(ap, (val.to_biguint() < (BigUint::from(1_u32) << 63)) as usize)?
```

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
```diff
@@ -242,6 +245,87 @@ pub fn blake2s_add_uint256_bigend(
...
+        .into_iter()
+        .map(|val| val.into_owned().to_biguint())
+        .flat_map(|val| {
+            if val < pow2(63) {
+                let (high, low) = val.div_rem(&(pow2(32)));
+                vec![high, low]
+            } else {
+                let mut limbs: [BigUint; 8] = Default::default();
+                let mut val: BigUint = val + (pow2(255));
+                for i in (0..8).rev() {
```
FrancoGiachetta, line 312: we could do `8..0` in this case, right? Avoiding having to reverse the range.

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
```diff
@@ -242,6 +245,87 @@ pub fn blake2s_add_uint256_bigend(
...
+    let packed_values = get_ptr_from_var_name("packed_values", vm, ids_data, ap_tracking)?;
+    let unpacked_u32s = get_ptr_from_var_name("unpacked_u32s", vm, ids_data, ap_tracking)?;
+
+    let vals = vm.get_integer_range(packed_values, felt_to_usize(&packed_values_len)?)?;
+    let pow2 = |exp: u32| BigUint::from(1_u32) << exp;
+
+    // Split value into either 2 or 8 32-bit limbs.
+    let out: Vec<MaybeRelocatable> = vals
+        .into_iter()
+        .map(|val| val.into_owned().to_biguint())
```
FrancoGiachetta, line 304: I think there's no need to call `into_owned()` in this case, since the `into_iter()` crates an iterator of already owned values.

FrancoGiachetta (changes requested):

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
```diff
@@ -242,6 +245,87 @@ pub fn blake2s_add_uint256_bigend(
...
+        .flat_map(|val| {
+            if val < pow2(63) {
+                let (high, low) = val.div_rem(&(pow2(32)));
+                vec![high, low]
+            } else {
+                let mut limbs: [BigUint; 8] = Default::default();
```
JulianGCalderon, line 310: We could use `Vec::with_capacity` directly instead of declaring and array and transforming it into a vector.

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
Alon-Ti: That yields an empty range.

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
Alon-Ti: `with_capacity` doesn't update the vector length so you can't access it at 7, changed to `vec![;8]` to avoid the cast.

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
FrancoGiachetta: You are right, my bad. You can dismiss this comment.

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
```diff
@@ -242,6 +245,83 @@ pub fn blake2s_add_uint256_bigend(
     Ok(())
 }
 
+/* Implements Hint:
+memory[ap] = (ids.end != ids.packed_values) and (memory[ids.packed_values] < 2**63)
+*/
+pub fn blake2s_unpack_nondet_jmp(
```
JulianGCalderon, line 251: I have some questions about this:
- What is this hint used for?
- What is the rationale behind the name of this function?

Maybe we can rename it to something more expressive.

Alon-Ti: It allows the program to nondeterministically jump to the branch where felts are over 63 bits, which is also the branch that exists the program. Changed the names to reflect the potentially more general nature of this.

JulianGCalderon (approved):
LGTM

FrancoGiachetta (approved):

vm/src/hint_processor/builtin_hint_processor/blake2s_utils.rs
```diff
@@ -604,6 +684,106 @@ mod tests {
             .is_none());
     }
 
+    #[test]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
+    fn blake2s_unpack_nondet_jmp_ends() {
```
JulianGCalderon, line 689: Could you rename the tests to comply with the new name `is_over_63_bits_or_end`?

vm/src/hint_processor/builtin_hint_processor/builtin_hint_processor_definition.rs
```diff
@@ -360,6 +360,12 @@ impl HintProcessorLogic for BuiltinHintProcessor {
             hint_code::BLAKE2S_ADD_UINT256_BIGEND => {
                 blake2s_add_uint256_bigend(vm, &hint_data.ids_data, &hint_data.ap_tracking)
             }
+            hint_code::IS_OVER_63_BITS_OR_END => {
```
gabrielbosio, line 363: If the suggestion for [this line](https://github.com/lambdaclass/cairo-vm/pull/1994/files#diff-a183e52558b64ce0463e14e5b3caa24d52b9de0dda34a58d46830bedafd4e572R251) is accepted then the enum variant name should also change to match the updated hint name.

PR #1995, opened 2025-03-11
Migrate from pyenv to uv

edg-l:
This migrates the python workflow from `pyenv` to the fast modern rust based `uv`.

The only takeaway is that the `pypy` versions you can install from `uv` doesn't seem to work well with the `bitarray` dependency we have, as it fails to compile. As `pypy` was only used in `run_benchmarks.sh`, which was an outdated script that we no longer used, I also removed it.

I also tried to update python to 3.13.2 but `cairo-run` doesnt work well, however `cairo-compile` does work well. In the end I kept the old python version 3.9.


edg-l:
The non-required ci workflow [Benchmark Hyper Threading / benchmark (pull_request)](https://github.com/lambdaclass/cairo-vm/actions/runs/13785759833/job/38553309589?pr=1995) fails because main doesnt use uv yet

.github/workflows/fresh_run.yml
```diff
@@ -40,16 +40,18 @@ jobs:
...
+        enable-cache: true
 
     - name: Install Rosetta
       if: matrix.os_name == 'macos'
       run: softwareupdate --install-rosetta --agree-to-license
 
     - name: Run make deps${{ matrix.deps_suffix }}
       id: deps
-      run: make deps${{ matrix.deps_suffix }}
+      run: make deps${{ matrix.deps_suffix }} && . cairo-vm-env/bin/activate && echo PATH=$PATH >> $GITHUB_ENV
```
FrancoGiachetta, line 54: what's the `echo PATH...` used for in this case?

edg-l: to keep the python venv on the next steps

FrancoGiachetta: oh I see, thanks!

FrancoGiachetta (approved):

bench/run_benchmarks.sh
```diff
@@ -9,7 +9,6 @@ for file in $(ls $tests_path | grep .cairo | sed -E 's/\.cairo//'); do
     export PATH="$(pyenv root)/shims:$PATH"
 
     hyperfine \
-	    -n "Cairo VM (CPython)" "PYENV_VERSION=3.9.15 cairo-run --proof_mode --memory_file /dev/null --trace_file /dev/null --layout starknet_with_keccak --program $tests_path/$file.json" \
-	    -n "Cairo VM (PyPy)" "PYENV_VERSION=pypy3.9-7.3.9 cairo-run --proof_mode --memory_file /dev/null --trace_file /dev/null --layout starknet_with_keccak --program $tests_path/$file.json" \
+	    -n "Cairo VM (CPython)" "cairo-run --proof_mode --memory_file /dev/null --trace_file /dev/null --layout starknet_with_keccak --program $tests_path/$file.json" \
```
JulianGCalderon, line 12: `uv` uses CPython by default?

JulianGCalderon: Is there a way to keep the pypy bench?

.github/workflows/hyperfine.yml
```diff
@@ -35,15 +35,19 @@ jobs:
...
+          enable-cache: true
+      - name: Install python deps
         if: ${{ steps.cache.outputs.cache-hit != 'true' }}
-        run: pip install -r requirements.txt
+        run: |
+          uv python install 3.9.15
+          uv venv --python 3.9.15 cairo-vm-env
+          . cairo-vm-env/bin/activate
+          uv pip install -r requirements.txt
+          echo PATH=$PATH >> $GITHUB_ENV
```
JulianGCalderon, line 50: Can we reuse `ci-python-deps` here?

.github/workflows/hyper_threading_benchmarks.yml
```diff
@@ -14,14 +14,19 @@ jobs:
...
+      uses: astral-sh/setup-uv@v5
       with:
-        python-version: '3.9'
+        enable-cache: true
+    - name: Install python deps
+      run: |
+          uv python install 3.9.15
+          uv venv --python 3.9.15 cairo-vm-env
+          . cairo-vm-env/bin/activate
+          uv pip install -r requirements.txt
```
JulianGCalderon, line 26: Can we reuse `ci-python-deps` here?

.github/workflows/bench.yml
```diff
@@ -19,12 +19,17 @@ jobs:
...
-        python-version: '3.9'
-    - name: Install test dependencies
-      run: pip install -r requirements.txt
+        enable-cache: true
+    - name: Install python deps
+      run: |
+          uv python install 3.9.15
+          uv venv --python 3.9.15 cairo-vm-env
+          . cairo-vm-env/bin/activate
+          uv pip install -r requirements.txt
```
JulianGCalderon, line 31: Can we reuse `ci-python-deps` here?

.github/workflows/hyper_threading_benchmarks.yml
JulianGCalderon: Also, we could update the path here to avoid activating the env later, right?

bench/run_benchmarks.sh
edg-l: Read the pr description on why pypy is removed., cpython is python

.github/workflows/hyper_threading_benchmarks.yml
edg-l: no because of different checkouts of head and main its better to not rely on the makefile

.github/workflows/hyperfine.yml
edg-l: no because of different checkouts of head and main its better to not rely on the makefile

bench/run_benchmarks.sh
JulianGCalderon: Oh my bad

Makefile
```diff
@@ -227,25 +227,21 @@ cargo-deps:
...
-	PYENV_VERSION=pypy3.9-7.3.9 python -m venv cairo-vm-pypy-env
-	. cairo-vm-pypy-env/bin/activate ; \
-	pip install -r requirements.txt ; \
-	pyenv install -s 3.9.15
-	PYENV_VERSION=3.9.15 python -m venv cairo-vm-env
+deps: create-proof-programs-symlinks cargo-deps build-cairo-1-compiler build-cairo-2-compiler cairo1-run-deps python-deps ;
+
+python-deps:
+	uv python install 3.9.15 ; \
+	uv venv --python 3.9.15 cairo-vm-env
```
gabrielbosio, line 234: This implies having installed `uv` to run `make deps`. This means that:

- `uv` is not an optional dependency
- `make deps` does not serve as an installation script

I suggest to add `uv` installation in the `make deps` command and set `uv` as a required dependency.

JulianGCalderon: If we set `uv` as a required dependency, we should also set `make` as a required dependency. Right know there is a distinction between required dependencies (for running only the CairoVM), and optional/development dependencies (for running tests, compiling programs, etc). `uv` is a development dependency.

We can already install `uv` by running the installation script: `install.sh`. I think `make deps` never was an instalation script (it didn't install pyenv before).

gabrielbosio: Ok, makes sense

gabrielbosio (approved):

PR #2005, opened 2025-03-14
Fix comment in `initialize_builtins()`

FrancoGiachetta:
# TITLE

## Description

This PR fixes the comment `initialize_builtins` which is not matching the function's actual behavior.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


JulianGCalderon (approved):

JulianGCalderon (changes requested):
I'm not sure there is an error:
When running in proof_mode:
- all builtins in the layout are created: this is true
- only those in the program will be included: this is also true, se that the variable `included` is given to the creation of each builtin.

FrancoGiachetta:
I wouldn't say the second point is always true. Take this snippet of the function as an example:
``` rust
if let Some(instance_def) = self.layout.builtins.pedersen.as_ref() {
            let included = program_builtins.remove(&BuiltinName::pedersen);
            if included || self.is_proof_mode() {
                self.vm
                    .builtin_runners
                    .push(HashBuiltinRunner::new(instance_def.ratio, included).into());
            }
        }
```
and this cairo code:

```cairo
%builtins output range_check bitwise keccak poseidon range_check96 add_mod mul_mod

from starkware.cairo.common.cairo_builtins import (
    BitwiseBuiltin,
    KeccakBuiltin,
    PoseidonBuiltin,
    ModBuiltin,
    HashBuiltin,
    SignatureBuiltin,
    EcOpBuiltin,
)
func main{
    output_ptr: felt*,
    range_check_ptr,
    bitwise_ptr: BitwiseBuiltin*,
    keccak_ptr: KeccakBuiltin*,
    poseidon_ptr: PoseidonBuiltin*,
    range_check96_ptr: felt*,
    add_mod_ptr: ModBuiltin*,
    mul_mod_ptr: ModBuiltin*,
}() {
}
```

Running this with `proof_mode`, the pedersen builtin would get intiliazed (which should be) and also would be included in the builtin_runners list (which should not according to the previous comment).
proof_mode will get every builtin in the layout initialized and included in the builtin_runners even if you haven't declared any of those in the program. This is the expected behavior, but the comment is not describing it correctly.

JulianGCalderon:
Regarding the snippet:
```rust
if let Some(instance_def) = self.layout.builtins.pedersen.as_ref() {
    let included = program_builtins.remove(&BuiltinName::pedersen);
    if included || self.is_proof_mode() {
        self.vm
        .builtin_runners
        .push(HashBuiltinRunner::new(instance_def.ratio, included).into());
    }
}
```

You could add some debug prints just in case, but the runner is added to the `builtin_runners` list, but with `included = false`. Calling `runner.included()` should return `false`.

FrancoGiachetta:
I agree with you, it will be `false`. El thing is this condition `included || self.is_proof_mode()` if `proof_mode` is enabled, it does not matter if the builtin is included or not. In other words, it does not matter if have declared it in your cairo program. It wil be included in the builtin_runner. However, the comment says that if it is not included (which means you haven't explicitly declared it in your program) it won't produce any effect, which is not actually what's happening.

FrancoGiachetta:
So perhaps we are having two different understandings of the comment.

JulianGCalderon (approved):

PR #1976, opened 2025-02-26
fix: handle unfilled values in mul mod runner

enitrat:
# Fix handling of unfilled values in mul mod runner

## Description
Closes #1975
Description of the pull request changes and motivation.

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added
- [ ] Integration tests added.
- [ ] This change requires new documentation.
  - [ ] Documentation has been added/updated.
  - [x] CHANGELOG has been updated.

PR #2012, opened 2025-03-17
[DNM] Fix/mulmod unfilled value

gabrielbosio:
Force the CI in #1976 to run so that we make sure the jobs don't fail because of issues with permissions.


JulianGCalderon:
Closing this as the related PR has already merged (and the CI passed correctly)

PR #2018, opened 2025-03-19
Removed CODEOWNERS file

igaray:
## Description
Removes the CODEOWNERS file as the requirement for codeowner reviews for merging has been temporarily lifted.


FrancoGiachetta (approved):

FrancoGiachetta:
The `Run Lints` job seems to be failing due to a program not being found. Perhaps we need to delete the cache and run the jobs again

gabrielbosio (approved):

PR #2029, opened 2025-03-26
[DNM] Always use a normal segment for first SegmentArena segment

gabrielbosio:
Clone of #1845


gabrielbosio:
Closing to see if by opening another PR we can trigger the failing jobs correctly.

PR #2043, opened 2025-04-01
fix: add VM Execution State

crStiv:
# Add VM Execution State

## Description

Added ExecutionState structure and get_execution_state() method to capture current VM state. Helps with debugging and simplifies integration with external tools.
What's included:
- Structure with key VM parameters (pc, ap, fp, current_step, etc.)
- Tests to verify functionality
- Documentation

## Checklist
-  [x] Linked to Github Issue
- [x] Unit tests added
- [x] Integration tests added
- [x] Documentation updated

### DIFF
--- a/vm/src/vm/vm_core.rs
+++ b/vm/src/vm/vm_core.rs
@@ -5434,6 +5434,7 @@ mod tests {
         );
     }
 
+    #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn get_execution_state_test() {
         let mut vm = vm!();
--- a/vm/src/vm/vm_core.rs
+++ b/vm/src/vm/vm_core.rs
@@ -103,6 +103,25 @@ pub struct VirtualMachine {
     pub(crate) relocation_table: Option<Vec<usize>>,
 }
 
+/// Структура для хранения состояния выполнения виртуальной машины
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExecutionState {
+    /// Текущий счетчик программы (Program Counter)
+    pub pc: Relocatable,
+    /// Текущий указатель распределения (Allocation Pointer)
+    pub ap: usize,
+    /// Текущий указатель кадра (Frame Pointer)
+    pub fp: usize,
+    /// Текущий шаг выполнения
+    pub current_step: usize,
+    /// Завершено ли выполнение
+    pub run_finished: bool,
+    /// Количество сегментов памяти
+    pub memory_segments_count: usize,
+    /// Информация о работающих builtins
+    pub active_builtins: Vec<String>,
+}
+
 impl VirtualMachine {
     pub fn new(trace_enabled: bool, disable_trace_padding: bool) -> VirtualMachine {
         let run_context = RunContext {
@@ -1247,6 +1266,30 @@ impl VirtualMachine {
         for info in segment_infos {
             self.segments
                 .finalize(Some(info.size), info.index as usize, None)
+        }
+    }
+
+    /// Возвращает текущее состояние выполнения виртуальной машины
+    /// 
+    /// Эта функция собирает все важные компоненты состояния VM, включая
+    /// текущие регистры (pc, ap, fp), информацию о шаге выполнения и 
+    /// активных builtin функциях.
+    /// 
+    /// # Возвращаемое значение
+    /// 
+    /// `ExecutionState` структура с текущим состоянием VM
+    pub fn get_execution_state(&self) -> ExecutionState {
+        ExecutionState {
+            pc: self.run_context.pc,
+            ap: self.run_context.ap,
+            fp: self.run_context.fp,
+            current_step: self.current_step,
+            run_finished: self.run_finished,
+            memory_segments_count: self.segments.num_segments(),
+            active_builtins: self.builtin_runners
+                .iter()
+                .map(|runner| runner.name().to_string())
+                .collect(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


vm/src/vm/vm_core.rs
```diff
@@ -103,6 +103,25 @@ pub struct VirtualMachine {
...
+    /// Текущий указатель кадра (Frame Pointer)
+    pub fp: usize,
+    /// Текущий шаг выполнения
+    pub current_step: usize,
+    /// Завершено ли выполнение
+    pub run_finished: bool,
+    /// Количество сегментов памяти
+    pub memory_segments_count: usize,
+    /// Информация о работающих builtins
+    pub active_builtins: Vec<String>,
```
FrancoGiachetta, line 122: Why are those comments not written in english?

vm/src/vm/vm_core.rs
```diff
@@ -1249,6 +1268,30 @@ impl VirtualMachine {
...
+
+    /// Возвращает текущее состояние выполнения виртуальной машины
+    /// 
+    /// Эта функция собирает все важные компоненты состояния VM, включая
+    /// текущие регистры (pc, ap, fp), информацию о шаге выполнения и 
+    /// активных builtin функциях.
+    /// 
+    /// # Возвращаемое значение
+    /// 
+    /// `ExecutionState` структура с текущим состоянием VM
```
FrancoGiachetta, line 1280: The same here

FrancoGiachetta: great!

vm/src/vm/vm_core.rs
```diff
@@ -5390,4 +5433,38 @@ mod tests {
...
+
+    /// Returns the current execution state of the virtual machine
+    /// 
+    /// This function gathers all essential components of the VM state, including  
+    /// the current registers (pc, ap, fp), execution step information, and
+    /// active built-in functions.
+    /// 
+    /// # Return Value
+    /// 
+    /// `ExecutionState` is a structure containing the current state of the VM
```
FrancoGiachetta, line 5445: I think this comment can be removed since it already is the method's firm, and this is actually a test

FrancoGiachetta (dismissed):

FrancoGiachetta:
Hey @crStiv, could you dismiss my approve. I did it by accident :expressionless:

JulianGCalderon:
Hi @crStiv, could you share a bit more information on the usecases of this new structure? Why are those particular fields necessary?

FrancoGiachetta:
Is it possible for you to tell us what you are doing that needs this information at some point in the program?

FrancoGiachetta:
Hi @crStiv! We've taken a look at this. You can already almost all the information (except for with `run_finished` attribute, since it's private). Here is an example of how you could create the same struct in you project:

```rust 
pub struct ExecutionState {
    /// Current program counter (Program Counter)
    pub pc: Relocatable,
    /// Current allocation pointer (Allocation Pointer)
    pub ap: usize,
    /// Current frame pointer (Frame Pointer)
    pub fp: usize,
    /// Current execution step
    pub current_step: usize,
    /// Number of memory segments
    pub memory_segments_count: usize,
    /// Information about active built-ins
    pub active_builtins: Vec<String>,
}

...

let state = ExecutionState {
    pc: vm.get_pc(),
    ap: vm.get_ap().offset,
    fp: vm.get_fp().offset,
    current_step: vm.get_current_step(),
    memory_segments_count: vm.segments.num_segments(),
    active_builtins: vm.get_builtin_runners()
        .iter()
        .map(|runner| runner.name().to_string())
        .collect(),
}
```
The only attribute you cannot retrieve from the vm would be run_finished. This is a boolean which is used to prevent marking a range memory cells as accessed, which can only be done if the execution has finished. Is it possible to know why do you need to read that attribute?

FrancoGiachetta:
Hi @crStiv! We've taken a look at this. You can already almost all the information (except for with `run_finished` attribute, since it's private). Here is an example of how you could create the same struct in you project:

```rust 
pub struct ExecutionState {
    /// Current program counter (Program Counter)
    pub pc: Relocatable,
    /// Current allocation pointer (Allocation Pointer)
    pub ap: usize,
    /// Current frame pointer (Frame Pointer)
    pub fp: usize,
    /// Current execution step
    pub current_step: usize,
    /// Number of memory segments
    pub memory_segments_count: usize,
    /// Information about active built-ins
    pub active_builtins: Vec<String>,
}

...

let state = ExecutionState {
    pc: vm.get_pc(),
    ap: vm.get_ap().offset,
    fp: vm.get_fp().offset,
    current_step: vm.get_current_step(),
    memory_segments_count: vm.segments.num_segments(),
    active_builtins: vm.get_builtin_runners()
        .iter()
        .map(|runner| runner.name().to_string())
        .collect(),
}
```
The only attribute you cannot retrieve from the vm would be run_finished. This is a boolean which is used to prevent marking a range memory cells as accessed, which can only be done if the execution has finished. Is it possible to know why do you need to read that attribute?

FrancoGiachetta:
The thing is that this method would be public, and thus it would increase the size of the api exposed. Any change to this method or the struct would be a breaking. Since you can already get almost all of these attributes by your self, it would be appropriate. Talking about `run_finished`, this value is set when you run a program. I'm not getting what is you use case here. Is it, by any change, possible for to tell?

gabrielbosio:
Hi @crStiv, for further details on how to contribute to the Cairo VM, see [CONTRIBUTING.md](https://github.com/lambdaclass/cairo-vm/blob/main/CONTRIBUTING.md).

PR #2056, opened 2025-04-07
CI: Make cache keys depend on Cairo version

JulianGCalderon:
# CI: Make cache keys depend on Cairo version

The CI sometimes breaks when:
- New Cairo programs are added
- We change te Cairo compiler
- We modify the Makefile

This PR changes the CI so that any changes to `[Makefile, requirements.txt, *.cairo]` invalidate the cache. This should fix the issue.


empty
FrancoGiachetta, line 1: You might forgotten to remove this file

FrancoGiachetta (changes requested):

empty
JulianGCalderon: true

JulianGCalderon: Done.

FrancoGiachetta (approved):

gabrielbosio (approved):

PR #2075, opened 2025-04-25
chore(ci): bump actions/cache to v4

PixelPil0t1:
Updated actions/cache from v3 → v4 

[GitHub Actions v4 release](https://github.com/actions/cache/releases/tag/v4)


gabrielbosio:
Thanks but we don't accept this kind of PRs

PR #2078, opened 2025-04-28
Fix WRITE_DIVMOD_SEGMENT hint

JulianGCalderon:
# Fix WRITE_DIVMOD_SEGMENT hint

We were using `SECP_P` prime instead of `PRIME`, as indicated by the hint code.

Additionally, I refactored the hint logic to:
- Return an error variant instead of an assert.
- Improve legibility.


gabrielbosio (approved):

vm/src/hint_processor/builtin_hint_processor/kzg_da/mod.rs
```diff
@@ -74,35 +72,36 @@ pub fn write_div_mod_segment(
...
 
-fn as_int(value: BigInt, prime: &BigInt) -> BigInt {
-    let half_prime = prime / 2u32;
-    if value > half_prime {
-        value - prime
-    } else {
+fn as_int(value: BigInt, prime: &BigUint) -> BigInt {
+    let half_prime: BigInt = (prime / 2u32).into();
+    let prime: BigInt = prime.clone().into();
+    if value < half_prime {
```
FrancoGiachetta, line 94: Shouldn't it be `value > half_prime`? Negative numbers are bigger than `half_prime`, right?

JulianGCalderon: If `value < half_prime`, that means `value` is positive, therefore we return `value`.

FrancoGiachetta: Oh right, I misread that part.

FrancoGiachetta (approved):

PR #2086, opened 2025-05-04
Stav/remove test compare to python vm

Stavbe:
Initially, we planned to fill holes only in the memory used for prover_input_info, but we realized it would be better for Stone to receive a memory file without holes as well —since this they are already computed by the VM.
Therefore, we no longer intend to compare the memory against the Python VM’s memory, as they will not be identical.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2086)

### DIFF
--- a/vm/src/tests/compare_factorial_outputs_all_layouts.sh
+++ b/vm/src/tests/compare_factorial_outputs_all_layouts.sh
@@ -21,15 +21,6 @@ for layout in "plain" "small" "dex" "recursive" "starknet" "starknet_with_keccak
     else
         passed_tests=$((passed_tests + 1))
     fi
-    # Compare memory
-    echo "Running memory comparison for layout $layout"
-    if ! ./vm/src/tests/memory_comparator.py factorial_rs.memory factorial_py.memory; then
-        echo "Memory differs for layout $layout"
-        exit_code=1
-        failed_tests=$((failed_tests + 1))
-    else
-        passed_tests=$((passed_tests + 1))
-    fi
     # Compare air public input
     echo "Running air public input  comparison for layout $layout"
     if ! ./vm/src/tests/air_public_input_comparator.py factorial_rs.air_public_input factorial_py.air_public_input; then
--- a/vm/src/tests/compare_outputs_dynamic_layouts.sh
+++ b/vm/src/tests/compare_outputs_dynamic_layouts.sh
@@ -207,15 +207,6 @@ for case in "${CASES[@]}"; do
         passed_tests=$((passed_tests + 1))
     fi
 
-    # Compare memory
-    echo "Running memory comparison for case: $case"
-    if ! ./vm/src/tests/memory_comparator.py program_rs.memory program_py.memory; then
-        echo "Memory differs for case: $case"
-        exit_code=1
-        failed_tests=$((failed_tests + 1))
-    else
-        passed_tests=$((passed_tests + 1))
-    fi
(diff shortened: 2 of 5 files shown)


JulianGCalderon:
Hi @Stavbe!

Is there a way to keep the memory comparison anyway? Comparing with cairo-lang is a great way for testing the Cairo VM, and ensuring that we don't break compatibility.

We came up with two solutions:
- Add support for both unfilled and filled memory (maybe with a flag `--unfilled-memory`). That way, we can keep the memory comparisons. The flag `--memory` would output the filled memory (incompatible with cairo-lang).
- Sync these changes with cairo-lang, so that cairo-lang also fills memory. That way we can update the VM behaviour, but still compare the results to cairo-lang.

JulianGCalderon:
Another alternative could be to add a script that fills the memory holes when necessary, and leaving the VM output without holes. Would this work?

DavidLevitGurevich (approved):
:lgtm:

> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2086#-:-OQ8POEr7Pk1u4Lv038m:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Stavbe)

Stavbe (commented):
Hi @JulianGCalderon,  
We moved the hole-filling logic to run only in proof mode, so I updated this PR to perform memory comparison checks only in the other cases.

> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2086#-:-OQJL5ON7PsFrQAvhUaT:b-o8381s)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @DavidLevitGurevich and @Stavbe)

___
*[`vm/src/tests/compare_outputs_dynamic_layouts.sh` line 197 at r3](https://reviewable.io/reviews/lambdaclass/cairo-vm/2086#-OQJMsLM8iONL1URLhE5:-OQJMsLM8iONL1URLhE6:bgpgpfo) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/69efbc594c32ea3adfe4c241cbfb2f5ab2c24731/vm/src/tests/compare_outputs_dynamic_layouts.sh#L197)):*
> ```Shell
>     echo "Running cairo-lang with case: $case"
>     cairo-run --program "$full_program" \
>         --layout "dynamic" --cairo_layout_params_file "$full_layout" --proof_mode \
> ```

proof mode
___
*[`vm/src/tests/compare_factorial_outputs_all_layouts.sh` line 14 at r3](https://reviewable.io/reviews/lambdaclass/cairo-vm/2086#-OQJMqT46wQ0E4mdRd_a:-OQJMqT46wQ0E4mdRd_b:bgpgpfo) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/69efbc594c32ea3adfe4c241cbfb2f5ab2c24731/vm/src/tests/compare_factorial_outputs_all_layouts.sh#L14)):*
> ```Shell
>     # Run cairo_lang
>     echo "Running cairo_lang with layout $layout"
>     cairo-run --layout $layout --proof_mode  --program $factorial_compiled --trace_file factorial_py.trace --memory_file factorial_py.memory --air_public_input factorial_py.air_public_input --air_private_input factorial_py.air_private_input
> ```

proof mode

PR #2113, opened 2025-06-03
Enable_using_secure_run_in_proof_mode

YairVaknin-starkware:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/vm/runners/cairo_runner.rs
+++ b/vm/src/vm/runners/cairo_runner.rs
@@ -1014,13 +1014,22 @@ impl CairoRunner {
     pub fn get_builtin_segments_info(&self) -> Result<Vec<(usize, usize)>, RunnerError> {
         let mut builtin_segment_info = Vec::new();
 
+        let proof_mode = self.is_proof_mode();
+
         for builtin in &self.vm.builtin_runners {
             let (index, stop_ptr) = builtin.get_memory_segment_addresses();
 
-            builtin_segment_info.push((
-                index,
-                stop_ptr.ok_or_else(|| RunnerError::NoStopPointer(Box::new(builtin.name())))?,
-            ));
+            if proof_mode {
+                // proof‐mode: only push if `stop_ptr` is Some
+                if let Some(stop_ptr) = stop_ptr {
+                    builtin_segment_info.push((index, stop_ptr));
+                }
+            } else {
+                // non‐proof mode: error if `stop_ptr` is None
+                let stop_ptr =
+                    stop_ptr.ok_or_else(|| RunnerError::NoStopPointer(Box::new(builtin.name())))?;
+                builtin_segment_info.push((index, stop_ptr));
+            }
         }
 
         Ok(builtin_segment_info)


YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @yuvalsw)

___
*[`vm/src/vm/runners/cairo_runner.rs` line 1032 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-ORtdUdL7_HQVfWPW3-B:-OSDZ-CX-9LECf_CaMQ3:b-306t1y) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/9ad478367dc8bd7f3d207b0f89f74d7c69964ec0/vm/src/vm/runners/cairo_runner.rs#L1032)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

suggestion
</blockquote></details>

Refactored so we only check if proof_mode once.
___
*[`vm/src/vm/runners/cairo_runner.rs` line 1032 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-ORtbEzm0t5N8Oun7661:-OSDZrh99ASUUTh-S5MI:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/9ad478367dc8bd7f3d207b0f89f74d7c69964ec0/vm/src/vm/runners/cairo_runner.rs#L1032)):*
> ```XML
>                     stop_ptr.ok_or_else(|| RunnerError::NoStopPointer(Box::new(builtin.name())))?;
>                 builtin_segment_info.push((index, stop_ptr));
>             }
> ```

Done.

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113)* status: all files reviewed, 1 unresolved discussion (waiting on @yuvalsw)

___
*[`vm/src/vm/runners/cairo_runner.rs` line 1023 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSE_O0_6PdNOLKmb6eS:-OSEdfsr7OxnkeDxONCw:brtod4l) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/e5ad90b8e68ea618aec0d260e80ce7d93fc69d78/vm/src/vm/runners/cairo_runner.rs#L1023)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

While saving the `if` inside the loop, it complicates the code, also adding other indirections - so I am not sure it's more efficient (saving the if is pretty negligible). Unless I missed some other reason to use a closure here, I prefer the previous way.
Though I will not block on it, if you choose to leave it this way, please explain why.
</blockquote></details>

It's not there in terms of efficiency. They should be pretty similar in that regard (most efficient would just be duplication of the loop, but still really negligible for our use case). I Just like this style better and I think this avoids polluting the loop with a logic that could be decided earlier on in the run (it's pretty standard practice to have different methods do different things based on some branch logic, just here there's no need to define them outside the func scope and the logic is pretty "thin").

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -1012,15 +1015,30 @@ impl CairoRunner {
...
         for builtin in &self.vm.builtin_runners {
             let (index, stop_ptr) = builtin.get_memory_segment_addresses();
-
-            builtin_segment_info.push((
-                index,
-                stop_ptr.ok_or_else(|| RunnerError::NoStopPointer(Box::new(builtin.name())))?,
-            ));
+            if let Some(pair) = process(index, stop_ptr, builtin.name())? {
+                builtin_segment_info.push(pair);
+            }
```
JulianGCalderon: I don't see much benefit on defining a closure over inlining the logic into the for loop.

- It depends on a complex type alias.
- Its only used in one place (line 1039).
- Its harder to follow as the developer would need to jump to the closure definition when following the code path.

> this avoids polluting the loop with a logic that could be decided earlier on in the run

While the logic inside the loop could be decided earlier, I think that adding the if expression inside the loop is simpler than calling a closure that is conditionally defined outside the loop. In this case I would prioritize simplicity.

yuvalsw (changes requested):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113)* status: all files reviewed, 6 unresolved discussions (waiting on @YairVaknin-starkware)

___
*[`vm/src/vm/runners/cairo_runner.rs` line 1023 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSE_O0_6PdNOLKmb6eS:-OSKf_5s2aPa6io2eoUv:b-ctu990) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/e5ad90b8e68ea618aec0d260e80ce7d93fc69d78/vm/src/vm/runners/cairo_runner.rs#L1023)):*
<details><summary><i>Previously, YairVaknin-starkware wrote…</i></summary><blockquote>

It's not there in terms of efficiency. They should be pretty similar in that regard (most efficient would just be duplication of the loop, but still really negligible for our use case). I Just like this style better and I think this avoids polluting the loop with a logic that could be decided earlier on in the run (it's pretty standard practice to have different methods do different things based on some branch logic, just here there's no need to define them outside the func scope and the logic is pretty "thin").
</blockquote></details>

I disagree - the if way is much more readable and straight forward IMO. Anyway as said, not blocking on this if you insist.
___
*[`vm/src/vm/runners/cairo_runner.rs` line 1016 at r4](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSKgIFcB2dxeXcGhnr7:-OSKgIFcB2dxeXcGhnr8:undefined) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/1043f0a8272ea1b22b423df9081c8f9235fc6be0/vm/src/vm/runners/cairo_runner.rs#L1016)):*
> ```XML
> 
>     // Returns a map from builtin base's segment index to stop_ptr offset
>     // Aka the builtin's segment number and its maximum offset
> ```





_Suggestion:_
```XML
    /// Returns a map from builtin base's segment index to stop_ptr offset
    /// Aka the builtin's segment number and its maximum offset
```
___
*[`vm/src/vm/runners/cairo_runner.rs` line 1017 at r4](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSKgp2u0OHzSeiDmpYu:-OSKgp2u0OHzSeiDmpYv:b-tn16yp) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/1043f0a8272ea1b22b423df9081c8f9235fc6be0/vm/src/vm/runners/cairo_runner.rs#L1017)):*
> ```XML
>     // Returns a map from builtin base's segment index to stop_ptr offset
>     // Aka the builtin's segment number and its maximum offset
>     pub fn get_builtin_segments_info(&self) -> Result<Vec<(usize, usize)>, RunnerError> {
> ```

Any reason not to use HashMap?
If there is, please just adjust the doc saying it's a vector mapping these values, not "a map".



_Code quote:_
```XML
<Vec<(usize, usize)>
```
___
*[`vm/src/vm/runners/cairo_runner.rs` line 3938 at r4](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSKftIn9QJ7wQyDvC_E:-OSKftIn9QJ7wQyDvC_F:b-fwiq4h) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/1043f0a8272ea1b22b423df9081c8f9235fc6be0/vm/src/vm/runners/cairo_runner.rs#L3938)):*
> ```XML
>     #[test]
>     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
>     fn get_builtin_segments_info_non_proof_mode() {
> ```

consider uniting the test using parametrrization
___
*[`vm/src/vm/runners/cairo_runner.rs` line 3952 at r4](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSKhAPBD8MjXJlnWn3_:-OSKhAPBD8MjXJlnWn3a:b8k7fuh) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/1043f0a8272ea1b22b423df9081c8f9235fc6be0/vm/src/vm/runners/cairo_runner.rs#L3952)):*
> ```XML
>         let mut hint_executor = BuiltinHintProcessor::new_empty();
>         let runner = cairo_run(program_data, &cairo_run_config, &mut hint_executor).unwrap();
>         assert_eq!(runner.get_builtin_segments_info(), Ok(vec![(2, 6)]));
> ```

how only a single segment?
Which builtin is it?
Why is it index 2?

* Same questions for the other test
* Please reply to those questions in comments.

vm/src/vm/runners/cairo_runner.rs
YairVaknin-starkware: ok

YairVaknin-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113)* status: all files reviewed, 4 unresolved discussions (waiting on @JulianGCalderon and @yuvalsw)

___
*[`vm/src/vm/runners/cairo_runner.rs` line 1017 at r4](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSKgp2u0OHzSeiDmpYu:-OSNV8b-2WzgSdYenrJj:ba0417q) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/1043f0a8272ea1b22b423df9081c8f9235fc6be0/vm/src/vm/runners/cairo_runner.rs#L1017)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

Any reason not to use HashMap?
If there is, please just adjust the doc saying it's a vector mapping these values, not "a map".
</blockquote></details>

No need for random access here, and later used as a tuple in a few places. I don't see a reason to change it.
___
*[`vm/src/vm/runners/cairo_runner.rs` line 3938 at r4](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSKftIn9QJ7wQyDvC_E:-OSNVrsrAnxEwJgWG20P:bdg93q4) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/1043f0a8272ea1b22b423df9081c8f9235fc6be0/vm/src/vm/runners/cairo_runner.rs#L3938)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

consider uniting the test using parametrrization
</blockquote></details>

they don't have the needed dep. prefer not to add it or create a custom macro for this case.
___
*[`vm/src/vm/runners/cairo_runner.rs` line 3952 at r4](https://reviewable.io/reviews/lambdaclass/cairo-vm/2113#-OSKhAPBD8MjXJlnWn3_:-OSNWPst8xiMlZrNjLPk:b-xgv6ad) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/1043f0a8272ea1b22b423df9081c8f9235fc6be0/vm/src/vm/runners/cairo_runner.rs#L3952)):*
<details><summary><i>Previously, yuvalsw wrote…</i></summary><blockquote>

how only a single segment?
Which builtin is it?
Why is it index 2?

* Same questions for the other test
* Please reply to those questions in comments.
</blockquote></details>

It's just range check. idx 2 for non proof mode, idx 4 for proof mode since the layout contains some before it, but not used.

PR #2114, opened 2025-06-03
Refactor QM31 test

### LINKED ISSUE #1969: Refactor QM31 inv proptest
Separate `test_qm31_packed_reduced_inv_extensive` into a proptest for random values and particular test cases for the rest.

DiegoCivi:
# Refactor QM31 test

## Description

The test `test_qm31_packed_reduced_inv_extensive` is now made with `proptest`

Fixes #1969 

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/math_utils/mod.rs
+++ b/vm/src/math_utils/mod.rs
@@ -621,7 +621,7 @@ mod tests {
     use num_prime::RandPrime;
 
     #[cfg(feature = "std")]
-    use proptest::prelude::*;
+    use proptest::{array::uniform4, prelude::*};
 
     // Only used in proptest for now
     #[cfg(feature = "std")]
@@ -1307,7 +1307,7 @@ mod tests {
 
     /// Used for the QM31 tests
     #[cfg(feature = "std")]
-    #[derive(Clone, Copy, Debug)]
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
     enum Configuration {
         Zero,
         One,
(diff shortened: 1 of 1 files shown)


vm/src/math_utils/mod.rs
```diff
@@ -613,13 +613,15 @@ mod tests {
     use assert_matches::assert_matches;
 
     use num_traits::Num;
+
+    #[cfg(feature = "std")]
```
JulianGCalderon: Why did you gate this import?

vm/src/math_utils/mod.rs
```diff
@@ -1353,8 +1305,62 @@ mod tests {
...
+                                                            .prop_filter("All configs cant be Zero variant",
+                                                            |arr| !arr.iter().all(|x| *x == Configuration::Zero))
+                                                            .no_shrink()
+        ) {
+            let mut rng = SmallRng::seed_from_u64(11480028852697973135);
+            let x_coordinates: [u64; 4] = configs.map(|x| match x {
+                    Configuration::Zero => 0,
+                    Configuration::One => 1,
+                    Configuration::MinusOne => STWO_PRIME - 1,
+                    Configuration::Random => rng.gen_range(0..STWO_PRIME),
```
JulianGCalderon: I don't have experience with proptest, but is there a way to define a strategy that already contains this logic? That way we won't require the `Configuration` enum.

I imagine something like this, but I don't know if its possible (the following is pseudocode and probably won't be valid):
```rust
fn qm31_coordinate_strategy() -> impl Strategy<Value = ...> {
    prop_oneof![
        Just(0),
        Just(1),
        Just(STWO_PRIME - 1),
        (0..STWO_PRIME),
    ]
}
```

vm/src/math_utils/mod.rs
DiegoCivi: That crate is only being used inside the macro `proptest!()` which also has that feature constraint. Without this the compiler sees the crate as being unused ([here](https://github.com/lambdaclass/cairo-vm/actions/runs/15420047714/job/43392601326#step:9:273))

vm/src/math_utils/mod.rs
DiegoCivi: Good idea, it looks way better. Done in [`adc23ad`](https://github.com/lambdaclass/cairo-vm/pull/2114/commits/adc23add81e12e6693dc80135b67a1a5fcef7beb)

JulianGCalderon (approved):
LGTM

gabrielbosio (approved):

PR #2122, opened 2025-06-08
Fix broken issue link in CHANGELOG

MamunC0der:
The original link used /issue/ instead of the correct GitHub path /issues/, resulting in a 404 Page Not Found error. This fix restores proper navigation and ensures users and contributors can directly access the related issue.


gabrielbosio:
Thanks but we are not receiving these types of PRs for the moment

PR #2130, opened 2025-06-24
feat: use Felt directly instead of Felt252 alias

eeemmmmmm:
# TITLE
feat: use Felt directly instead of Felt252 alias

## Description
Implements a simple but important change: using the standardized `Felt` type directly from `starknet_types_core` instead of aliasing it as `Felt252`.

The change is minimal but strategic - it adds a backward-compatible type alias while setting the foundation for future code to use the proper `Felt` type. This aligns cairo-vm with the broader Starknet ecosystem's standardization efforts.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/lib.rs
+++ b/vm/src/lib.rs
@@ -66,7 +66,6 @@ pub mod types;
 pub mod utils;
 pub mod vm;
 
-// Using the standardized Felt type from starknet_types_core
 pub use starknet_types_core::felt::Felt;
 // For backwards compatibility - alias Felt252 to Felt
 pub type Felt252 = Felt;


vm/src/lib.rs
```diff
@@ -66,8 +66,10 @@ pub mod types;
 pub mod utils;
 pub mod vm;
 
-// TODO: use `Felt` directly
-pub use starknet_types_core::felt::Felt as Felt252;
+// Using the standardized Felt type from starknet_types_core
```
FrancoGiachetta: I think this comment can be removed since it's not adding any extra info.

eeemmmmmm: Okey, done

gabrielbosio:
Hi @eeemmmmmm, I am curious to know what project needs to use this alias so that we are sure we are not adding something to the VM API that is not used by anyone

eeemmmmmm:
Hey @gabrielbosio, this change improves code clarity by using the direct type name `Felt` while maintaining backward compatibility with a `Felt252` type alias. The alias ensures existing code continues to work without changes while allowing new code to use the more straightforward name

eeemmmmmm:
To answer your question: The Felt252 alias is widely used throughout the cairo-vm codebase itself. The alias maintains API stability for all existing code while allowing new code to use the more intuitive name.

gabrielbosio:
I realized that I have mixed terms... What I meant is if there is a project that needs to use the felt type from the VM using the original name `Felt` instead of the `Felt252` alias (One reason I think of is for name consistency with other dependencies, but I am curious if there is a concrete use case).

gabrielbosio:
> I realized that I have mixed terms... What I meant is if there is a project that needs to use the felt type from the VM using the original name Felt instead of the Felt252 alias (One reason I think of is for name consistency with other dependencies, but I am curious if there is a concrete use case).

Hi, @eeemmmmmm, any news on this? Is there a project that needs this change?

gabrielbosio:
I will close this PR because of inactivity. It can be reopened if necessary.

PR #2133, opened 2025-07-10
Stav/disable relocate trace with flag

Stavbe:
Up until now, whenever we wanted to keep the trace during execution, we also performed relocation at the end.
In the new adapter flow, we want to enable the option to keep the trace without allocating it.
I’ve set the default value to true to stay consistent with the current behavior.


Stavbe (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @anatgstarkware)

___
*[`vm/src/cairo_run.rs` line 31 at r1](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OVN0VZT-iCIuneRIWJ_:-OWACble1ol6r-ZZ2Dns:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/e0eec12f19b782d3b4233c7041ad7721413c92b2/vm/src/cairo_run.rs#L31)):*
<details><summary><i>Previously, anatgstarkware (anatg) wrote…</i></summary><blockquote>

Please document
</blockquote></details>

Done.

Stavbe (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @anatgstarkware)

___
*[`cairo-vm-cli/src/main.rs` line 195 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK440S4LqpDqtAdoAD:-OWKRpT1CokK696Da2XS:bwooj94) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/cairo-vm-cli/src/main.rs#L195)):*
<details><summary><i>Previously, anatgstarkware (anatg) wrote…</i></summary><blockquote>

I think this main is only used for stone, so maybe these should be true here, WDYT?
</blockquote></details>

I think it’s also used to run the VM outside of any prover context, which is why `prove_mode` is an argument, along with the public and private inputs.
I agree that it definitely won’t be used by Stwo, so I think that the best approach is to set 'reloacte_trace' to match `trace_enabled' and by that to turn off  Stwo "feature" .
As for the original `relocate_mem` flag—I think it’s best not to touch it.
___
*[`vm/src/cairo_run.rs` line 33 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK4hdv5_Qb7H1P475s:-OWKWC9W7RcUWjRSR_q3:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/vm/src/cairo_run.rs#L33)):*
<details><summary><i>Previously, anatgstarkware (anatg) wrote…</i></summary><blockquote>

It's not clear if you refer to the default below or to the main default args. I would remove this line
</blockquote></details>

Done.
___
*[`vm/src/cairo_run.rs` line 59 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK8g8H4he99bRdtsSL:-OWKWD4G0VCuGGO86MMC:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/vm/src/cairo_run.rs#L59)):*
<details><summary><i>Previously, anatgstarkware (anatg) wrote…</i></summary><blockquote>

This should probably be false, as the rest of the fields here
</blockquote></details>

Done.
___
*[`vm/src/cairo_run.rs` line 499 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK9J0bEVLROFQ-qruL:-OWKWDsX3q8yTlKrDKDZ:b-896fix) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/vm/src/cairo_run.rs#L499)):*
<details><summary><i>Previously, anatgstarkware (anatg) wrote…</i></summary><blockquote>

Since this test runs with no trace, it makes more sense to put here false
</blockquote></details>

Done.

Stavbe (commented):
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @anatgstarkware)

___
*[`cairo-vm-cli/src/main.rs` line 195 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK440S4LqpDqtAdoAD:-OWV7Qx-8xU32zojj8w2:b-j4qlis) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/cairo-vm-cli/src/main.rs#L195)):*
<details><summary><i>Previously, anatgstarkware (anatg) wrote…</i></summary><blockquote>

Didn't the vm relocate the trace also when `prove_mode` was false?
What do you mean by "original `relocate_mem` flag"? Was it in stone?
</blockquote></details>

1. The trace was relocated when either `args.trace_file.is_some()` or `args.air_public_input.is_some()`  
air_public_input requires `proof_mode = true`  
trace_file does not require it. 
So in theory the trace could be relocated even when `proof_mode = false` but I don't see why anyone would want that.

2. Yes
___
*[`vm/src/cairo_run.rs` line 59 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK8g8H4he99bRdtsSL:-OWVAMDW7IK9BsGB2sT1:bbbzcvr) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/vm/src/cairo_run.rs#L59)):*
<details><summary><i>Previously, anatgstarkware (anatg) wrote…</i></summary><blockquote>

I see you had to change the default in many tests, which means I was wrong. Are these tests only for stone? If so, maybe the default should be the same as in main (i.e., true, as you did before, or `trace_enabled`), WDYT?
Also, if `relocated_mem` was added for stwo, then maybe its default should also change.
</blockquote></details>

These tests are old, and they run the VM with the Stone configuration—so yes.  
I think the best approach is to align it with `trace_enabled`, but you need to fix a value in the default configuration so it is not really an option. I changed it back to `true` so it won't affect the current VM expected behavior.

Relocate mem was not added for stwo

anatgstarkware (approved):
:lgtm:

> Reviewed 1 of 4 files at r1, 1 of 2 files at r3, 2 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-:-OWVHkKU10ulubKPnApe:bnfp4nl)* status: 4 of 9 files reviewed, all discussions resolved (waiting on @Stavbe)

___
*[`cairo-vm-cli/src/main.rs` line 195 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK440S4LqpDqtAdoAD:-OWVGfzW-eM7bjPkHEt5:b-k73z76) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/cairo-vm-cli/src/main.rs#L195)):*
<details><summary><i>Previously, Stavbe wrote…</i></summary><blockquote>

1. The trace was relocated when either `args.trace_file.is_some()` or `args.air_public_input.is_some()`  
air_public_input requires `proof_mode = true`  
trace_file does not require it. 
So in theory the trace could be relocated even when `proof_mode = false` but I don't see why anyone would want that.

2. Yes
</blockquote></details>

thanks

anatgstarkware:
*[`vm/src/cairo_run.rs` line 59 at r2](https://reviewable.io/reviews/lambdaclass/cairo-vm/2133#-OWK8g8H4he99bRdtsSL:-OWVHOWjDfEnXru2bMo2:blk9znd) ([raw file](https://github.com/lambdaclass/cairo-vm/blob/b4a5c88a1cdcda5fddf8a9c66129909e3874ab7a/vm/src/cairo_run.rs#L59)):*
<details><summary><i>Previously, Stavbe wrote…</i></summary><blockquote>

These tests are old, and they run the VM with the Stone configuration—so yes.  
I think the best approach is to align it with `trace_enabled`, but you need to fix a value in the default configuration so it is not really an option. I changed it back to `true` so it won't affect the current VM expected behavior.

Relocate mem was not added for stwo
</blockquote></details>

Can you please document why it was set to true?

PR #2135, opened 2025-07-12
Optimize is_subsequence function to linear time complexity

reallesee:
# TITLE
Optimize is_subsequence function to linear time complexity

## Description
Improve performance by replacing repeated position() calls with a single-pass algorithm that maintains proper subsequence order checking.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -521,7 +521,7 @@ jobs:
         ls ${{ matrix.programs-dir }}/*.json | cut -f1 -d'.' | \
         xargs -P 2 -I '{program}' \
         ./target/release/cairo-vm-cli '{program}'.json --layout starknet_with_keccak \
-        --memory_file '{program}'.rs.memory --trace_file '{program}'.rs.trace \
+        --memory_file '{program}'.rs.memory --trace_file '{program}'.rs.trace --fill-holes false \
         ${{ matrix.extra-args }}
     - name: Update cache
       uses: actions/cache/save@v3
@@ -737,7 +737,6 @@ jobs:
         path: lcov-test-no_std#4-cairo-0-secp-hints.info
         key: codecov-cache-test-no_std#4-cairo-0-secp-hints-${{ github.sha }}
         fail-on-cache-miss: true
-
     - name: Upload coverage to codecov.io
       uses: codecov/codecov-action@v3
       with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,38 @@
 
 #### Upcoming Changes
 
+* feat: Add `--fill-holes` CLI flag instead of relying on `--proof-mode` [#2165](https://github.com/lambdaclass/cairo-vm/pull/2165)
+
+* feat: Use BTreeMap in PIE additional data [#2162](https://github.com/lambdaclass/cairo-vm/pull/2162)
+
+* feat: Remove prover input info struct and add getters instead [#2149](https://github.com/lambdaclass/cairo-vm/pull/2149)
+
+* feat: Added support for large files in PIE [#2136](https://github.com/lambdaclass/cairo-vm/pull/2136)
+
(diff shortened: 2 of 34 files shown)


vm/src/utils.rs
```diff
@@ -28,14 +28,20 @@ macro_rules! any_box {
...
 
 pub fn is_subsequence<T: PartialEq>(subsequence: &[T], mut sequence: &[T]) -> bool {
-    for search in subsequence {
-        if let Some(index) = sequence.iter().position(|element| search == element) {
-            sequence = &sequence[index + 1..];
-        } else {
-            return false;
+    if subsequence.is_empty() {
+        return true;
+    }
```
FrancoGiachetta, line 33: I think this is redundant, right? If the subsequence was empty, it wouldn't enter in the for loop and would return `true` (as it was before).

vm/src/utils.rs
```diff
@@ -28,14 +28,20 @@ macro_rules! any_box {
...
+    
+    let mut subseq_idx = 0;
+    for element in sequence {
+        if element == &subsequence[subseq_idx] {
+            subseq_idx += 1;
+            if subseq_idx == subsequence.len() {
+                return true;
+            }
         }
```
FrancoGiachetta, line 45: This if statement:
```rust
if element == &subsequence[subseq_idx] {
    ...            
}
```
Can panic silently if the index is not in bounds. To avoid any panics, you could do something like:
```rust
match subsequence.get(&subseq_idx) {
    Some(sub_elem) if element == &sub_elem => { ... }
    None => return false
}
```
With this change, I think you can also remove the `false` from line 44.

FrancoGiachetta:
Hi @reallesee! Thanks for the contribution. I left you some comments.

reallesee:
Hey [FrancoGiachetta](https://github.com/FrancoGiachetta), ready for review, thanks!

vm/src/utils.rs
```diff
@@ -28,17 +28,16 @@ macro_rules! any_box {
...
+            Some(sub_elem) if element == &sub_elem => {
+                subseq_idx += 1;
+                if subseq_idx == subsequence.len() {
+                    return true;
+                }
             }
+            None => return false
         }
     }
     false
```
FrancoGiachetta, line 47: This should be `true` actually

reallesee: should I squash commits?

vm/src/utils.rs
```diff
@@ -28,14 +28,19 @@ macro_rules! any_box {
...
+    let mut subseq_idx = 0;
+    for element in sequence {
+        match subsequence.get(&subseq_idx) {
+            Some(sub_elem) if element == &sub_elem => {
+                subseq_idx += 1;
+                if subseq_idx == subsequence.len() {
+                    return true;
+                }
+            }
+            None => return true
```
FrancoGiachetta: Hey @reallesee. I've just realized I read wrongly the implementation, sorry for that. Could you bring back the early return if the subsequence is empty? Also, in case of `None`, `false` should be returned since that would mean the index is out of bounds.

reallesee:
@FrancoGiachetta all good, to be honest I got a bit confused — could you please check?

gabrielbosio:
Hi @reallesee, thanks for contributing. Are there any benchmarks that show improvements with this change?

reallesee:
Hey @gabrielbosio 
Here are the benchmarks showing performance improvements:

New O(n) implementation results:
- 100 elements: 49.6 ns
- 500 elements: 259 ns  
- 1000 elements: 499 ns
- 5000 elements: 2.5 µs


image

gabrielbosio:
Ok, @reallesee, please add the corresponding benchmarks to the repo. You can add them [here](https://github.com/reallesee/cairo-vm/blob/main/bench/criterion_benchmark.rs).

gabrielbosio:
Hi, @reallesee! What's the status of this?

gabrielbosio:
I will close this PR because of inactivity. It can be reopened if necessary.

PR #2140, opened 2025-07-23
Pin types-rs version to the one set in lockfile

gabrielbosio:


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* chore: Pin types-rs version to the one set in lockfile [#2140](https://github.com/lambdaclass/cairo-vm/pull/2140)
```
FrancoGiachetta, line 5: I would also add that types-rs was updated to 0.1.8, right?

gabrielbosio: Because types-rs was not pinned, it was already updated to 0.1.8 before this PR, so I am not sure this counts as a dependency update.

FrancoGiachetta: Oh I see

igaray (approved):

PR #2152, opened 2025-08-07
fix: complete From<AirPrivateInputSerializable> implementation

Galoretka:
Add missing field handling for range_check96, add_mod, and mul_mod in the 
From<AirPrivateInputSerializable> implementation to prevent data loss during 
conversion between serializable and internal formats.

- Added range_check96 field processing
- Added add_mod field processing (wrapped in vector as per to_serializable logic)
- Added mul_mod field processing (wrapped in vector as per to_serializable logic)


gabrielbosio:
Hi, @Galoretka, it seems the linter job and some test jobs are failing, Could you please fix them?
Also, you have to add an entry in the changelog.

gabrielbosio:
Hi, @Galoretka, there are some jobs that are failing, could you please fix them?

gabrielbosio:
Closing due to inactivity. It can be reopened later if work needs to be resumed.

PR #2165, opened 2025-08-11
Enable memory comparisons again

### LINKED ISSUE #2154: Add memory comparison tests after merging `starkware-development` branch
Merging `starkware-development` into `main` implies removing memory comparison tests:
- https://github.com/lambdaclass/cairo-vm/pull/2153/files#diff-eafc39e18ed7bb654e966992615c4466f3f9e98af7d233dc5237a13b65033c03
- https://github.com/lambdaclass/cairo-vm/pull/2153/files#diff-2e8fb4d03f6f9a40f763254e5f3a34039d2abedbd78b3d1f5b5bc7289df19762

When adding these tests consider #2086 which explains the reason behind the removals. Either way, the memory comparison tests should be present in the repo.

JulianGCalderon:
# Enable memory comparisons again

Closes #2154.

Depends on #2076.

When in proof mode, the builtin segment holes are filled at the end of the run. This generates a discrepancy between this VM and the original Python Cairo VM. For this reason, we disabled the memory comparisons temporarily. See https://github.com/lambdaclass/cairo-vm/pull/2088 and https://github.com/lambdaclass/cairo-vm/pull/2086.

This PR adds a new flag `fill_holes` that can be used independently from `proof_mode`. This enables us to compare proof mode memory with the original Python Cairo VM.

The test coverage workflow is failing because this PR modifies a function that was not being tested.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -521,7 +521,7 @@ jobs:
         ls ${{ matrix.programs-dir }}/*.json | cut -f1 -d'.' | \
         xargs -P 2 -I '{program}' \
         ./target/release/cairo-vm-cli '{program}'.json --layout starknet_with_keccak \
-        --memory_file '{program}'.rs.memory --trace_file '{program}'.rs.trace \
+        --memory_file '{program}'.rs.memory --trace_file '{program}'.rs.trace --fill-holes false \
         ${{ matrix.extra-args }}
     - name: Update cache
       uses: actions/cache/save@v3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Cairo-VM Changelog
 
 #### Upcoming Changes
+
+* feat: Add `--fill-holes` CLI flag instead of relying on `--proof-mode` [#2165](https://github.com/lambdaclass/cairo-vm/pull/2165)
+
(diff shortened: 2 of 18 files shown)


gabrielbosio (approved):

.github/workflows/rust.yml
```diff
@@ -521,7 +521,7 @@ jobs:
         ls ${{ matrix.programs-dir }}/*.json | cut -f1 -d'.' | \
         xargs -P 2 -I '{program}' \
         ./target/release/cairo-vm-cli '{program}'.json --layout starknet_with_keccak \
-        --memory_file '{program}'.rs.memory --trace_file '{program}'.rs.trace \
+        --memory_file '{program}'.rs.memory --trace_file '{program}'.rs.trace --fill-holes false \
```
FrancoGiachetta, line 524: Clap's default value for booleans is `false`. So there's no need of adding `--fill-holes false`. The same for the makefile and the tests scripts. This is just to simplify things, you can ignore it if you prefer.

JulianGCalderon: The default value in this case is `None`, which implies that it should fallback to the value of `proof_mode`. Just to be safe I set it to false everywhere.

FrancoGiachetta: You are right, I didn't see it was an option. Well, you can this dismiss this comment.

PR #2180, opened 2025-08-21
Unpin `starknet-types-core` dependency

gabrielbosio:
Avoids dependency conflicts in crates that use the VM.

Reverts #2140.


gabrielbosio:
Both 2.x.y and 3.x.y versions now use `types-rs` 0.2.0. Closing.

PR #2191, opened 2025-09-02
Store constants in Hint Data

JulianGCalderon:
In some situations, we have a top-level Cairo program that itself executes other Cairo programs.

The Cairo Runner is being initialized with the top-level program. This means that, when executing hint, the constants passed to the hint belong to the top-level program. If the inner programs has hints that depends on constants defined only in the inner program, then the execution would fail as the constants are being searched in the top-level program's constants, and not in the inner program's constants.

This PR fixes this by having the constants specified on `compile_hint`, rather than on `execute_hint`. When compiling a hint for the inner program, we can specify the constants of that inner program.

# Breaking Changes

This PR introduces breaking changes:
- The signature of the HintProcessor methods changed. We now specify constants on `compile_hint`, rather than on `execute_hint`.
- The `CairoRunner::step_hint` method no longer receives the program constants, as they no longer require it.
- The `CairoRunner::step` method only receives constants when the `test-utils` features is enabled, as its used for the pre/post step hooks.


vm/src/vm/vm_core.rs
```diff
@@ -617,15 +614,14 @@ impl VirtualMachine {
         #[cfg(feature = "extensive_hints")] hint_datas: &mut Vec<Box<dyn Any>>,
         #[cfg(not(feature = "extensive_hints"))] hint_datas: &[Box<dyn Any>],
         #[cfg(feature = "extensive_hints")] hint_ranges: &mut HashMap<Relocatable, HintRange>,
-        constants: &HashMap<String, Felt252>,
+        #[cfg(feature = "test_utils")] constants: &HashMap<String, Felt252>,
```
FrancoGiachetta, line 617: Where is `constants` used after this change?

vm/src/vm/vm_core.rs
```diff
@@ -617,15 +614,14 @@ impl VirtualMachine {
...
     ) -> Result<(), VirtualMachineError> {
         self.step_hint(
             hint_processor,
             exec_scopes,
             hint_datas,
             #[cfg(feature = "extensive_hints")]
             hint_ranges,
-            constants,
         )?;
```
FrancoGiachetta, line 625: Before this change, `constants` was passed as an argument in the `step_hint` call. I can't find any usage of it now. Is it being used?

JulianGCalderon: Yes https://github.com/lambdaclass/cairo-vm/blob/e8978b6f6bb930e54e8f84aaf13182ea29ce2ce9/vm/src/vm/vm_core.rs?plain=1#L628

FrancoGiachetta: Oh, I missed it.

FrancoGiachetta (approved):

PR #2192, opened 2025-09-02
Resolve constants by full path

JulianGCalderon:
Depends on #2191

Resolves constants by full path, instead of only constant name. Currently, we are matching the first constant that shares the name, ignoring the full path to the constant. This has undefined behaviour if multiple constants have the same name.

To properly implement the lookup of constants, we need access two new elements:
- Identifiers (not only constants, but also aliases).
- Accessible scopes - included in https://github.com/lambdaclass/cairo-vm/pull/2042
This implied changing the signature of the hint processor.

The other big change in this PR is in the implementation of `get_constant_from_var_name`. This function finds the correct identifier by looking at the full path, and the accessible scopes.

I've added a test that fails with the current implementation. This PR solves it.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/hint_processor/builtin_hint_processor/uint384.rs
+++ b/vm/src/hint_processor/builtin_hint_processor/uint384.rs
@@ -109,8 +109,7 @@ pub fn add_no_uint384_check(
     let a = Uint384::from_var_name("a", vm, ids_data, ap_tracking)?;
     let b = Uint384::from_var_name("b", vm, ids_data, ap_tracking)?;
     // This hint is not from the cairo commonlib, and its lib can be found under different paths, so we cant rely on a full path name
-    let shift =
-        get_constant_from_var_name("SHIFT", identifiers, accessible_scopes)?.to_biguint();
+    let shift = get_constant_from_var_name("SHIFT", identifiers, accessible_scopes)?.to_biguint();
 
     let sum_d0 = (a.limbs[0].as_ref().to_biguint()) + (b.limbs[0].as_ref().to_biguint());
     let carry_d0 = BigUint::from((sum_d0 >= shift) as usize);


vm/src/hint_processor/builtin_hint_processor/sha256_utils.rs
```diff
@@ -53,25 +54,31 @@ fn sha256_main(
...
+    let input_chunk_size_felts = match input_chunk_size_felts.to_usize() {
+        Some(size) if size < 100 => size,
+        _ => {
+            return Err(HintError::AssertionFailed(
+                "assert 0 <= _sha256_input_chunk_size_felts < 100"
+                    .to_string()
+                    .into_boxed_str(),
+            ));
+        }
+    };
```
JulianGCalderon: In addition to updating the function to use accessible scopes, I also slightly changed the implementation to make it easier to understand.

cairo_programs/bad_programs/split_felt_bad_constants.cairo
```diff
@@ -3,10 +3,10 @@ const MAX_LOW = 1;
 
 func main() {
 	let value =1;
-	hint_func(MAX_HIGH, MAX_LOW, value);
+	hint_func(value);
 	return();
 }
-func hint_func(MAX_HIGH: felt, MAX_LOW: felt, value: felt) -> (felt, felt) {
+func hint_func(value: felt) -> (felt, felt) {
```
JulianGCalderon: By passing `MAX_HIGH` and `MAX_LOW` as references to the function, the hint was actually matching with the reference, instead of the actual constant (which is in the outer scope). This failed, because in the our Cairo VM, we are looking for only constants, and not any type of identifier.

I changed the test to not pass those values as arguments, and have the hints use the actual constant.

vm/src/hint_processor/builtin_hint_processor/mod_circuit.rs
```diff
@@ -50,21 +51,13 @@ pub fn run_p_mod_circuit_with_large_batch_size(
...
-    let batch_size = constants
-        .get(LARGE_BATCH_SIZE_PATH)
-        .ok_or_else(|| HintError::MissingConstant(Box::new(LARGE_BATCH_SIZE_PATH)))?;
-    #[cfg(not(feature = "mod_builtin"))]
+    let batch_size = get_constant_from_scoped_name("BATCH_SIZE", identifiers, accessible_scopes)?;
     let batch_size = batch_size
         .to_usize()
         .ok_or_else(|| MathError::Felt252ToUsizeConversion(Box::new(*batch_size)))?;
-    #[cfg(feature = "mod_builtin")]
-    let batch_size = 8; // Hardcoded here as we are not importing from the common lib yet
```
JulianGCalderon: I believe this was outdated, as the constants are part of the common library now. There is no need to have the value hardcoded anymore.

vm/src/hint_processor/builtin_hint_processor/hint_utils.rs
```diff
@@ -174,15 +178,74 @@ pub fn get_reference_from_var_name<'a>(
...
+            // If there is a match for at least the first component, we should
+            // not continue with the next scope.
+            if let Some(first_component) = scoped_name.split('.').next() {
+                if identifiers
+                    .get(&format!("{}.{}", scope, first_component))
+                    .is_some()
+                {
+                    break;
+                }
+            }
```
FrancoGiachetta: I don't really understand this part. What would be the value of `identifier` if the `break` statement was reached?

vm/src/hint_processor/builtin_hint_processor/secp/bigint_utils.rs
```diff
@@ -151,16 +153,15 @@ pub fn bigint_to_uint256(
...
 ) -> Result<(), HintError> {
     let x_struct = get_relocatable_from_var_name("x", vm, ids_data, ap_tracking)?;
     let d0 = vm.get_integer(x_struct)?;
     let d1 = vm.get_integer((x_struct + 1_i32)?)?;
     let d0 = d0.as_ref();
     let d1 = d1.as_ref();
-    let base_86 = constants
-        .get(BASE_86)
-        .ok_or_else(|| HintError::MissingConstant(Box::new(BASE_86)))?;
+    let base_86 = get_constant_from_scoped_name("BASE", identifiers, accessible_scopes)?;
```
FrancoGiachetta: Why call it BASE instead of `BASE_86`?

JulianGCalderon: The name of the constant is actually "BASE". "BASE_86" was the name of the Rust constant.

vm/src/hint_processor/builtin_hint_processor/hint_utils.rs
JulianGCalderon: None, we would break form the loop and return an error.

vm/src/hint_processor/builtin_hint_processor/secp/bigint_utils.rs
JulianGCalderon: I replace the `break` with a `return` instead, to make it more clear: d993cac576687f68e02f545bd8172ca308aa576a.

vm/src/hint_processor/builtin_hint_processor/hint_utils.rs
FrancoGiachetta: Oh you are right, didn't realize of that.

cairo_programs/bad_programs/split_felt_bad_constants.cairo
JulianGCalderon: I changed the implementation, and now this change is no longer required.

PR #2206, opened 2025-09-23
Optimize Cairo 0 execution

JulianGCalderon:
# Optimize Cairo 0 execution

## Description

This PR includes 2 minor optimizations, mainly targeted to Cairo 0 executions. I benchmarked block 10000 and compared the execution with v2.5.0. All benchmarks were ran in my M4 Macbook Pro.

- a8daa0aff977d1a215f0d2fcfbc87510187b92e2: Using with_capacity to avoid reallocs when inserting to HashMap, in `get_ids_data`.

- b4c8768a9aa411d26f8324069101fe633ff888f4: Use insert_all to load contiguous memory cells all at once, instead of one at a time.

## Benchmarks

I replayed multiple block ranges with my Macbook M4 Pro:

- Mainnet 10000 - 5% improvement
- Mainnet 20000 to 20010 - 5% improvement
- Mainnet 2000000 to 2000010 - 2% improvement

### DIFF
--- a/vm/src/vm/vm_memory/memory.rs
+++ b/vm/src/vm/vm_memory/memory.rs
@@ -789,6 +789,11 @@ impl Memory {
             ))));
         }
 
+        // Validate inserted memory cells
+        for i in 0..vals.len() {
+            self.validate_memory_cell((key + i)?)?;
+        }
+
         Ok(())
     }
 }


vm/src/vm/vm_memory/memory.rs
```diff
@@ -744,6 +744,53 @@ impl Memory {
...
+        key: Relocatable,
+        vals: &[MaybeRelocatable],
+    ) -> Result<(), MemoryError> {
+        let segment = self.get_segment(key)?;
+        let (_, value_offset) = from_relocatable_to_indexes(key);
+
+        // Allocate space for all new elements.
+        if segment.len() < value_offset + vals.len() {
+            segment.reserve(value_offset + vals.len() - segment.len());
+        }
```
DiegoCivi, line 766: nit: I think we can remove the ifsince the [documentation](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve) of `reserve()` says:

> Does nothing if capacity is already sufficient

DiegoCivi, line 766: nit: I think we can remove the if since the documentation of `reserve()` says:

> Does nothing if capacity is already sufficient

vm/src/vm/vm_memory/memory.rs
```diff
@@ -744,6 +744,53 @@ impl Memory {
...
+        // Fill middle with NONE.
+        if segment.len() < value_offset {
+            segment.resize(value_offset, MemoryCell::NONE);
+        }
+        // Insert new elements.
+        let last_element_to_replace = segment.len().min(value_offset + vals.len());
```
DiegoCivi, line 772: Isn´t the last index always `value_offset + vals.len()`? I don´t get in which case the segments len would be higher than that.

vm/src/vm/vm_memory/memory.rs
JulianGCalderon: I believe they refer to different things:
-  `segment.len() < value_offset + vals.len()` checks whether the _length_ of the segment is enough for holding the new elements. If not, it reserves capacity for the *additional* elements.
- the inner function checks whether the *capacity* of the segment is enough for holding the new elements. The capacity refers to the vector's allocated memory.

The if is required, as we only need to reserve new elements if the segment's length is not enough already. Without the condition, we would be sometimes be calling `reserve` with a negative argument (which would cause underflow as we are using a `usize`).

vm/src/vm/vm_memory/memory.rs
JulianGCalderon: The behavior of `splice` is a bit tricky. 

It receives two arguments:
- The range to replace.
- the elements to replace it with.

The length of the range and the length of the replacement does not need to coincide. For example, consider the following array:

```
[0, 1, 2, 3, 4, 5]
```

If we want to insert `[6,7,8]` at index `4`, we would be inserting 3 elements, but replacing only 2. The splice call would look like this:

```
splice(4..6, [6,7,8])
```

The result would look like this:

```
[0, 1, 2, 3, 6, 7, 8]
```

The following, instead, fails with index out of bounds, because we are replacing an element that does not exist.

```
splice(4..7, [6,7,8])
```

vm/src/vm/vm_memory/memory.rs
DiegoCivi: > ...we would be sometimes be calling reserve with a negative argument.

I thought that `value_offset` is always higher than `segment.len()`. 

Maybe I have the wrong understanding about segments, but if the lenght of a segment is 5. Doesn't it mean it has 5 allocated elements? If that is the case, having an offset lower than the lenght means it would want to write on already used memory which is something that it cannot be done, right?

vm/src/vm/vm_memory/memory.rs
DiegoCivi: Ohh I see. Awesome, thanks!

vm/src/vm/vm_memory/memory.rs
JulianGCalderon: > I thought that value_offset is always higher than segment.len().

I think that in `load_data` that is usually the case, but I'm not sure it would happen always. Consider the following segment:
```
[NONE, NONE, NONE, 10, 20]
```

We may want to call `insert_all` to insert 3 elements at the start of the segment. In that case, there is no need to reserve more space.  Note that having `NONE` is completely valid in a segment, those are commonly known as "memory gaps".

DiegoCivi: Perfect, I forgot you could have those. Thanks!

JulianGCalderon: Currently, `insert_all` is generic and supports the use case of inserting multiple elements at the middle of a segment.

If we make sure that `load_data` can only insert elements at the end of a segment, we could have another method (i.e. `extend_at`), only used for when inserting elements at the end of a segment. This could improve performance.

vm/src/vm/vm_memory/memory.rs
```diff
@@ -744,6 +744,53 @@ impl Memory {
...
     }
+
+    /// Inserts multiple values into contiguous memory addresses.
+    ///
+    /// If the address isn't contiguous with previously inserted data, memory gaps will be represented by NONE values
+    ///
+    /// Returns an error if:
+    /// - The segment index given by the address corresponds to a non-allocated segment.
+    /// - An inserted value is inconsistent with the current value at the memory cell
+    pub fn insert_all(
```
FrancoGiachetta, line 755: Validation rules are run whenever an insertion is performed. You can check [here](https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/vm_memory/memory.rs#L238). I think this is currently not being done. This function should validate every memory cell inserted.

FrancoGiachetta (changes requested):

vm/src/vm/vm_memory/memory.rs
JulianGCalderon: Done https://github.com/lambdaclass/cairo-vm/pull/2206/commits/4612ef6e116b41ed1497ac75a448c125a0a731c6

PR #2207, opened 2025-09-24
[cairo1] Fix system builtin as implicit

DiegoCivi:
# TITLE

## Description

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -469,20 +469,13 @@ fn load_arguments(
     cairo_run_config: &Cairo1RunConfig,
     main_func: &Function,
 ) -> Result<(), Error> {
-    let got_gas_builtin = main_func
-        .signature
-        .param_types
-        .iter()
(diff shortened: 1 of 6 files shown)


cairo1-run/src/cairo_run.rs
```diff
@@ -1008,6 +1004,11 @@ fn check_only_array_felt_return_type(
         _ => false,
     }
 }
+fn got_implicit_builtin(param_types: &[ConcreteTypeId], builtin_name: &str) -> bool {
+    param_types
+        .iter()
+        .any(|ty| ty.debug_name.as_ref().is_some_and(|n| n == builtin_name))
+}
```
FrancoGiachetta, line 1011: I would use `BuiltinName` here instead of a `&str`, just to be safe. You can convert a `ty.debug_name` with `BuiltinName::from(s)`

DiegoCivi: I was trying to implement it but there is no `BuiltinName` variant for the gas builtin or system builtin

PR #2216, opened 2025-09-30
Support wasm with cairo1

FrancoGiachetta:
# Support wasm with cairo1

## Description

This PR makes some changes so that a cairo 1 program can be run with wasm:
1. `cairo1-run` depends on cairo-vm with  the `std` feature set. Since wasm does not support std features, this PR creates an `std` feature for `cairo1-run`, which was set as a default feature to maintain compatibility. When used with wasm, default-features are set to `false`.
2. Changes `std` implementations for `HashMap` and `Peekable` with their `no_std` ones.
3. Due to point 2., I made cairo1-run's cli to requiere `std` feature.
4. Changes the initial gas used in `cairo1-run/src/cairo-run.rs` to be `u64` instead of `usize`. This was needed by wasm to be able to compile properly. 
5. Added a workflow to check wasm with cairo 1.

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* chore: Pin generic-array version to 0.14.7 or lower. [#2227](https://github.com/lambdaclass/cairo-vm/pull/2227)
+
 * feat: Add support for WASM with Cairo 1 [#2216](https://github.com/lambdaclass/cairo-vm/pull/2216)
 
 * fix: Added `cairo_1_test_contracts` and `cairo_2_test_contracts` as dependencies for `test-extensive_hints` target [#2201](https://github.com/lambdaclass/cairo-vm/pull/2201)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ lazy_static = { version = "1.4.0", default-features = false, features = [
 ] }
 nom = { version = "7", default-features = false }
 sha2 = { version = "0.10.7", features = ["compress"], default-features = false }
-generic-array = { version = "0.14.7", default-features = false }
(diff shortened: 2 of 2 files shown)


examples/wasm-demo-cairo1/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+    let cairo_run_config = Cairo1RunConfig {
+        layout: LayoutName::all_cairo,
+        relocate_mem: true,
+        trace_enabled: true,
+        serialize_output: true,
+        ..Default::default()
+    };
+
+    let sierra_program = {
+        let program_str = include_str!("../../../cairo_programs/cairo-1-programs/bitwise.sierra");
```
DiegoCivi, line 36: I think we are missing the `bitwise.sierra`. I am running `make test` and seeing the following error:
```
error: couldn't read `examples/wasm-demo-cairo1/src/../../../cairo_programs/cairo-1-programs/bitwise.sierra`: No such file or directory (os error 2)
  --> examples/wasm-demo-cairo1/src/lib.rs:36:27
   |
36 |         let program_str = include_str!("../../../cairo_programs/cairo-1-programs/bitwise.sierra");
```

FrancoGiachetta: Changed the Makefile so that it takes into account that file when testing.

gabrielbosio: I'm curious why this didn't get triggered in any CI job

FrancoGiachetta: Because, in the ci, that file gets built and cached before running the tests.

PR #2221, opened 2025-10-03
Add support for caching program segment

JulianGCalderon:
# Add support for caching program segment

I benchmarked some transactions and found a performance improvement of 3% when running old transactions (Cairo 0).
- The cache used is really simple (thread local). A proper cache for hint data would have more overhead than what I used.
- In my benchmark I ensured that everything is cached beforehand. This won't happen in an actual execution.

Some limitations:

- This PR is not breaking, but to achieve this I had to expose an unsafe API. See safety notes on the `load_cached_program` function.

- I had to make the memory public, so that the user is able to cache the program segment. We could add a wrapper type to avoid exposing the memory.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added


gabrielbosio:
Closing because it got outdated. I suggest going with #2223 before adding other Cairo 0 performance improvements.

PR #2223, opened 2025-10-07
Add Cairo Runner Builder

JulianGCalderon:
# Add Cairo Runner Builder

The `CairoRunner` has two main responsibilities:
- Setting the initial state for the VM execution
- Executing the VM

This PR introduces a new structure `CairoRunnerBuilder` that handles only the initialization of the Cairo VM. Doing so not only improves the API, but also allows for some optimizations that were not possible before.

## Optimizations

Each time we call the same contract, we do the same initialization logic:

- **Hint compilation.** As long as we have the same hints, and the same hint processor implementation, the compiled hints don't change.
- **Program loading (segment 0).** As long as the program is the same, the segment 0 doesn't change.
- **Instruction decoding.** As the long as the program is the same, the instruction decoding doesn't change.

This could be cached. The new builder implementes `Clone`, so we can perform the initialization logic only once, and reuse it for all the calls to that contract.

More over, when compiling hints, we are cloning the program's hints in order to put it into an `Rc`, which is required by the `compile_hint` function. To fix this, we are actually taking the program's hints (`mem::take`), instead of cloning it. This is safe, as long as the program's hints are not accessed **after** compiling them (which is not the case). Note that this optimization was possible even without the new builder.

## Alternatives

The whole idea of this new abstraction is that we allow a way for "cloning" the state of a runner, prior to the execution. I considered other approaches before settling for this one:

- The current builder has a method `fn build(self)` that consumes itself, and builds a CairoRunner. We could, instead, have something closer to a "factory" with a method `fn build(&self)`. This implies that we wouldn't need to clone something, just call `build` multiple times. The downside of this is that if the user doesn't want to cache the builder, and just build the runner once, it would be forced to clone the internal fields. The flow would look like this:
  1. Create a CairoRunerFactory
  2. Initialize it
  3. Build many CairoRunner instances

- Allow a way for Cloning the CairoRunner directly. We cannot derive Clone for the CairoRunner, so we need to have a manual implementation that skips some of fields. This would work, but implies cluttering the CairoRunner with even more initialization logic. Having a dedicated structure for this is a step in the right direction of improving the API. The flow would look like this:
  1. Create a CairoRuner
  2. Initialize it
  3. Clone it multiple times, to create many CairoRunner instances.

- Similar to the last alternative, we could do the whole initialization in the CairoRunner, and provide a way for building an alternative struct `CairoRunnerSnapshot`. This struct should be clonable, and should be able to be used to build a `CairoRunner` again. The flow would look like this:
  1. Create a CairoRuner
  2. Initialize it
  3. Create a CairoRunnerSnapshot
  4. Build many CairoRunner instances from the snapshot.

## Benchmarks

I did some benchmarks over block ranges:
- 20000-20010, found a 40% performance increase.
- 124000-124010, found a 37% performance increase.
- 124900-124910, found a 32% performance increase.
- 2000000-2000010, found a 30% performance increase.
- 2001000-2001010, found a 28% performance increase.

Some caveats:

- The benchmark assumes that everything is cached.
- The cache used is really simple: a static thread local.

## Sequencer Code

The required code for the benchmark can be found here: https://github.com/lambdaclass/sequencer/compare/main-v0.14.0..main-v0.14.0-builder

### Cairo 0 code

In the sequencer, I added the following code:

```rust
thread_local! {
    pub static CACHE: RefCell<HashMap<ClassHash, CairoRunnerBuilder>> = RefCell::new(HashMap::default());
}
let mut cairo_runner_builder = CACHE
    .with_borrow_mut(|cache| -> Result<CairoRunnerBuilder, VirtualMachineError> {
        match cache.entry(call.class_hash) {
            hash_map::Entry::Occupied(occupied_entry) => Ok(occupied_entry.get().clone()),
            hash_map::Entry::Vacant(vacant_entry) => {
                let mut cairo_runner_builder = CairoRunnerBuilder::new(
                    &compiled_class.program,
                    LayoutName::starknet,
                    None,
                    RunnerMode::ExecutionMode,
                )?;
                cairo_runner_builder.enable_trace(false);
                cairo_runner_builder.disable_trace_padding(false);
                cairo_runner_builder.allow_missing_builtins(false);
                cairo_runner_builder.initialize_base_segments();
                cairo_runner_builder.load_program()?;
                cairo_runner_builder.compile_hints(&mut syscall_handler).unwrap();
                cairo_runner_builder.initialize_builtin_runners_for_layout()?;
                cairo_runner_builder.initialize_builtin_segments();
                Ok(vacant_entry.insert(cairo_runner_builder).clone())
                
            }
        }
    })
    .unwrap();
```

After retrieving the builder, we call:

```rust
let mut runner = cairo_runner_builder.build()?;
```

After execution, we set the instruction cache for the following executions:

```rust
CACHE.with_borrow_mut(|cache| -> Result<(), VirtualMachineError> {
        match cache.get_mut(&call.class_hash) {
            Some(builder) => {
                builder.load_cached_instructions(runner.vm.take_instruction_cache())?
            }
            None => (),
        }
        Ok(())
    })?;
```

### Cairo 1 Code

For Cairo 1, I did something similar:

In the sequencer, I added the following code:
```rust
thread_local! {
    pub static CACHE: RefCell<HashMap<ClassHash, CairoRunnerBuilder>> = RefCell::new(HashMap::default());
}
let mut cairo_runner_builder: CairoRunnerBuilder = CACHE
    .with_borrow_mut(|cache| -> Result<_, VirtualMachineError> {
        match cache.entry(class_hash) {
            hash_map::Entry::Occupied(occupied_entry) => Ok(occupied_entry.get().clone()),
            hash_map::Entry::Vacant(vacant_entry) => {
                let mut cairo_runner_builder = CairoRunnerBuilder::new(
                    &compiled_class.program,
                    LayoutName::starknet,
                    None,
                    RunnerMode::ExecutionMode,
                )?;
                cairo_runner_builder.enable_trace(execution_runner_mode.trace_enabled());
                cairo_runner_builder.disable_trace_padding(false);
                cairo_runner_builder.allow_missing_builtins(false);
                cairo_runner_builder.initialize_base_segments();
                cairo_runner_builder.load_program()?;
                cairo_runner_builder.compile_hints(&mut syscall_handler).unwrap();
                cairo_runner_builder
                    .preallocate_segment(cairo_runner_builder.get_program_base().unwrap(), 2)?;
                Ok(vacant_entry.insert(cairo_runner_builder).clone())
            }
        }
    })
    .unwrap();
```

After retrieving the builder, we call:

```rust
cairo_runner_builder.initialize_builtin_runners(&entry_point.builtins)?;
cairo_runner_builder.initialize_builtin_segments();
let mut runner = cairo_runner_builder.build()?;
```

After execution, we set the instruction cache for the following executions:

```rust
CACHE.with_borrow_mut(|cache| -> Result<(), VirtualMachineError> {
        match cache.get_mut(&syscall_handler.base.call.class_hash) {
            Some(builder) => {
                builder.load_cached_instructions(runner.vm.take_instruction_cache())?
            }
            None => (),
        }
        Ok(())
    })?;
```

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/vm/src/vm/runners/cairo_runner.rs
+++ b/vm/src/vm/runners/cairo_runner.rs
@@ -255,6 +255,8 @@ impl CairoRunnerBuilder {
     /// Note that *initializing* a builtin implies creating a runner for it,
     /// and *including* a builtin refers to enabling the builtin runner flag:
     /// `included`.
+    ///
+    /// Analogue to [CairoRunner::initialize_builtins]
     pub fn initialize_builtin_runners_for_layout(&mut self) -> Result<(), RunnerError> {
         let builtin_ordered_list = vec![
             BuiltinName::output,
@@ -385,6 +387,9 @@ impl CairoRunnerBuilder {
     /// *Initializes* and *includes* all the given builtins.
     ///
     /// Doesn't take the current layout into account.
+    ///
+    /// Analogue to [CairoRunner::initialize_program_builtins], but receives the
+    /// builtins instead of reusing the program builtins.
     pub fn initialize_builtin_runners(
         &mut self,
         builtins: &[BuiltinName],
@@ -426,7 +431,8 @@ impl CairoRunnerBuilder {
     ///
     /// Depends on:
     /// - [initialize_base_segments](Self::initialize_base_segments)
-    /// - [initialize_builtin_runners_for_layout](Self::initialize_builtin_runners_for_layout)
+    /// - [initialize_builtin_runners_for_layout](Self::initialize_builtin_runners_for_layout) or
+    ///   [initialize_builtin_runners](Self::initialize_builtin_runners)
     pub fn initialize_builtin_segments(&mut self) {
         for builtin_runner in self.builtin_runners.iter_mut() {
             builtin_runner.initialize_segments(&mut self.memory);


vm/src/vm/vm_core.rs
```diff
@@ -568,6 +588,44 @@ impl VirtualMachine {
...
+    #[cfg(feature = "extensive_hints")]
+    pub(crate) fn step_hint_v2(
+        &mut self,
+        hint_processor: &mut dyn HintProcessor,
+        exec_scopes: &mut ExecutionScopes,
+        hint_datas: &mut Vec<Rc<Box<dyn Any>>>,
+        hint_ranges: &mut HashMap<Relocatable, HintRange>,
+    ) -> Result<(), VirtualMachineError> {
+        // Check if there is a hint range for the current pc
+        if let Some((s, l)) = hint_ranges.get(&self.run_context.pc) {
```
FrancoGiachetta: What do s and l mean? I think it might be better to rename them to some thing more understandable.

Cargo.toml
```diff
@@ -54,7 +54,7 @@ lazy_static = { version = "1.4.0", default-features = false, features = [
 ] }
 nom = { version = "7", default-features = false }
 sha2 = { version = "0.10.7", features = ["compress"], default-features = false }
-generic-array = { version = "0.14.7", default-features = false }
+generic-array = { version = "=0.14.7", default-features = false }
```
FrancoGiachetta: There's a [PR](https://github.com/lambdaclass/cairo-vm/pull/2227) for this. So this change should be removed once it's merged.

vm/src/vm/vm_core.rs
JulianGCalderon: I agree. The function I added is an exact copy of the original function (without `_v2`). We should later modify the original function also.

JulianGCalderon: Done 8df8954d6bafa79e15d52be0f2a4bf0b06941f39. I also modified the original function as it was simpler.

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -141,6 +149,452 @@ impl ResourceTracker for RunResources {
...
+                    RangeCheckBuiltinRunner::<RC_N_PARTS_96>::new(Some(1), true).into()
+                }
+                BuiltinName::add_mod => {
+                    ModBuiltinRunner::new_add_mod(&ModInstanceDef::new(Some(1), 1, 96), true).into()
+                }
+                BuiltinName::mul_mod => {
+                    ModBuiltinRunner::new_mul_mod(&ModInstanceDef::new(Some(1), 1, 96), true).into()
+                }
+            };
+            self.builtin_runners.push(builtin_runner);
```
DiegoCivi, line 420: Could this lead to a builtin_runners with the same builtin runner multiple times? If so, isn't that a problem? Also this lets the user to add runners that are not part of the layout. Is that a problem?

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -141,6 +149,452 @@ impl ResourceTracker for RunResources {
...
+    pub fn initialize_base_segments(&mut self) {
+        self.program_base = Some(self.add_memory_segment());
+        self.execution_base = Some(self.add_memory_segment());
+    }
+
+    /// Initializing the builtin segments.
+    ///
+    /// Depends on:
+    /// - [initialize_base_segments](Self::initialize_base_segments)
+    /// - [initialize_builtin_runners_for_layout](Self::initialize_builtin_runners_for_layout)
```
DiegoCivi: This means that both of those methods have to be used before this one? If so, why not unifying them into one single method?

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -141,6 +149,452 @@ impl ResourceTracker for RunResources {
...
+    pub fn initialize_builtin_segments(&mut self) {
+        for builtin_runner in self.builtin_runners.iter_mut() {
+            builtin_runner.initialize_segments(&mut self.memory);
+        }
+    }
+
+    /// Loads the program into the program segment.
+    ///
+    /// If this function is not called, the program will be loaded
+    /// automataically when initializing the entrypoint.
```
DiegoCivi:
```suggestion
    /// automatically when initializing the entrypoint.
```

DiegoCivi: If this is done anyways, is it necessary to let the user do it too? Maybe we could remove it from the initialization of the entrypoint so it is compulsory for the user to use it?

vm/src/vm/runners/cairo_runner.rs
```diff
@@ -486,13 +944,15 @@ impl CairoRunner {
...
-            self.vm.segments.memory.mark_as_accessed((prog_base + i)?);
+        if !self.loaded_program {
+            self.vm
+                .load_data(prog_base, &self.program.shared_program_data.data)
+                .map_err(RunnerError::MemoryInitializationError)?;
+
+            // Mark all addresses from the program segment as accessed
+            for i in 0..self.program.shared_program_data.data.len() {
+                self.vm.segments.memory.mark_as_accessed((prog_base + i)?);
+            }
```
DiegoCivi, line 961: Is it necessary to set `self.loaded_program` to _true_ here? I think it is not, but I am not sure

DiegoCivi: It is not necessary, you can dismiss this comment

vm/src/vm/runners/cairo_runner.rs
JulianGCalderon: Doing that would break the current behavior of the Cairo Runner. This way, we allow the user to preload the program segment if he wants to, but not break current behavior if the user does not do it.

JulianGCalderon: Ideally, we would force the user do call this method, but we need to wait until next release for that.

vm/src/vm/runners/cairo_runner.rs
JulianGCalderon: We *could* do it - it would be technically correct, but its not necessary. The current behavior is that if the user calls this method twice, the program would be loaded twice, which isn't ideal but _should_ never happen anyway.

I wanted to minimize changing the current behavior of the Cairo Runner, just in case. No one should depend on this behavior, but you never know.

Cargo.toml
JulianGCalderon: That PR is for the main branch, not the 2.x.y branch, right?

vm/src/vm/runners/cairo_runner.rs
JulianGCalderon: There are at least 2 ways of initializing the builtin runners (there may be more):
- Initializing the entrypoint builtin runners (Cairo 1 - sequencer).
- Initializing the layout builtin runners, according to the program builtins (Cairo 0 - sequencer).

The flow, at least for the use cases that I researched, seems to be like this:

- `initialize_base_segments`.
- Either `initialize_builtin_runners_for_layout` or `initialize_builtin_runners`. Now that I think of it, the code comment is wrong.
- `initialize_builtin_segments`.

There are two reasons why I didn't abstract it yet into higer-level functions:

- I didn't want to create abstractions prematurely, as there are some usecases that I am not considering yet.
- I didn't want to depart much from the current Cairo Runner building API.

I totally agree in that we need to find a better/safer API, but I am not sure what is the best way to do it yet.

cc: @FrancoGiachetta @gabrielbosio

vm/src/vm/runners/cairo_runner.rs
JulianGCalderon: > Could this lead to a builtin_runners with the same builtin runner multiple times?

Yes, if the user specifies multiple times the same builtin. Receiving the builtin runners is required as the user may only want to initialize the builtins required by the entrypoint that is going to be executed. This is a problem in the [current Cairo Runner API ](https://github.com/lambdaclass/cairo-vm/blob/84d3e77f2afb00ab17075a32199dd75c1fbb457a/vm/src/vm/runners/cairo_runner.rs?plain=1#L1785) also.

 >  If so, isn't that a problem?
 
 I am not sure if its a problem, but it never came up.
 
 > Also this lets the user to add runners that are not part of the layout. Is that a problem?
 
 The current behavior ignores the layout, so I don't think its a problem, but I am not sure really.

vm/src/vm/runners/cairo_runner.rs
DiegoCivi: Yess, I agree

vm/src/vm/runners/cairo_runner.rs
FrancoGiachetta: I think it's fine to let the user decide how to build the runner, at least with the same flexibility as the previous one. It is true that it may be quite unsafe, as calling functions in a different order might result in an inconsistent state. However, we've seen some use cases where the runner needs to be built in exotic ways. So I think that, in our way of creating a safer api, we you consider these things as they shouldn't be change.

PR #2224, opened 2025-10-08
Refactor of hintcode match

DiegoCivi:
# Refactor of hintcode match

## Description

The idea of this optimization is to move the match done in `execute_hint()` to `compile_hint()`. This could bring performance improvements since we would be doing the match only once for each hint and not for every time we execute them.

To do this change, the HintProcessorData now has to hold a function pointer to the function it implements that hint. To do this, every function that implements a hint needed to have the same signature.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/cairo_programs/benchmarks/repeated_hint.cairo
+++ b/cairo_programs/benchmarks/repeated_hint.cairo
@@ -36,7 +36,7 @@ func inv_mod_p_uint512{range_check_ptr}(x: Uint512) -> Uint256 {
 }
 
 func recursive_hint{range_check_ptr: felt}(n: felt) {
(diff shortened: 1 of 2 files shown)


DiegoCivi:
This change didn´t show the expected improvements. If we run the benchmarks that were added on an AMD Ryzen 9 with 125 gb of memory we get virtually the same times with this change or without it:
- With change (this branch)
    - cairo0 (avg):  3.0962 s
    - cairo (avg): 2.6181 s
- Without change (main)
    - cairo0 (avg): 3.0189 s
    - cairo (avg): 2.6450 s

PR #2225, opened 2025-10-09
Enable pipefail in macOS install script to fail on pipeline errors

sashaodessa:
### Description
- **What**: Add `-o pipefail` to `install-scripts/install-macos.sh` shell options.
- **Why**: Without `pipefail`, commands like `curl ... | sh` can mask failures in the left side of the pipeline, leading to false‑positive installs and harder debugging.
- **Impact**: More reliable CI/local setup; script now exits on any pipeline error.

### DIFF
--- a/install-scripts/install-macos.sh
+++ b/install-scripts/install-macos.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -ex
+set -ex -o pipefail
 
 # Install Rust and Cargo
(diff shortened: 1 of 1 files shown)


gabrielbosio:
Why does this change only affect one bash script?

gabrielbosio:
Hi, @sashaodessa, I'm closing this PR because the change is too small to be considered. I suggest to is to add `-o pipefail` in the rest of the bash scripts that need it unless there's a specific reason to do it in the affected script only.

PR #2231, opened 2025-10-15
Compile hints without unnecessary references

FrancoGiachetta:
# Compile hints without unnecessary references

## Description

Currently, when compiling a hint, all accessible references are taken into account. This means, there may be references which are computed end stored in `HintProcessorData` but actually are not used by the hint at all. 
For every hint that gets compiled, the function `get_ids_data` gets called. This function's code is as follows:

```rust
fn get_ids_data(
    reference_ids: &HashMap<String, usize>,
    references: &[HintReference],
) -> Result<HashMap<String, HintReference>, VirtualMachineError> {
    let mut ids_data = HashMap::<String, HintReference>::new();
    for (path, ref_id) in reference_ids {
        let name = path
            .rsplit('.')
            .next()
            .ok_or(VirtualMachineError::Unexpected)?;
        ids_data.insert(
            name.to_string(),
            references
                .get(*ref_id)
                .ok_or(VirtualMachineError::Unexpected)?
                .clone(),
        );
    }
    Ok(ids_data)
}
```
Depending on the hint, and the context in which it gets called, there may be quite a lot of references which will be iterated in this function without any reason at all. To avoid having to compute unnecessary references, this PR changes how these references are computed for some hints. Basically, for those cases where the hint code does not make use of any reference, it will not call `get_ids_data` and simply assign an empty `HashMap`. This may avoid unnecessary overhead. 

Benchmarks

Run some ranges of blocks comparing cairo vm 2.5.0 with this PR. These numbers correspond to the first run of each range. Running them a second time tends to lower the improvement, sometime to levels which are indistinguishable between the two versions. This could probably be due to caching. 

20000-20010: 3% improvement.
2000000: 1% improvement.

## Checklist
- [ ] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -232,8 +232,8 @@ jobs:
     - name: Check all features (${{ matrix.crate }})
       run: |
         cd ${{ matrix.crate }}
-        cargo check-all-features
-        cargo check-all-features --workspace --all-targets
+        cargo all-features check
+        cargo all-features check --workspace --all-targets
 
   smoke-workspace:
     needs: merge-caches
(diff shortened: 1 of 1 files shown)


gabrielbosio:
Closing because it's outdated. I think it doesn't show a significant improvement to consider open a new one pointing to v3.

PR #2236, opened 2025-10-18
fix: error mapping for fee_provision in excess_balance hint

radik878:
The fee_provision computation incorrectly mapped both a missing fee entry and a Decimal.checked_mul failure to ExcessBalanceKeyError("fees"). This conflated a data lookup issue with a calculation overflow/invalid-state issue. Updated logic to:
- Return ExcessBalanceKeyError("fees") only when fees.get(&account) is None.
- Return ExcessBalanceCalculationFailed("fee_provision") when checked_mul fails.

This aligns error semantics with the rest of the function, which consistently distinguishes key-missing vs calculation failures.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
     - name: Install Rust
-      uses: dtolnay/rust-toolchain@1.87.0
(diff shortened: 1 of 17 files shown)


JulianGCalderon:
Hi @radik878, thanks for the contribution! Could you update the changelog?

radik878:
> Hi @radik878, thanks for the contribution! Could you update the changelog?

Yes boss!

PR #2246, opened 2025-10-24
fix(fuzz): correct num1/num2 substitution in bitwise Cairo template

gap-editor:
**Pull Request description:**

My PR fixes a small but critical issue in the `program_bitwise` Cairo template used by the fuzzer.

Previously, some calls (e.g. `bitwise_xor(num1, num2)` and `bitwise_or(num1, num2)`) referenced identifiers `num1` and `num2` that were **never declared inside the generated Cairo program**.
This caused Cairo compilation errors such as `Unknown identifier 'num1'` and could result in false positives during fuzzing runs.

The fix ensures that all function calls consistently substitute actual numeric values using `{num1}` and `{num2}`, just like the rest of the template.

### DIFF
--- a/fuzzer/src/cairo_compiled_programs_fuzzer.rs
+++ b/fuzzer/src/cairo_compiled_programs_fuzzer.rs
@@ -68,7 +68,7 @@ fn program_array_sum(
     let populated_array = array
         .iter()
         .enumerate()
-        .map(|(index, num)| format!("assert [ptr + {}] = {};  \n", index, num))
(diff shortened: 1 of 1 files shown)


JulianGCalderon:
Hey @gap-editor, thanks for finding this bug in our fuzzer! However, this PR only fixes the immediate bug, and does not tackle the root cause. We prefer to solve both problems at the same time.

I've created a more detailed issue on this topic: https://github.com/lambdaclass/cairo-vm/issues/2254. Feel free to tackle it if you want. Let me know and I'll assign it to you.

PR #2256, opened 2025-11-05
perf(tracer): precompute prime/constants in get_data and avoid clones

radik878:
Precompute prime_bigint, half_prime, 2^40, and 2^100 once per request in get_data() to avoid per-element parsing and exponentiation.
Refactor field_element_repr() to accept precomputed constants by reference, removing unnecessary clone()s and repeated BigInt computations.
Rationale: Reduce CPU and allocations when mapping memory values to strings for the tracer API, leveraging existing cached CAIRO_PRIME.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* perf: precompute prime/constants in get_data and avoid clones in field_element_repr [#2256](https://github.com/lambdaclass/cairo-vm/pull/2256)
+
 #### [3.0.0-rc.4] - 2025-28-10
 
 * fix: error mapping for fee_provision in excess_balance hint [#2236](https://github.com/lambdaclass/cairo-vm/pull/2236)


gabrielbosio:
Hi, @radik878! Do you have any benchmark that shows any improvement? If so, please share the results

radik878:
Before: 
image
image

and after: 
image
image
image
image

gabrielbosio:
It seems there is no improvement from this benchmarks from what I see. Also, if there is a benchmark that show any improvement it would be nice if it's also added to this PR. Otherwise I think it should be better to close this PR because it intends to improve performance but no improvements are shown in the posted results.

radik878:
> It seems there is no improvement from this benchmarks from what I see. Also, if there is a benchmark that show any improvement it would be nice if it's also added to this PR. Otherwise I think it should be better to close this PR because it intends to improve performance but no improvements are shown in the posted results.

I thinks if there is no improvement, i should close this, thank you for your review and sorry for taking time

PR #2265, opened 2025-11-13
Make blake2s API internal

gabrielbosio:
This is a step towards moving Blake2s logic outside of the VM so that it can be used by Native or other crates without depending on the VM or repeating code.


FrancoGiachetta (approved):

JulianGCalderon (approved):
LGTM!

PR #2266, opened 2025-11-13
fix: save x1 variable to execution scope in fast_ec_add_assign_new_x

Bilogweb3:
The function computes x1 from point1.x and uses it in calculations, but never saves it to execution scope. This doesn't match the Python hint spec which explicitly defines x1 in scope, and breaks the pattern used in similar functions like ec_double_assign_new_x where all computed variables are saved.

Added exec_scopes.insert_value("x1", x1) to fix this oversight.

### DIFF
--- a/docs/hint_processor/builtin_hint_processor/README.md
+++ b/docs/hint_processor/builtin_hint_processor/README.md
@@ -82,7 +82,7 @@ cairo_run(
     .expect("Couldn't run program");
 ```
 #### Final notes:
-The example used in this guide can be found [here](../../../custom_hint_example/).
+The example used in this guide can be found [here](../../../examples/custom_hint/).
 The example can be ran using `make example`
 
 ### How to code your hint implementation:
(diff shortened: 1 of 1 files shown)


gabrielbosio:
Hi, @Bilogweb3!

- Can you link to the Python hint you mention in the description?
- Do you have a Cairo program that causes the VM to fail in `main` but not in this branch? It would be nice to have it in a new test case.

Bilogweb3:
@gabrielbosio HI!

The Python hint is actually right in our test file here: https://github.com/lambdaclass/cairo-vm/blob/main/cairo_programs/fast_ec_add_v2.cairo#L41-L51

You can see it creates x1 in the scope along with x0 and y0.

About a failing test - honestly, I don't have one. The thing is, x1 isn't actually read from the scope afterwards, so programs run fine without this fix. 

What bothered me was that we're computing x1 and using it in the calculation, but then not saving it to scope like the hint says we should. Every other similar function (like ec_double_assign_new_x) saves all the variables it computes, so this looked like someone just forgot to add that line when copying code.

I can write a test that verifies x1 exists in the scope after execution if you think that would be useful? It wouldn't catch a crash, but it would verify we're actually doing what the hint specification says.

Let me know what you think.

gabrielbosio:
> What bothered me was that we're computing x1 and using it in the calculation, but then not saving it to scope like the hint says we should. Every other similar function (like ec_double_assign_new_x) saves all the variables it computes, so this looked like someone just forgot to add that line when copying code.

I think it wasn't added to the execution scope because there are no other hints that use that variable so it wasn't necessary.

> I can write a test that verifies x1 exists in the scope after execution if you think that would be useful? It wouldn't catch a crash, but it would verify we're actually doing what the hint specification says.

It might be useful since a crate that uses the `BuiltinHintProcessor` would expect the `execution_scopes` to include that variable, so yes I would go with that.

Bilogweb3:
Added a test for this. It checks that x0, x1, and y0 are in the scope after execution - same pattern as the other tests in the file.

gabrielbosio (approved):
Nice job!

gabrielbosio:
One more thing @Bilogweb3, you have to add the corresponding entry in the changelog to enable merging this PR

PR #2268, opened 2025-11-16
fix: assign validation rules in place to preserve segment-index alignment

Galoretka:
Fix Memory::add_validation_rule to assign in place instead of inserting, preventing Vec index shifting that misaligns validation rules with segment indices; add a unit test to verify that adding rules out of order does not shift existing entries and that validate_existing_memory applies rules to the correct segments.

### DIFF
--- a/.github/workflows/iai_main.yml
+++ b/.github/workflows/iai_main.yml
@@ -27,7 +27,8 @@ jobs:
       run: |
         sudo apt update
         sudo apt install -y valgrind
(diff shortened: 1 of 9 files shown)


gabrielbosio:
Hi, @Galoretka! This PR looks good but please add the corresponding changelog entry.

Galoretka:
> Hi, @Galoretka! This PR looks good but please add the corresponding changelog entry.

Did it

PR #2271, opened 2025-11-20
fix: rename y_cube_int to y_square_int in get_point_from_x

strmfos:
Fixed incorrect variable naming in secp signature hint. The variable was called y_cube_int but actually contains y² (y squared), not y³.

This matches the Python hint which explicitly names it y_square_int,and reflects the actual math: we compute x³ + β which equals y² on the elliptic curve.

The variable is used for modular square root extraction, so the name now correctly describes what it contains.


gabrielbosio:
Hi @strmfos, thanks for contributing! You have to add the corresponding changelog entry

DiegoCivi (approved):

Issue #2255, opened 2025-10-29
Review diff_fuzzer fuzzer


JulianGCalderon:
We should:

- Ensure that the fuzzer is up to date: that all the commands work correctly.
- Add more documentation on how to run it, and their requirements.

When trying to run it, I get the following error:

```
 === Uncaught Python exception: ===
IndexError: list index out of range
Traceback (most recent call last):
  File "/Users/julian/work/cairo-vm/fuzzer/diff_fuzzer/random_hint_fuzzer.py", line 37, in diff_fuzzer
    hint = fdp.PickValueInList(LOADED_HINTS)
IndexError: list index out of range

==83281== ERROR: libFuzzer: fuzz target exited
```

PR #2272, opened 2025-11-22
docs: align README with current fuzzer and generator behavior

VolodymyrBg:
The README contained outdated details about how hints are sourced and selected, file generation flow, and numeric ranges used for fuzzing. It previously mentioned a get_random_hint function and whitelists under hint_accountant, but the actual selection comes from load_hints in hint_reader.py, which extracts hints from vm/src/hint_processor/builtin_hint_processor/hint_code.rs and filters those with right-shifts or SECP_P before choosing via Atheris. The document also implied writing .cairo and .json files during normal runs, whereas the implementation compiles in-memory and only writes a failing .cairo when reproduction is needed. The range documentation used a 2^128 bound, but the code uses PRIME = 2251 + 17 * 2192 + 1 and enforces 70/15/15 probabilities over PRIME-derived intervals. Examples for main and hint_func were updated to pass inputs and return the union of declared variables to reflect cairo_program_gen.py’s real signatures. Minor wording corrections and a note on constants import (e.g., ADDR_BOUND) were added for accuracy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* docs: align README with current fuzzer and generator behavior [#2272](https://github.com/lambdaclass/cairo-vm/pull/2272)
+
 #### [3.0.0] - 2025-11-19
 
 #### [3.0.0-rc.5] - 2025-11-14


gabrielbosio:
Hi, @VolodymyrBg, thanks for your contribution, however #2255 tracks the status of that fuzzer because it's not working at the moment so there's a high chance that this documentation will become outdated once #2255 is finished. The issue mentioned also includes updating documentation so the work done in this PR might be included in the PR that solves the issue. That's why I'm closing this PR for now.

PR #2269, opened 2025-11-18
fix: preserve verified offset flags

reject-i:
## Description

 Replace Vec::insert with resize+direct assignment in extend_additional_data to avoid shifting existing offsets in the Pedersen verified-address cache.

PR #2275, opened 2025-12-02
fix(hash): keep verified_addresses index=offset invariant

Fibonacci747:
The hash builtin tracks verified output cells in a Vec where the index equals the memory offset. Using Vec::insert shifted elements and broke this invariant, especially with non-sorted inputs or pre-existing entries. Replace insert with resize(offset + 1, false) and direct assignment to ensure stable indexing and alignment with deduce_memory_cell() and the documented semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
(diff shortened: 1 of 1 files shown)


gabrielbosio:
Duplicate of #2269

PR #2276, opened 2025-12-02
fix: correct duplicate tuple index in InvalidTrackingGroup error

Fallengirl:
Fixed copy-paste bug in InvalidTrackingGroup error formatting.

The error message was using (*.0).0 twice instead of (*.0).0 and (*.0).1, causing both placeholders to display the same value from the tuple.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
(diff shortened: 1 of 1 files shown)


gabrielbosio:
Hi, @Fallengirl, thanks for contributing! Please add the corresponding changelog entry

Fallengirl:
Hi! Sure

PR #2278, opened 2025-12-10
refactor: remove unused SECP256R1_P entries from exec scopes

Bashmunta:
dropped redundant SECP256R1_P inserts into ExecutionScopes in Cairo 0 secp hints, since that value is never read back from the scope and all computations already use the global constant directly. This removes unnecessary BigInt cloning and keeps the hint implementations aligned with the original Cairo semantics without altering behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 #### Upcoming Changes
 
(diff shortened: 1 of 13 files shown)


gabrielbosio:
This is breaking because with this change the hint processor mutates the exec scopes differently. So this seems to alter behavior and its inclusion should be planned and discussed in a future major release.

PR #2289, opened 2025-12-29
fix: simplify tracer instruction encoding fetch

Snezhkko:
Previous implementation of get_instruction_encoding computed the immediate address using a modulo by the Cairo prime via BigUint, and accepted the program prime as a string parameter. In the relocated trace we work with plain usize indices into a contiguous memory vector, so (pc + 1) % prime is both unnecessary and misleading: the field modulus does not apply to memory indices. It also added avoidable allocations and parsing work per instruction and risked panics. The function now uses straightforward bounds-checked indexing (pc and pc + 1) and no longer depends on prime or BigUint, while the call site in 
TracerData::new was updated accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
+* fix: simplify tracer instruction encoding fetch [#2289](https://github.com/lambdaclass/cairo-vm/pull/2289)
+
 #### [3.0.1] - 2025-12-22
 
 * fix: Fix compute_ids_high_low hint constant path [#2285](https://github.com/lambdaclass/cairo-vm/pull/2285)


cairo-vm-tracer/src/tracer_data.rs
```diff
@@ -205,17 +203,22 @@ impl TracerData {
 pub fn get_instruction_encoding(
     pc: usize,
     memory: &[Option<Felt252>],
-    prime: &str,
 ) -> Result<(Felt252, Option<Felt252>), TraceDataError> {
-    if memory[pc].is_none() {
+    if pc >= memory.len() {
         return Err(TraceDataError::InstructionIsNone(pc.to_string()));
     }
```
FrancoGiachetta, line 209: I would rather use a different kind of error. This condition now represents an out of bounds indexing, so it might be misleading while debugging.

FrancoGiachetta:
Hi [Snezhkko](https://github.com/Snezhkko)! 
Thanks for the contribution! I left you a little comment.

gabrielbosio:
@Snezhkko what's the status of this PR?

gabrielbosio:
Closing due to inactivity. We can reopen it if necessary.

PR #2294, opened 2026-01-12
Breaking: Remove `bincode` dependency

### LINKED ISSUE #2290: Remove `bincode`

FrancoGiachetta:
# Breaking: Remove `bincode` dependency

## Description

The PR removes `bincode` crate dependency as it has been left permanently unsupported. This PR also removes `write_encoded_trace` and `write_encoded_memory` from the api exposed. 

Closes #2290

## Checklist
- [x] Linked to Github Issue
- [ ] Unit tests added

### DIFF
--- a/cairo-vm-cli/src/main.rs
+++ b/cairo-vm-cli/src/main.rs
@@ -1,8 +1,6 @@
 #![deny(warnings)]
 #![forbid(unsafe_code)]
 use cairo_vm::air_public_input::PublicInputError;
-use cairo_vm::cairo_run;
-use cairo_vm::cairo_run::{write_encoded_memory, write_encoded_trace, EncodeTraceError};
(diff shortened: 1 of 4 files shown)


DiegoCivi (approved):

DiegoCivi:
> _`std::io::Write::write` does not do the same as `bincode::enc::write::Writer::write` as the function that does the same is `std::io::Write::write::write_all` so we should change the calls to `write` with `write_all`_

Done in [`9de6e9a`](https://github.com/lambdaclass/cairo-vm/pull/2294/commits/9de6e9a9023b3416cddeb553e9f126ea457dd7a4)

gabrielbosio:
> std::io::Write::write does not do the same as bincode::enc::write::Writer::write as the function that does the same is std::io::Write::write::write_all so we should change the calls to write with write_all

Leaving here why this change was reverted: Making `write_encoded_*` functions public that use `std::io::Write` loses no_std support (which makes sense). #2300 addresses this.

PR #2295, opened 2026-01-14
Add `StepHooks` trait

piotmag769:
# Add `StepHooks` trait

## Description

This PR introduces a `StepHooks` trait that `Hooks` implements. This allows having hooks that are stateful. It is necessary for [cairo-debugger](https://github.com/software-mansion-labs/cairo-debugger) since its struct `CairoDebugger` acts as a stateful post step vm hook - check [this PR](https://github.com/software-mansion-labs/cairo-debugger/pull/55/files#diff-f659afe4d46f31f955149ac44d3e19cdfbcbbbba53eb64e0dd5a65add4db3cd1R12).

## Checklist
- [ ] Linked to Github Issue - irrelevant
- [ ] Unit tests added - irrelevant
- [ ] Integration tests added - irrelevant
  - [ ] Documentation has been added/updated - irrelevant

### DIFF
--- a/vm/src/vm/hooks.rs
+++ b/vm/src/vm/hooks.rs
@@ -13,11 +13,11 @@ use crate::stdlib::{any::Any, collections::HashMap, prelude::*, sync::Arc};
 
 use crate::Felt252;
 
+use super::{errors::vm_errors::VirtualMachineError, vm_core::VirtualMachine};
+use crate::vm::runners::cairo_runner::CairoRunner;
 use crate::{
     hint_processor::hint_processor_definition::HintProcessor, types::exec_scope::ExecutionScopes,
 };
-
-use super::{errors::vm_errors::VirtualMachineError, vm_core::VirtualMachine};
 
 type BeforeFirstStepHookFunc = Arc<
     dyn Fn(&mut VirtualMachine, &[Box<dyn Any>]) -> Result<(), VirtualMachineError> + Sync + Send,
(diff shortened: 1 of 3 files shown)


vm/src/vm/vm_core.rs
```diff
@@ -110,7 +110,7 @@ pub struct VirtualMachine {
     pub(crate) disable_trace_padding: bool,
     instruction_cache: Vec<Option<Instruction>>,
     #[cfg(feature = "test_utils")]
-    pub(crate) hooks: crate::vm::hooks::Hooks,
+    pub(crate) hooks: Option<Box<dyn crate::vm::hooks::StepHooks>>,
```
piotmag769, line 113: It has to be `Option` to be able to move it out of the VM struct for the time of executing the actual hook - otherwise we get double mutable borrow.

vm/src/vm/hooks.rs
```diff
@@ -59,16 +86,62 @@ impl Hooks {
...
+    fn before_first_step(
+        &mut self,
+        vm: &mut VirtualMachine,
+        hint_data: &[Box<dyn Any>],
+    ) -> Result<(), VirtualMachineError> {
+        if let Some(before_first_step) = &self.before_first_step {
+            before_first_step(vm, hint_data)
+        } else {
+            Ok(())
+        }
```
FrancoGiachetta: I think there's no need of this `else` case. It might be better:

```rust
if let Some(before_first_step) = &self.before_first_step {
    ...
}

Ok(())
```

FrancoGiachetta, line 97: I think it is simpler not to use the else clause and just do:

```rust
if let Some(before_first_step) = &self.before_first_step {
    return before_first_step(vm, hint_data);
}

Ok(())
```

I've seen some changes similar to this one. I think it make the code simpler.

FrancoGiachetta:
Hi [piotmag769](https://github.com/piotmag769)!
I think the PR is fine overall, just a small comment. Also, could you update the changelog?

PR #2296, opened 2026-01-16
refactor(hint_processor): make reference deref count explicit

sashaodessa:
Make dereference semantics in `get_maybe_relocatable_from_reference` explicit by replacing overlapping conditionals with a small loop.

The previous code duplicated `vm.get_maybe(&val)?` behind two overlapping `if` branches, which makes it harder to reason about whether we dereference 0/1/2 times and easier to “simplify” incorrectly later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 #### Upcoming Changes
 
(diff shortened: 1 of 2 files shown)


gabrielbosio:
Hi @sashaodessa please add the corresponding changelog entry

sashaodessa:
> Hi @sashaodessa please add the corresponding changelog entry

@gabrielbosio added

gabrielbosio (approved):

PR #2299, opened 2026-01-21
fix(ec_op): avoid false-positive same-x failures when no addition occurs

Bashmunta:
The ec_op_impl routine returned EcOpSameXCoordinate even on iterations where the scalar bit was 0 and no point addition was performed. This could reject valid inputs due to a same-x condition that was never actually used in an addition.

What was done: 
- Moved the same x-coordinate check under the bit=1 branch so it triggers only when an actual addition is attempted.
- Updated the existing same-x failure test to guarantee the failure happens on the first real addition.
- Added a regression test ensuring we do not fail when same-x occurs on an iteration with bit=0.


gabrielbosio:
Hi @Bashmunta, there are failing CI jobs, please fix

gabrielbosio:
Closing due to inactivity.

PR #2389, opened 2026-08-31
fix: SegmentArena recovery under append_return_values + RelocateAllDictionaries no-op

fmkra:
When running Cairo 1 programs with `--append_return_values` (or proof mode), two related bugs in the exit wrapper caused failures for programs that use dictionaries together with other builtins, or that contain dict code on untaken branches.

### Bug 1 — SegmentArena pointer recovered from a fragile FP offset

**Symptom**

Programs whose `main` carries **SegmentArena and Poseidon** (and often Bitwise) fail arena finalization:

```text
VirtualMachine(Math(RelocatableSubUsizeNegOffset((Relocatable { segment_index: N, offset: 0 }, 2))))
```

or:

```text
VirtualMachine(Memory(AddressNotRelocatable))
```

**Minimal repro**

```cairo
use core::dict::{Felt252Dict, Felt252DictTrait};
use core::poseidon::hades_permutation;

pub fn main(input: Array<felt252>) -> Array<felt252> {
    let mut d: Felt252Dict<felt252> = Default::default();
    Felt252DictTrait::insert(ref d, 0, 0);
    let (h, _, _) = hades_permutation(*input[0], 0, 0);
    let _b = 1_u32 & 1_u32; // optional; Bitwise also triggers
    array![h]
}
```

```bash
scarb build  # enable-gas = false recommended for cairo1-run
cairo1-run target/dev/<pkg>.sierra.json \
  --layout all_cairo \
  --append_return_values \
  --cairo_pie_output /tmp/pie.zip \
  --args '[7]'
```

**Why it failed**

After `main` returns, the exit wrapper serializes the return (and input) array into the output segment. That path uses `rescope{…}` blocks, which drop AP-relative CASM variables — including the live SegmentArena pointer.

The wrapper then tried to recover the pointer with a hard-coded offset:

```text
fp + 2 * builtins.len() + 2
```

That cell was the **initial** SegmentArena implicit (`base+3`), not the pointer `main` returned. Dict alloc/destruct copies the 3-cell header forward, so `ptr[-2]` on `base+3` reads the stale header zeros and arena validation becomes a no-op. Extra builtins (Poseidon / Bitwise) or gas shift that FP cell onto a felt or a segment base; `segment_arena_ptr[-2]` then crashes.

| Program                  | Result (before)         |
| ------------------------ | ----------------------- |
| dict only                | OK (validation skipped) |
| poseidon only            | OK                      |
| dict + hades_permutation | **FAIL**                |
| dict + hades + bitwise   | **FAIL**                |

**Fix**

Reserve a dedicated FP-stable local at entry (`[fp + builtins.len()]`), save the final SegmentArena pointer there **before** output-serialization rescopes, and read it back for `RelocateAllDictionaries` / arena validation. Update `load_arguments` (+1 to `ap_offset` when SegmentArena is present) and the input-serialization FP offset so user args stay aligned.

Using the final pointer means arena validation actually runs. `RelocateAllDictionaries` already left a one-cell gap between later dicts (`prev_end += 1`) to match `assert curr_start = prev_end + one`, but the first relocated dict was placed at `first.end` with no gap. Advance `prev_end` by one before that first relocation.

### Bug 2 — `RelocateAllDictionaries` when no dict was allocated

**Symptom**

```text
VirtualMachine(Hint((0, VariableNotInScopeError("dict_manager_exec_scope"))))
```

**Minimal repro**

```cairo
use core::dict::{Felt252Dict, Felt252DictTrait};

fn dict_path(input: Span<felt252>) -> felt252 {
    let mut d: Felt252Dict<felt252> = Default::default();
    Felt252DictTrait::insert(ref d, 0, *input[0]);
    Felt252DictTrait::get(ref d, 0)
}

pub fn main(input: Array<felt252>) -> Array<felt252> {
    if *input[0] == 1 {
        array![dict_path(input.span())]
    } else {
        array![42]
    }
}
```

```bash
# fails
cairo1-run … --append_return_values --args '[0]'
# succeeds
cairo1-run … --append_return_values --args '[1]'
```

**Why it failed**

Using `Felt252Dict` anywhere in the Sierra program adds a SegmentArena implicit, so the exit wrapper always injects `RelocateAllDictionaries`. That hint required `dict_manager_exec_scope`, which is only created when a dict allocation hint actually runs. On the `[0]` path no dict is allocated → scope missing → crash. Dict allocation hints already lazy-init the scope; the relocate cheatcode did not.

**Fix**

Make `RelocateAllDictionaries` a no-op when `dict_manager_exec_scope` is absent (nothing to relocate). Behavior is unchanged when a dict was used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo-vm/2389)

### DIFF
--- a/cairo1-run/src/cairo_run.rs
+++ b/cairo1-run/src/cairo_run.rs
@@ -455,6 +455,7 @@ After calling this function (before running the VM):
     (*1+2+3) gap
     (*1+2) gap
     (*1+2) gap
+    (*1+2) gap
     (*2) gap
     (*2) gap
     (*2) gap
@@ -475,6 +476,7 @@ After the entry_code (up until calling main) has been ran by the VM:
     (*1) gap (for output_builtin final ptr)
     (*1) gap (for builtin_0 final ptr)
     (*1) gap (for builtin_1 final ptr)
+    (*1+2) gap (for final SegmentArena ptr)
     (*2) segment_arena_ptr
     (*2) infos_ptr
     (*2) 0
@@ -506,6 +508,8 @@ fn load_arguments(
     // These include:
     // * The builtin bases (not including output)
     // * (Only if the output builtin is added) A gap for each builtin's final pointer
+    // * (Only if the output builtin is added and SegmentArena is present) A gap for the
+    //   final SegmentArena pointer
     // * The segment arena values (if present), including:
     //  * segment_arena_ptr
     //  * info_segment_ptr
@@ -514,6 +518,10 @@ fn load_arguments(
     let mut ap_offset = runner.get_program().builtins_len();
     if cairo_run_config.copy_to_output() {
         ap_offset += runner.get_program().builtins_len() - 1;
+        // Extra local reserved for the final SegmentArena pointer (see create_entry_code).
+        if got_segment_arena {
+            ap_offset += 1;
+        }
     }
     if got_segment_arena {
         ap_offset += SEGMENT_ARENA_GAPS;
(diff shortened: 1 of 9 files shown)


cairo1-run/src/cairo_run.rs
```diff
@@ -701,6 +716,15 @@ fn create_entry_code(
             let local = ctx.add_var(CellExpression::Deref(deref!([fp + i.to_i16().unwrap()])));
             casm_build_extend!(ctx, assert local = var;);
         }
+        // Preserve SegmentArena across output-serialization `rescope`s in the dedicated
+        // local at [fp + builtins.len()] reserved above.
```
fmkra, line 720: Extra explanation:
Dedicated [fp + builtins.len()] local: the final SegmentArena pointer must be saved here before output-serialization rescopes drop AP-relative vars. A hardcoded offset (previously fp + 2*builtins.len() + 2) recovered the initial base+3 implicit — or, once Poseidon/Bitwise/gas shifted the frame, a felt or segment base (AddressNotRelocatable / RelocatableSubUsizeNegOffset).

PR #2392, opened 2026-09-09
fix: run hints of each visited pc in run_for_steps

orizi:
`run_for_steps` computed the hint slice for the entry pc once and reused it for every step, so hints at subsequent pcs never ran and the entry pc's hints re-ran on every step. Resolve the slice per step, as `run_until_pc` does. Only the non-`extensive_hints` path was affected.

Adds a regression test with a `vm_enter_scope()` hint attached to a non-entry pc.

Based on #2391.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo-vm/2392)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Both branches support Stwo prover opcodes (Blake2s, QM31) since v2.0.0.
 ---
 
 #### Upcoming Changes
+* fix: `run_for_steps` now executes the hints of every pc it visits; previously it reused the hints of the entry pc for all steps [#2392](https://github.com/starkware-libs/cairo-vm/pull/2392)
+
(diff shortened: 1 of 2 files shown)


ilyalesokhin-starkware:
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/starkware-libs/cairo-vm/2392#-P1dO8Ue64OKcSGPL8TB:-P1dO8Ue64OKcSGPL8TC:bt0ueup):*
rephrase



_Code quote:_
```
  run_for_steps computed the hint slice for the entry pc once and reused
  it for every step, so hints at subsequent pcs never ran and the entry
  pc's hints re-ran on every step. Resolve the slice per step, as
  run_until_pc does. Only the non-extensive_hints path was affected.
```

PR #2401, opened 2026-09-15
refactor: remove the extensive_hints feature

orizi:
Prototype: deletes the `extensive_hints` feature entirely (major-release change, no feature back-compat).

## Why

The feature was a non-additive flag that changed public signatures (`step`, `step_hint`, `relocate_value`, `add_relocation_rule`), leaking ~64 cfg pairs through vm_core/cairo_runner/memory — while *enabling* it cost 20–30% on hint-heavy workloads (a `HashMap<Relocatable, _>` lookup per step instead of a vector index, plus a per-run clone of the full ranges map).

## Design

Both capabilities, no flag:

- The program's own hints stay in the vector-indexed `hints_ranges` (the fast path, borrowed from the program — no per-run clone).
- Hints added at runtime (bootloader / Starknet OS loading programs together with their hints) go into an overflow `HashMap<Relocatable, (start, len)>`, consulted only when non-empty — one `is_empty` branch per step. Runtime-added hints override static ones, preserving the extensive insert semantics.
- Relocation rules always accept `MaybeRelocatable` destinations; the VM always dispatches through `execute_hint_extensive` (whose default still calls `execute_hint`).
- Per-pc hint resolution moved from the runner loops into `step_hint`, so `run_until_pc`/`run_for_steps` share one shape (the #2392 bug class can no longer occur).

## Performance

Interleaved best-of-11, `--layout all_cairo --proof_mode`, local x86-64 Linux, noise ~±2.5%:

| benchmark | base (feature off) | old `extensive_hints` build | this PR (unified) |
|---|---|---|---|
| big_fibonacci | 802ms | +0.5% | +2.5% |
| pedersen | 844ms | +0.3% | −4.2% |
| dict_integration | 647ms | **+18.8%** | +1.3% |
| keccak_integration | 1149ms | **+21.4%** | +0.2% |
| math_integration | 647ms | **+19.0%** | +1.9% |
| linear_search | 816ms | **+32.2%** | +3.8% |

i.e. the always-on capability now costs ~0–4% (worst case, most dispatch-sensitive benchmark) instead of 19–32%.

## Validation

- clippy clean on default / `--all-features` / `--no-default-features`; full test-target type-check clean.
- `cargo test -p cairo1-run` end-to-end suites pass.
- `run_deprecated_contract_class_simplified` — the runtime-hint-loading test, previously compiled only under `extensive_hints`, now unconditional — **passes** on the unified path.

## Downstream

`starknet_os` (sequencer) should drop `"extensive_hints"` from its cairo-vm features when picking this up; the capability is now unconditional.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Performance (measured, vs main)

CPU-pinned, interleaved runs, best-of-13, `--layout all_cairo --proof_mode`, local x86-64 Linux. A layout control (main rebuilt with 64 dead bytes appended) measured −1.0% to +0.1%, so the noise floor of this comparison is ~±1-2%.

| benchmark | main | old `extensive_hints` build | this PR |
|---|---|---|---|
| big_fibonacci | 961ms | +4.9%* | +0.9% |
| pedersen | 997ms | −1.5%* | +2.7% |
| dict_integration | 698ms | **+22.7%** | +1.0% |
| keccak_integration | 1339ms | **+17.3%** | +1.2% |
| math_integration | 713ms | **+23.8%** | +0.3% |
| linear_search | 1031ms | **+25.6%** | +1.9% |

\* hint-free programs; the feature's cost is per hint-range lookup.

i.e. the always-on capability costs at most ~2% (at the measurement noise floor) instead of 17–26%. The repo's iai (callgrind) CI job can settle the residual with exact instruction counts.

**A finding worth noting for reviewers:** the first version of this PR measured a uniform +6% — bisected (via single-change builds and a dead-byte layout control) to `Memory::get`'s inner `relocate_value` no longer being inlined once its return type widened to `MaybeRelocatable` (a call + 40-byte struct-return copy per memory read). The `#[inline]` on the two `RelocateValue` impls is load-bearing, and is commented as such.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/cairo-vm/2401)


orizi:
Added a second commit packing the run's hint state into a single `RunHints` struct (compiled hint datas + program's static ranges + runtime-added ranges), per review feedback — `step`/`step_hint` now take one argument for all hint state, and range resolution/extension live behind its methods. Re-validated (clippy all configs, test-target compile, dynamic-hint-loading test) and re-benchmarked pinned/interleaved: +0.3% to +0.9% vs the fast-path baseline on the most sensitive benchmarks — within the ±3% cross-binary layout-noise floor established in the PR body.

orizi:
Cherry-picked the three CI fixes from #2399 onto this branch (with `-x`), same as the perf-stack PRs carry them: this PR is based on main, whose workflows still have the pre-existing failures #2399 fixes (unpinned cache action, transitively-unpinned machete action, EOL debian-11), which is what the red checks here were. The cherry-picks dissolve once #2399 merges and this branch rebases.

